PR #13, opened 2019-03-04
Problem: TX validation not in an enclave (WIP)

tomtau:
For comparison with https://github.com/crypto-com/chain/pull/11


tomtau:
closing it, as the dev will continue for some time -- moves to a forked repo

PR #23, opened 2019-03-22
(CRO-5 CRO-68) Crypto.com Chain Client: Wallet implementation

devashishdxt:


client-core/src/key.rs
```diff
@@ -0,0 +1,61 @@
...
+    /// Generates a new private key
+    pub fn new() -> Result<PrivateKey> {
+        let mut rng = OsRng::new().context(ErrorKind::KeyGenerationError)?;
+        let secret_key = SecretKey::new(&mut rng);
+
+        Ok(PrivateKey(secret_key))
+    }
+
+    /// Serializes current private key
+    pub fn serialize(&self) -> Result<Vec<u8>> {
```
tomtau: are these for import/export?

client-core/src/key.rs
```diff
@@ -0,0 +1,61 @@
...
+
+    /// Deserializes private key from bytes
+    pub fn deserialize_from(bytes: &[u8]) -> Result<PrivateKey> {
+        let secret_key: SecretKey = deserialize(bytes).context(ErrorKind::DeserializationError)?;
+
+        Ok(PrivateKey(secret_key))
+    }
+}
+
+impl Zeroize for PrivateKey {
```
tomtau: perhaps `impl Drop for PrivateKey` that would call zeroize?

client-core/src/wallet.rs
```diff
@@ -0,0 +1,95 @@
...
+
+                private_key.zeroize();
+
+                Ok(public_key)
+            }
+        }
+    }
+
+    /// Generates a new address
+    fn generate_address(&self, name: &str, passphrase: &str) -> Result<String> {
```
tomtau: What is this one used for?

client-core/src/lib.rs
```diff
@@ -0,0 +1,35 @@
+#![deny(missing_docs, unsafe_code, unstable_features)]
+//! # Crypto.com Chain Client
+//!
+//! This crate exposes following functionalities for interacting with Crypto.com Chain:
+//! - Wallet creation
+//! - Address generation
+//! - Balance tracking
```
tomtau: I suppose balance tracking is not implemented yet?

devashishdxt: No. Right now, only wallet creation and address generation are implemented. I'm still working on those features and also, a PR with all the features will become too big to review.

client-core/src/wallet.rs
devashishdxt: This will be used by CLI / RPC to generate new addresses for a wallet.

PR #25, opened 2019-03-26
(CRO-53) Balance tracking and synchronisation

devashishdxt:


client-core/src/balance/transaction_change.rs
```diff
@@ -0,0 +1,94 @@
...
+use crate::balance::BalanceChange;
+use crate::Result;
+
+use chain_core::init::coin::Coin;
+
+/// Represents balance change in a transaction
+#[derive(Debug)]
+pub struct TransactionChange {
+    /// ID of transaction which caused this change
+    pub transaction_id: String,
```
tomtau: why not `TxId` type from chain_core?

client-core/src/balance/transaction_change.rs
```diff
@@ -0,0 +1,94 @@
...
+    /// Address which is affected by this change
+    pub address: String,
```
tomtau: similarly here, ` ExtendedAddr`?

devashishdxt:
This is a basic implementation of balance tracking. `Chain` trait is not implemented in this pull request. The implementation of `Chain` trait will involve querying Crypto.com Chain full nodes through ABCI.

client-core/src/wallet.rs
```diff
@@ -10,21 +10,29 @@ use zeroize::Zeroize;
 
 use chain_core::init::address::RedeemAddress;
 
-use crate::service::{KeyService, WalletService};
-use crate::{ErrorKind, PublicKey, Result, Storage};
+use crate::service::{BalanceService, KeyService, WalletService};
+use crate::{Chain, ErrorKind, PublicKey, Result, Storage};
 
 /// Interface for a generic wallet
```
tomtau: there's not much functionality yet, and only dozes of these layered wrappers -- it's hard to know what problem this is aiming to solve. It may become clear when there's an actual implementation, but still it'd be good to document what this is supposed to be: e.g. what's the difference between `Wallet` and `WalletService` traits? 
We currently don't have a style guide on comments beyond formatting, but I'd envision something like this: https://docs.corda.net/codestyle.html#comments

> Comments that simply repeat the story already told by the code are best deleted.
> Explain what the code is doing at a higher level than is obtainable from just examining the statement and surrounding code.
> Explain why certain choices were made and the trade-offs considered.
> Explain how things can go wrong, which is a detail often not easily seen just by reading the code.

I think here the top-level comment for `Wallet` trait or the whole module would be some sort of guide for implementing this trait

client-core/src/wallet.rs
```diff
@@ -10,21 +10,29 @@ use zeroize::Zeroize;
...
     W: Storage,
+    B: Storage,
 {
+    /// Returns assiciated Crypto.com Chain client
+    fn chain(&self) -> &C;
+
     /// Returns associated key service
     fn key_service(&self) -> &KeyService<K>;
 
     /// Returns associated wallet service
```
tomtau: As above, these comments may better be deleted, as they don't add any information, or rephrased if there is some extra information.

for example, some of these services are used inside other services, so why are they exposed individually -- if all their functionality is only used inside a different service, why are they exposed?

tomtau:
@devashishdxt btw, nightly fails: https://travis-ci.org/crypto-com/chain/jobs/513949529#L818

devashishdxt:
I have added `README.md` in `client-core` which explains the overall design of `client-core` and also the role and responsibilities of each trait and struct.

client-core/src/wallet.rs
devashishdxt: These comments are intended to be use with `cargo doc` command to generate API documentation. Also, these are exposed because these are required functions for any implementations of `Wallet` trait. Currently, `SledWallet` is the default implementation of this trait.

client-core/src/wallet.rs
devashishdxt: We should follow official rust documentation guidelines: https://rust-lang-nursery.github.io/api-guidelines/documentation.html

devashishdxt: Also, I have put a detailed explanation of each trait and struct in README.md

devashishdxt:
You can check the documentation here: https://github.com/devashishdxt/chain/tree/transaction-storage/client-core

client-core/src/wallet.rs
tomtau: Yep, the official API docs guidelines are in a similar spirit and have a convention on documenting panics etc.
I've created a task for that

client-core/src/wallet.rs
tomtau: The point of comments is to provide some additional information that's not immediately obvious from the name or types -- if `cargo doc` would complain about it, it could be suppressed with a justification comment.

devashishdxt:
Nightly build failure is because of a bug introduced in latest nightly: https://github.com/rust-lang/rust/issues/59553

This is a regression introduced by this pull request: https://github.com/rust-lang/rust/pull/59421

Centril:
> Nightly build failure is because of a bug introduced in latest nightly: [rust-lang/rust#59553](https://github.com/rust-lang/rust/issues/59553)

Note: The PR referenced *fixed* a bug introduced in some versions prior.

estebank:
@devashishdxt the new behavior is the correct behavior and used to be the active behavior until a couple of releases ago before a parser refactor introduced the incorrect behavior. We might introduce a full release cycle of warnings in beta and stable for the new behavior depending on how extensive the fallout is. If you have reasons to believe this is an incorrect course of action, please voice them in the regression issue.

devashishdxt:
@estebank I believe new behaviour is the correct behaviour. But, we’ll have to wait until changes are made in `quote` or in crates using it to fix our build.

estebank:
Quote doesn't need to change. You need to wrap your integer loop using string interpolation with an ident so that quote doesn't interpret it as a number. That will fix your build and is backwards and forwards compatible.

devashishdxt:
@estebank Our build fails because, one of our dependency, `zeroize_derive`, exposes a derive macro which internally uses `quote`. So, to fix this issue, one of them (`quote` or `zeroize_derive`) will have to make some changes. Please let me know if I’m missing something and if we can make any changes in our code to fix this.

estebank:
I see. The fix will have to happen in `zeroize_derive`.

PR #43, opened 2019-04-10
Problem: No easy way to call tendermint APIs (CRO-91)

devashishdxt:
Solution: Created Tendermint RPC Client (currently, doesn't have broadcast_tx_* commands)


client-index/src/tendermint/rpc_client.rs
```diff
@@ -0,0 +1,75 @@
...
+        self.call("block", &params)
+    }
+
+    fn block_results(&self, height: u64) -> Result<BlockResults> {
+        let params = [json!(height.to_string())];
+        self.call("block_results", &params)
+    }
+}
+
+// #[cfg(test)]
```
tomtau: For integration tests, I guess it'll be best to separate it out into `tests/...` -- you can move it there with some dummy integration test with a TODO that we'll do something automated and reproducible (e.g. it'd have a fixed Tendermint config and start abci+tendermint processes -- tracked under CRO-61)

PR #47, opened 2019-04-11
Problem: (CRO-53) No easy way to query transaction data from Chain

devashishdxt:
Solution: Created transaction index

Details:
- `Index` is the main trait which has functions to get balance, transaction history, etc.
- A default implementation of `Index` trait is provided by `RpcSledIndex` which uses `jsonrpc` for making rpc calls to tendermint and `sled` for storage
- There are other services which is used for synchronising current state of `Index` with Crypto.com Chain

### DIFF
--- a/client-common/src/balance/balance_change.rs
+++ b/client-common/src/balance/balance_change.rs
@@ -1,6 +1,7 @@
 use std::ops::Add;
 
 use failure::ResultExt;
+use rlp::{Decodable, DecoderError, Encodable, Rlp, RlpStream};
 
 use chain_core::init::coin::Coin;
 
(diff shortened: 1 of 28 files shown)


client-index/src/tendermint/types/block.rs
```diff
@@ -37,20 +37,27 @@ impl Block {
...
+        }
+    }
+}
 
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn check_transactions() {
```
tomtau: Why exactly are here these tests of `Default` implementation?

client-common/src/balance/balance_change.rs
```diff
@@ -1,20 +1,46 @@
...
-#[derive(Debug)]
+#[derive(Debug, Clone, PartialEq)]
 pub enum BalanceChange {
     /// Represents balance addition
     Incoming(Coin),
     /// Represents balance reduction
     Outgoing(Coin),
 }
 
+impl Encodable for BalanceChange {
```
tomtau: Why to RLP-encode/decode these client-internal types?
`Coin` has deserialize/serialize (as it's used in `InitConfig` which needs it for Tendermint's genesis.json) -- so these client-internal types could auto-derive serde traits and use bincode (like elsewhere in the client code)?

tomtau: TxId perhaps doesn't have it, but we may potentially add it or a client could represent it as a raw `[u8;32]` alias rather than that opaque type

tomtau: similarly address

tomtau: But there may be other reason to have RLP here rather than just that those chain-core types didn't have serde traits

devashishdxt: There's no particular reason to use RLP here. I used it because internal types don't implement `Serialize` and `Deserialize`, and also, to keep encoding and decoding consistent everywhere.

client-index/src/tendermint/types/block.rs
devashishdxt: This is testing `transactions()` with valid `Block` data. `Default` implementation just provides a valid `Block` data.

Other reason for `Default` implementation is that these are used in `MockClient`. The need of these `Default` implementations will go away once we have a way to test client with actual tendermint node (integration tests).

devashishdxt: The values in these `Default` implementations are coupled with tests in `RpcSledIndex` where we transfer some coins from one address to other address and check their balances and transaction history.

client-common/src/balance/balance_change.rs
tomtau: For "consistency", client has been mainly using bincode internally, so I assume that would be preferable.
RLP is meant for "external" binary data, what's broadcasted / stored on TM.

So if the client backend exposes any "binary" data externally, it'll make sense in RLP. If it doesn't expose that binary data (only exposes JSON), it'll be more consistent to store internally in bincode.

Anyway, I'll leave it as it is now and we can come back to it later.

PR #50, opened 2019-04-16
Problem: (CRO-53) No way to get balance in client-core

devashishdxt:
Solution: Integrated client-core with client-index


client-core/src/wallet/rpc_sled_wallet_client.rs
```diff
@@ -0,0 +1,213 @@
...
+impl WalletClient for RpcSledWalletClient {
+    fn new_wallet(&self, name: &str, passphrase: &str) -> Result<String> {
+        self.wallet_service.create(name, passphrase)
+    }
+
+    fn public_keys(&self, name: &str, passphrase: &str) -> Result<Vec<PublicKey>> {
+        let wallet_id = self.wallet_service.get(name, passphrase)?;
+
+        match wallet_id {
+            None => Err(ErrorKind::WalletNotFound.into()),
```
calvinaco: Is it possible for us to distinguish if it is incorrect passphase or missing wallet?

devashishdxt: Yes.
`get_secure()` returns `DecryptionError` when passphrase is incorrect and `None` in cases where no value corresponding to a particular key is found ([code](https://github.com/crypto-com/chain/blob/master/client-common/src/storage.rs#L87)). Here, we're converting `None` into error of type `WalletNotFound`.

PR #56, opened 2019-04-23
Problem: (CRO-56) No way to create new transaction from wallet client

devashishdxt:
Solution: Added functions to create new transactions

### DIFF
--- a/client-core/src/wallet/default_wallet_client.rs
+++ b/client-core/src/wallet/default_wallet_client.rs
@@ -1,15 +1,18 @@
 use failure::ResultExt;
+use rlp::encode;
 use zeroize::Zeroize;
 
 use chain_core::init::address::RedeemAddress;
 use chain_core::init::coin::{sum_coins, Coin};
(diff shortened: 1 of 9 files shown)


client-core/src/wallet/default_wallet_client.rs
```diff
@@ -79,6 +79,24 @@ where
...
+        &self,
+        name: &str,
+        passphrase: &str,
+        address: &ExtendedAddr,
+    ) -> Result<Option<PrivateKey>> {
+        let private_keys = self.private_keys(name, passphrase)?;
+        let addresses = self.addresses(name, passphrase)?;
+
+        for (i, known_address) in addresses.iter().enumerate() {
+            if known_address == address || cfg!(test) {
```
calvinaco: What is the purpose of `cfg!(test)` here?

devashishdxt: `cfg!` performs conditional compilation. This returns `true` when running `cargo test`. Here, I'm using it to return any private key for testing the signature.

PR #60, opened 2019-04-28
Problem: (CRO-58) No way to serialize TransactionChange in json

calvinaco:
Solution: Implemented serde traits for TransactionChange and BalanceChange


devashishdxt:
> seems ok, but not sure if `TransactionChange` is internal only or should be exposed to UI, @devashishdxt ?

Using `TransactionChange` for UI should not be a problem. `TransactionChange` is similar to a materialized view of data for an address.

PR #61, opened 2019-04-28
Problem: (CRO-99) Tendermint Client RPC does not handle null values properly

calvinaco:
### Solution:
Changed fields which has possible null values to use `Option` type

### Background:
The tendermint RPC call `block` and `block_results` sometimes return null value of transaction related fields when the block has no transaction. An example of such response are

block
```
{
    "jsonrpc": "2.0",
    "id": "block",
    "result": {
        "block_meta": ...,
        "block": {
            "header": {
                ...
                "height": "1",
                "time": "2019-04-27T06:49:34.718186Z",
                ...
            },
            "data": {
                "txs": null
            },
            ...
        }
    }
}
```

block_results
```
{
    "jsonrpc": "2.0",
    "id": "block_results",
    "result": {
        "height": "1",
        "results": {
            "DeliverTx": null,
            ...
        }
    }
}
```


calvinaco:
> LGTM sans some small fmt issues: https://travis-ci.org/crypto-com/chain/jobs/525590386#L1161
> (`cargo fmt` should fix it)

Code updated

PR #62, opened 2019-04-29
Problem: (CRO-99) jsonrpc client does not handle underlying connection reset properly

calvinaco:
### Background:
On testing the Tendermint RPC client it shows unstable request from time to time. The problem is that the `jsonrpc` crate use Hyper as the underlying HTTP client, but the crate does not handle connection reset correctly, hence causing unstable request. The issue is filed below
https://github.com/apoelstra/rust-jsonrpc/issues/26

### Solution:
Create a new client on every JSON-RPC call as a workaround

### Future:
The current solution is a workaround and is inefficient. In the future we should keep track of the `jsonrpc` issue and take appropriate response.

### DIFF
--- a/client-common/src/tendermint/rpc_client.rs
+++ b/client-common/src/tendermint/rpc_client.rs
@@ -1,7 +1,5 @@
 #![cfg(feature = "rpc")]
 
-use std::sync::Arc;
-
 use failure::ResultExt;
 use jsonrpc::client::Client as JsonRpcClient;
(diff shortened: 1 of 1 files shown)


devashishdxt:
The stable build has failed because of formatting. Try running `cargo fmt`. Also, you can turn on `Format On Save` setting on VS Code to automatically format files when you save them (RLS should be installed for this to work).

PR #63, opened 2019-04-30
Problem: cross-language client API implementation (CRO-99) (CRO-58)

calvinaco:
### Background:

Added wallet JSON-RPC client server with the following RPC calls
- wallet_create
    - Create a wallet with a new address
- wallet_addresses
    - List all addresses of a wallet
- wallet_balance
    - Return balance of a wallet
- wallet_sendtoaddress
    - Send funds from wallet to an address
- wallet_transactions
    - List all transactions of a wallet
- sync
    - Synchronize the index
- sync_all
    - Clean synchronize of the index

### How to use:

```
client-rpc --chain-id {CHAIN_ID}
```

where `{CHAIN_ID}` is the hex digit of the chain id


client-rpc/src/wallet_rpc.rs
```diff
@@ -0,0 +1,364 @@
...
+            Ok(transaction_change) => Ok(transaction_change),
+            Err(e) => Err(to_rpc_error(e)),
+        }
+    }
+}
+
+#[derive(Debug, Deserialize)]
+pub struct WalletRequest {
+    name: String,
+    passphrase: String,
```
tomtau: Is the passphrase cleared?

It'd be better to use something like https://github.com/myfreeweb/secstr

tomtau: client-core takes `&str`, so I guess this would be a bigger task -- perhaps put a TODO/WARNING comment and open a task for @devashishdxt to convert passphrase's strs to use secstrs or equivalents

calvinaco: I am looking into the `secstr`. It seems not very difficult to integrate but do require some work to change the underlying wallet client implementation. I will leave a TODO here first and open a new task for it.

tomtau:
It's probably ok, the remaining items (passphrases and configuration) could be done in different PRs -- I'll wait for @devashishdxt to comment if there's something else missing before it's merged

PR #66, opened 2019-04-30
Problem: (CRO-106) Some front end client cannot parse big number from JSON-RPC correctly

calvinaco:
Solution: Changed JSON-RPC to use string for number in `wallet_balance` and `wallet_sendtoaddress`

Limitation: `wallet_transactions` is still returning `Incoming` and `Outgoing` as number, will handle it in CRO-109 when changing its response format.


tomtau:
Closing this PR for the moment, as there were quite a few changes in the meantime.
I'm a bit worried about "string-y" typed API, so I'd suggest to discuss this more before making this change -- e.g. if changing the client returned / accepted values into a tuple `(amount / 1_0000_0000`, `amount % 1_0000_0000)` could solve this frontend parsing problem

PR #70, opened 2019-05-03
Problem: (CRO-104) Plain insecure strings for passphrases

devashishdxt:
**Solution**: Started using secstr which provides secure strings.

**Note on security**: Current implementation uses deserialization feature provided by `secstr` crate which uses `serde`. To obtain a `SecStr` object from `&str`, they copy all the bytes from `&str` provided by serde. Its not possible to zeroize `&str` provided by serde because we never get a mutable reference from serde. [here](https://github.com/myfreeweb/secstr/blob/master/src/lib.rs#L220) is the line where they make a copy of `&str`.


devashishdxt:
> looks ok -- @devashishdxt the Serde bit is only relevant to the RPC server, or elsewhere? So the original buffer with a request is owned by some `jsonrpc_http_server` internals?

Yes, `serde` is only relevant for RPC server. The original buffer should ideally be a reference to tcp socket data (only if `jsonrpc_http_server` and `serde` are not making additional copies).

PR #73, opened 2019-05-06
Problem: (CRO-59) No CLI for interacting with client backend

devashishdxt:
Solution: Added a CLI layer on client backend

### DIFF
--- a/client-cli/src/command/transaction_command.rs
+++ b/client-cli/src/command/transaction_command.rs
@@ -1,0 +1,119 @@
+use std::str::FromStr;
+
+use failure::ResultExt;
+use hex::decode;
+use quest::{ask, choose, text, yesno};
+use structopt::StructOpt;
+
+use chain_core::common::H256;
+use chain_core::init::address::RedeemAddress;
+use chain_core::init::coin::Coin;
+use chain_core::tx::data::address::ExtendedAddr;
+use chain_core::tx::data::attribute::TxAttributes;
+use chain_core::tx::data::output::TxOut;
+use client_common::{ErrorKind, Result};
+use client_core::WalletClient;
+
+use crate::ask_passphrase;
+
+#[derive(Debug, StructOpt)]
+pub enum TransactionCommand {
+    #[structopt(name = "new", about = "New transaction")]
+    New {
+        #[structopt(
+            name = "chain-id",
+            short,
+            long,
+            help = "Chain ID for transaction (Last two hex digits of chain-id)"
+        )]
+        chain_id: String,
+        #[structopt(name = "name", short, long, help = "Name of wallet")]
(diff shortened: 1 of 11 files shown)


devashishdxt:
> The `ask_passphrase` returns String
> 
> https://github.com/crypto-com/chain/blob/af98d5f993d084804d0a6bfee6ef4566287d90bc/signer-cli/src/commands/address_command.rs#L54
> 
> -- would that get cleared by the wallet client?

This doesn't use `ask_passphrase` from `signer-cli`. Here's the `ask_passphrase` definition in `main.rs`: https://github.com/crypto-com/chain/pull/73/files#diff-22eda7059b43224de055e8390df378ddR56

client-cli/src/command/transaction_command.rs
```diff
@@ -0,0 +1,119 @@
...
+use chain_core::tx::data::address::ExtendedAddr;
+use chain_core::tx::data::attribute::TxAttributes;
+use chain_core::tx::data::output::TxOut;
+use client_common::{ErrorKind, Result};
+use client_core::WalletClient;
+
+use crate::ask_passphrase;
+
+#[derive(Debug, StructOpt)]
+pub enum TransactionCommand {
```
tomtau: It's almost copy-pasted from https://github.com/crypto-com/chain/blob/master/signer-cli/src/commands/transaction_command.rs

Is there any use case for signer-* ? Would it make sense to remove signer-* and move that functionality into client-* crates (if any)?

devashishdxt:
> Some of the commands are almost identical / duplicate to what's in `signer-*` -- if it makes sense, signer-* could be removed and the functionality moved to client... e.g. if I see it correctly, the only difference between client's transaction command and signer's transaction command is that the client code broadcasts the transaction at the end, whereas the signer just prints it out in the command line.
> I think this difference could just be captured in some special debugging/development flag... something like if "print_only" (default false) is set, it won't broadcast the transaction.
> 
> For address command, I guess the only difference is that the client one is tied to a "wallet"?

It would make sense to remove `signer-*` once we have all the functionalities of `signer` in `client`. Currently, `client` only supports simple transactions whereas `signer` also has an infant version of multisig support. So, I'd vote against removing `signer` until we have multisig support in `client`.

client-cli/src/command/transaction_command.rs
devashishdxt: `signer-*` should only be removed once we have multisig support in `client`.

tomtau:
@devashishdxt Well, the functionality of `signer` can't be really used at the moment easily, because it doesn't display multisig addresses, so while the signing is present, one won't know what to put in the output.
I've opened CRO-111 epic to track this -- feel free to break it down into more issues

PR #80, opened 2019-05-14
Problem: (CRO-112) Client does not add fee during UTXO selection

devashishdxt:
Solution: Modified UTXO selection algorithm to take fee into account

---
This change is [Reviewable](https://reviewable.io/reviews/crypto-com/chain/80)

### DIFF
--- a/client-core/src/transaction_builder/default_transaction_builder.rs
+++ b/client-core/src/transaction_builder/default_transaction_builder.rs
@@ -1,0 +1,142 @@
+use failure::ResultExt;
+use secstr::SecStr;
+
+use chain_core::init::coin::Coin;
+use chain_core::tx::data::attribute::TxAttributes;
+use chain_core::tx::data::output::TxOut;
+use chain_core::tx::data::Tx;
+use chain_core::tx::fee::{Fee, FeeAlgorithm};
+use chain_core::tx::TxAux;
+use client_common::{Error, ErrorKind, Result};
+
+use crate::{TransactionBuilder, WalletClient};
+
+/// Default implementation of `TransactionBuilder`
+pub struct DefaultTransactionBuilder<F>
+where
+    F: FeeAlgorithm,
+{
+    fee_algorithm: F,
+}
+
+impl<F> DefaultTransactionBuilder<F>
+where
+    F: FeeAlgorithm,
+{
+    /// Creates a new instance of `DefaultTransactionBuilder`
+    pub fn new(fee_algorithm: F) -> Self {
+        Self { fee_algorithm }
+    }
+
+    fn build_with_fee<W: WalletClient>(
+        &self,
+        name: &str,
+        passphrase: &SecStr,
+        mut outputs: Vec<TxOut>,
+        attributes: TxAttributes,
+        wallet_client: &W,
+        fee: Fee,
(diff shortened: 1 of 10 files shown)


client-core/src/transaction_builder/default_transaction_builder.rs
```diff
@@ -0,0 +1,441 @@
...
+use chain_core::tx::data::output::TxOut;
+use chain_core::tx::data::Tx;
+use chain_core::tx::fee::{Fee, FeeAlgorithm};
+use chain_core::tx::witness::TxWitness;
+use chain_core::tx::TxAux;
+use client_common::{ErrorKind, Result};
+
+use crate::{TransactionBuilder, WalletClient};
+
+/// Default implementation of `TransactionBuilder`
```
tomtau: It'd be good to document the "default" behaviour briefly (the main idea), so that one doesn't need to dig it up from the code. I guess currently the behaviour is:
```
repeat until the fee is enough:
1. sort the UTXOs by amounts (`sort_by(|a, b| a.1.cmp(&b.1).reverse()` -- is it desc/asc? I can't tell without looking up sort_by+cmp refs)
2. take UTXOs until the target amount is reached and return a difference that's taken as the fee
``` ?

File: client-core/src/transaction_builder/default_transaction_builder.rs [hunk 2]
```diff
@@ -0,0 +1,441 @@
...
+                    value: Coin::new(30).unwrap(),
+                    valid_from: None,
+                }],
+                TxAttributes::new(171),
+                &wallet_client,
+            )
+            .expect("Unable to build transaction");
+
+        match tx_aux {
+            TxAux::TransferTx(tx, _) => assert_eq!(1, tx.outputs.len()),
```
tomtau: any assertions about the correct fee?

client-core/src/transaction_builder/default_transaction_builder.rs
```diff
@@ -0,0 +1,441 @@
...
+    fn build_with_fee<W: WalletClient>(
+        &self,
+        name: &str,
+        passphrase: &SecStr,
+        mut outputs: Vec<TxOut>,
+        attributes: TxAttributes,
+        wallet_client: &W,
+        fee: Fee,
+    ) -> Result<TxAux> {
+        let unspent_transactions = wallet_client.unspent_transactions(name, passphrase)?;
```
tomtau: does this return all unspent transactions? if so, why not to pass them as an argument instead of retrieving them and re-sorting them (`build_with_fee` is called in a while loop below https://github.com/crypto-com/chain/pull/80/files#diff-d3adbcf3f20e967015c575da2cd85dfdR158)?

devashishdxt:
> The input selection strategy could potentially also be abstracted out into a trait, so that it could be swapped out or selectable later (e.g. largest-input first, smallest-input first, random, exact match / some sort of "bin packing"). Not sure how it'd look -- I guess it could be some sort of Iterator one would query when building a TX?

It can be abstracted out, but right now, we don't have enough use cases to understand the right granularity of abstraction. It'd make sense to do this when we have a new `FeeAlgorithm` and a different way of input selection so that we have a better idea of things that we should abstract out.

PR #93, opened 2019-05-22
Problem: Client RPC requires passphrase to be a byte array

calvinaco:
Solution: Changed client RPC to accept string passphrase and transform to SecStr


tomtau:
Is that a problem to be a byte array? Points for not using String were not to accidentally log / print it and automatically zero it out when not needed

devashishdxt:
> Is that a problem to be a byte array? Points for not using String were not to accidentally log / print it and automatically zero it out when not needed

I think this is just a workaround to directly deserialise `String` from incoming `JSON` and immediately converting it into `SecStr` because `SecStr` cannot be directly deserialised from `String`. There’ll be a new type `SecUtf8` in upcoming version of `secstr` crate which would be the ideal solution for our scenario.

calvinaco:
I am going to change to use `SecUtf8` available in `SecStr` crate `0.3.2`. It will be a more proper solution to the problem. Closing this PR now.

PR #96, opened 2019-05-23
Problem: missing types and integration for account/staking-related functionality (CRO-96 / CRO-97)

tomtau:
Solution: extended TX types, added TODO stubs in pattern matches; extended storage to choose between main and account-related storage; moved to use a nightly merkle_bit crate (as the old struct wasn't Sync+Send)

--

this is more of WIP PR to add core data types before breaking changes due to InitConfig, InitChain processing, TX validation etc. come in.

### DIFF
--- a/chain-core/src/state/account.rs
+++ b/chain-core/src/state/account.rs
@@ -1,8 +1,14 @@
 use crate::common::{hash256, Timespec, HASH_SIZE_256};
 use crate::init::address::RedeemAddress;
 use crate::init::coin::Coin;
+use crate::tx::data::input::TxoPointer;
+use crate::tx::data::output::TxOut;
+use crate::tx::witness::{tree::RawSignature, EcdsaSignature};
 use blake2::Blake2s;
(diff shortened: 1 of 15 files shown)


chain-core/src/state/account.rs
```diff
@@ -51,3 +57,91 @@ impl Account {
         hash256::<Blake2s>(&self.address)
     }
 }
+
+/// attributes in account-related transactions
+#[derive(Debug, Default, PartialEq, Eq, Clone, Serialize, Deserialize, Encode, Decode)]
+pub struct AccountOpAttributes {
+    pub chain_hex_id: [u8; 1],
```
devashishdxt: Why an array of size `1` and not just `u8`?

tomtau: oops, that was a temporary laziness to get encode/decode derive

tomtau:
True, I agree, these TODOs are essential missing functionality, so would be better to mark them as such

PR #99, opened 2019-05-24
Problem: (CRO-121) No easy to use implementation of merkle tree

devashishdxt:
**Solution**: Implemented merkle tree with inclusion proofs
**Note**: I haven't integrated it with current usages of merkle tree. I wanted to get it reviewed before doing that.

### DIFF
--- a/chain-core/src/common/merkle_tree.rs
+++ b/chain-core/src/common/merkle_tree.rs
@@ -1,0 +1,742 @@
+use std::vec::IntoIter;
+
+use blake2::Blake2s;
+
+use super::{hash256, H256, H512, HASH_SIZE_256};
+
+// txid_hash(&vec![])
+const EMPTY_HASH: H256 = [
+    105, 33, 122, 48, 121, 144, 128, 148, 225, 17, 33, 208, 66, 53, 74, 124, 31, 85, 182, 72, 44,
+    161, 165, 30, 27, 37, 13, 253, 30, 208, 238, 249,
+];
+
+/// Represents inner tree structure of Merkle Tree
+#[derive(Debug)]
+pub enum Tree<T> {
+    /// Empty Node
+    Empty,
+    /// Leaf Node
+    Leaf { hash: H256, value: T },
+    /// Middle Node
+    Node {
+        hash: H256,
+        left: Box<Tree<T>>,
+        right: Box<Tree<T>>,
+    },
+}
+
+impl<T> Tree<T> {
+    /// Creates a tree with one empty node
+    #[inline]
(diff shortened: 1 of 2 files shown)


chain-core/src/common/merkle_tree.rs
```diff
@@ -0,0 +1,643 @@
...
+    where
+        T: AsRef<[u8]> + Clone,
+    {
+        self.tree.generate_path(value)
+    }
+}
+
+#[inline]
+fn combine(left: &H256, right: &H256) -> H512 {
+    let mut hash = [0; HASH_SIZE_256 * 2];
```
tomtau: probably not a useful attack in the scenarios how Merkle trees are used here, but it wouldn't hurt to have a fix
https://en.wikipedia.org/wiki/Merkle_tree#Second_preimage_attack
https://flawed.net.nz/2018/02/21/attacking-merkle-trees-with-a-second-preimage-attack/

I guess two ways to fix it:
1) if sparse, then the depth could be stored alongside the root hash and the proof verification would check the proof length against the expected one
2) as suggested in the links: not to construct tree from hashes, but from some hashable / encodable items; prefix encoded leaf items with 0 before hashing them as leafs, intermediate nodes with 1

chain-core/src/common/merkle_tree.rs
```diff
@@ -0,0 +1,643 @@
...
+use super::{hash256, H256, H512, HASH_SIZE_256};
+
+// hash(&vec![])
+const EMPTY_HASH: H256 = [
+    105, 33, 122, 48, 121, 144, 128, 148, 225, 17, 33, 208, 66, 53, 74, 124, 31, 85, 182, 72, 44,
+    161, 165, 30, 27, 37, 13, 253, 30, 208, 238, 249,
+];
+
+/// Represents inner tree structure of Merkle Tree
+#[derive(Debug)]
```
tomtau: some binary serialization / encode would be needed

devashishdxt: This'll part of future PRs (when I start integrating this merkle tree with existing usages).

chain-core/src/common/merkle_tree.rs
devashishdxt: Implemented solution by prefixing `0` (for leaf nodes) or `1` (for intermediate nodes) before hashing.

devashishdxt:
> looks ok -- @devashishdxt would you want to merge it now or after the current uses are ported to it?

I don't see any problem in doing that. By merging this, we'll have multiple small PRs instead of having a huge one.

tomtau:
ok

PR #104, opened 2019-05-27
Problem: (CRO-121) Still using old merkle tree implementation

devashishdxt:
Solution: Changed all the usages from old to new merkle tree implementation

---
This change is [Reviewable](https://reviewable.io/reviews/crypto-com/chain/104)


chain-core/src/common/merkle_tree.rs
```diff
@@ -190,7 +160,7 @@ impl Path {
 }
 
 /// Inclusion proof of a value
-#[derive(Debug, PartialEq)]
+#[derive(Debug, PartialEq, Eq, Clone, Encode, Decode)]
```
tomtau: perhaps a TODO note (that was in the original) that the encoding may not be that efficient, so a custom one with BitVec+vec of hashes could be considered

PR #110, opened 2019-05-30
Problem: (CRO-119) No support for creating multi-sig transactions in client

devashishdxt:
**Solution**: 
- Added basic support for creating multi-sig transactions
- This PR also removes `signer-*` crates.

---
This change is [Reviewable](https://reviewable.io/reviews/crypto-com/chain/110)


devashishdxt:
> 1. more robust / misuse-resistant API for the signing session protocol (one can't call internal operations that are unsafe or guaranteed to fail)

I don't think we should implement our own state machine on top of already existing state machine implemented by underlying C library to check validity of all the nonces. The only problem that I could see in current implementation is that one can modify existing nonce commitments and nonces after setting them and then reuse the same session with different values. I've added a restriction that each value can only be set once for each signer. All other validations and failures are delegated to underlying C library.

devashishdxt:
> 2\. "simple schnorr" -- currently for leafs, it always runs the n-of-n multi-stage signing session even if n=1. I'm not sure if it still works with n=1, but even if it does, it's an unnecessary complexity to go through these stages -- one could just generate the inclusion proof and call `schnorr_sign` without creating co-signing sessions.

Simple schnorr is a totally different flow and will be done as part of a separate task.

PR #117, opened 2019-06-03
Problem: (CRO-127) No support for simple schnorr signatures for 1-of-n transactions

devashishdxt:
Solution: Added support for simple schnorr signatures for 1-of-n transactions

---
This change is [Reviewable](https://reviewable.io/reviews/crypto-com/chain/117)

### DIFF
--- a/client-core/src/key/private_key.rs
+++ b/client-core/src/key/private_key.rs
@@ -1,6 +1,7 @@
 use failure::ResultExt;
 use parity_codec::{Decode, Encode, Input, Output};
 use rand::rngs::OsRng;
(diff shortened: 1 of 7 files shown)


devashishdxt:
> @devashishdxt looks good -- just a question: the multi-sig isn't integrated with transaction builder parts yet, right?
> For 1-of-n, it should fit with what's there more or less (depending on the unspent output's type, one would either call `sign` or `schnorr_sign`+ generate inclusion proof)

@tomtau Yes. We should be able to integrate current transaction builder with 1-of-n schnorr signatures. I’ll take that up next.

PR #119, opened 2019-06-03
(CRO-125) Change Client RPC to use numeric string as amount [NFM][NEED_REVIEW]

calvinlauco:
This is a draft PR to try to introduce `SerializedCoin` which survives only in `client-*`. Test cases haven't passed but the skeleton is there.

### `SerializedCoin`
- Designed to be a data structure wrapping `Coin`
- Responsible only to do serialization, deserialization and conversion between string, u64 and Coin.

### This PR is for draft review only
I realized on the way of changing everything in `client-*` to use `SerializedCoin` that, some places are relying on `Coin` to do computation. They require conversion between different types, making the code looks clumsy.

But at the same time, I don't think it is a good idea to implement again the arithmetic operation on `SerializedCoin`. So I am considering to use @tomtau  suggestion to change the `Coin` struct directly by changing its serialization and deserialization is even better. @tomtau @devashishdxt more comment is welcomed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -388,6 +388,7 @@ dependencies = [
  "jsonrpc-http-server 11.0.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "secstr 0.3.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde 1.0.91 (registry+https://github.com/rust-lang/crates.io-index)",
+ "serde_json 1.0.39 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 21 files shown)


tomtau:
@calvinlauco since the "on-chain" binary data operations were separated into Encode/Decode independent of serde, changes in serde-impls have a limited scope, so I think it's fine to do.
Non-string JSON Coin is currently present in genesis.json (one will just need to regenerate these) that's passed in RequestInitChain message from ABCI... and then it's "dumped" in JSON as a part of app state into ResponseInfo's data field (but I think client ignored the data field -- @devashishdxt  ?)

PR #128, opened 2019-06-05
Problem: (CRO-128) Unspent transaction selection cannot be configured

devashishdxt:
**Solution**: 
- Implemented decorators for unspent transactions to make unspent transaction selection configurable.
- This PR does not add any new functionality. It only modifies the structure of code to ease future development.


client-core/src/unspent_transactions.rs
```diff
@@ -0,0 +1,216 @@
...
+/// let mut unspent_transactions = UnspentTransactions::new(unspent_transactions);
+///
+/// // This will filter and sort unspent transactions: only transactions with
+/// // redeem addresses and in decreasing order of their value.
+/// unspent_transactions
+///     .decorate_with(OnlyRedeemAddresses)
+///     .decorate_with(HighestValueFirst);
+/// ```
+#[derive(Debug)]
+pub struct UnspentTransactions {
```
tomtau: I may be missing some future use case, but could you clarify why this / what cannot be done with plain iterators and operations over them?

e.g. one could have functions:
```
pub fn is_redeem_address(utxo: (TxoPointer, TxOut)) -> bool {
...
}

pub fn is_tree_address(utxo: (TxoPointer, TxOut)) -> bool {
...
}

pub fn higher_value_first<I>(iterable: I) -> IntoIter<I::Item> 
where
    I: IntoIterator,
    I::Item: (TxoPointer, TxOut), {
...
}
```

and then could do something like:
```
let unspent_transactions: Vec<(TxoPointer, TxOut)> = ...;
...
higher_value_first(unspent_transactions.iter().filter(is_redeem_address))
```

I think this may lead to less code, as one doesn't need to list out and implement these operations in this decorator wrapper, plus should be more idiomatic Rust / helpful to the compiler (could try to do loop fusion etc.)

devashishdxt: A configurable pattern is important because user will be able to give all the configurations as input and we'll filter and sort based on user's input. So, the filtering and sorting conditions here are dynamic and cannot be done statically using iterators. To do it dynamically using iterators, we need an enum with different type of conditions supported and based on the input provided by user, execute those conditions (this is exactly what is done right now).
 
Any thoughts?

tomtau: Oh, so is the idea that `Decorator` is going to be serialisable (input from RPC, cli...)?

devashishdxt: That depends on how we want our user experience to be. One options is, as you suggested, serialising and deserialising `Decorator`. There can also be other ways to do this.

devashishdxt: The idea is to keep backend fully configurable so that it is not limiting UX in any way. We can decide to keep this static through UX but it'd be easy to change it afterwards.

tomtau: My thoughts would be:

* if the Decorator is internal calls only, then I'm not sure why it is beneficial over plain iterators -- input would be something like `filter_condition: FnMut(&(TxoPointer, TxOut)) -> bool` and `sort_condition: ...` or `sorting_function: ...`

* If it's for external input, then it may be worth splitting into filter and sort parts, as `utxo_selection_conditions: Vec<Decorator>` would be too general / not always meaningful

devashishdxt: > * if the Decorator is internal calls only, then I'm not sure why it is beneficial over plain iterators -- input would be something like `filter_condition: FnMut(&(TxoPointer, TxOut)) -> bool` and `sort_condition: ...` or `sorting_function: ...`

`Decorator` is just an `enum` and each enum variant denotes an existing `Fn`. We're still using iterators internally. One point for not using `Fn` pattern is that it'll make it too tedious to expose RPC APIs with different configurations.

devashishdxt: > * If it's for external input, then it may be worth splitting into filter and sort parts, as `utxo_selection_conditions: Vec<Decorator>` would be too general / not always meaningful

Agree, will separate sorting and filtering operations.

tomtau: But for end-user RPC/CLI, I'm not sure whether this fine-grained sorting or filter conditions are needed -- I guess there'll just be some default option and a way to select from a few

devashishdxt: > But for end-user RPC/CLI, I'm not sure whether this fine-grained sorting or filter conditions are needed -- I guess there'll just be some default option and a way to select from a few

In that case, providing an API with `Fn` as input does not sound right. Current API provides enough configurability without exposing too many details about internals.

PR #133, opened 2019-06-10
Problem: latest parity/scale-codec failed to compile (CRO-161) #120

leejw51crypto:
Solution: converted usize fields (in datatypes that use scale-codec) to u64

---
This change is [Reviewable](https://reviewable.io/reviews/crypto-com/chain/133)

### DIFF
--- a/chain-core/src/tx/data/input.rs
+++ b/chain-core/src/tx/data/input.rs
@@ -13,7 +13,7 @@ use crate::tx::data::TxId;
 )]
 pub struct TxoPointer {
     pub id: TxId,
-    pub index: usize,
+    pub index: u64,
 }
 
 impl fmt::Display for TxoPointer {
@@ -25,6 +25,9 @@ impl fmt::Display for TxoPointer {
 impl TxoPointer {
     /// Constructs a new TX input (mainly for testing/tools).
     pub fn new(id: TxId, index: usize) -> Self {
-        TxoPointer { id, index }
+        TxoPointer {
+            id,
+            index: index as u64,
+        }
     }
 }
--- a/client-core/src/service/root_hash_service.rs
+++ b/client-core/src/service/root_hash_service.rs
@@ -4,7 +4,7 @@ use secstr::SecUtf8;
 
 use chain_core::common::{MerkleTree, Proof, H256};
 use chain_core::tx::witness::tree::RawPubkey;
-use client_common::{ErrorKind, Result, SecureStorage, Storage};
+use client_common::{Error, ErrorKind, Result, SecureStorage, Storage};
 
 use crate::PublicKey;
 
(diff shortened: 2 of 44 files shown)


client-core/src/service/root_hash_service.rs
```diff
@@ -52,7 +52,7 @@ where
         let merkle_tree = MerkleTree::new(combinations);
         let root_hash = merkle_tree.root_hash();
 
-        let multi_sig_address = MultiSigAddress { m, n, merkle_tree };
+        let multi_sig_address = MultiSigAddress {m: m as u64, n:n as u64, merkle_tree };
```
devashishdxt: Why not take `u64` in arguments instead of casting `usize` to `u64`?

client-core/src/service/root_hash_service.rs
```diff
@@ -100,7 +100,7 @@ where
             Some(address_bytes) => {
                 let address = MultiSigAddress::decode(&mut address_bytes.as_slice())
                     .ok_or_else(|| client_common::Error::from(ErrorKind::DeserializationError))?;
-                Ok(address.m)
+                Ok(address.m as usize)
```
devashishdxt: This function should return `u64`.

client-core/src/transaction_builder/default_transaction_builder.rs
```diff
@@ -104,7 +104,7 @@ where
         let mut witnesses = Vec::with_capacity(transaction.inputs.len());
 
         for input in &transaction.inputs {
-            let input = wallet_client.output(&input.id, input.index)?;
+            let input = wallet_client.output(&input.id, input.index as usize)?;
```
devashishdxt: Change the `output()` function to take `u64` instead?

leejw51crypto:
> A few comments:
> 
> * There's a compilation error in tests. You'll have to change `usize` to `u64` in tests as well.
> * Run `cargo fmt` and `cargo clippy` before pushing your code. This will properly format your code and also warn you about non-idiomatic Rust code. Current CI has lint and formatting checks in place so, build won't succeed without properly formatted code.

thanks
i'll fix today

chain-core/src/tx/data/input.rs
```diff
@@ -13,7 +13,7 @@ use crate::tx::data::TxId;
 )]
 pub struct TxoPointer {
     pub id: TxId,
-    pub index: usize,
+    pub index: u64,
```
tomtau: here and `TxAccess`, perhaps add a TODO comment:
```
// TODO: u16 and Vec size check in Decode implementation
```

chain-core/src/tx/fee/mod.rs
```diff
@@ -67,7 +67,7 @@ impl Milli {
 #[derive(Debug)]
 pub enum MilliError {
     /// An invalid length of parts (should be 2)
-    InvalidPartsLength(usize),
+    InvalidPartsLength(u64),
```
tomtau: I think this one isn't necessary, as it's not using the parity-scale-codec?

chain-core/src/init/address.rs
```diff
@@ -51,7 +51,7 @@ where
 #[derive(Debug)]
 pub enum ErrorAddress {
     /// An invalid length
-    InvalidLength(usize),
+    InvalidLength(u64),
```
tomtau: I think this one isn't necessary (not implementing parity scale codec)

leejw51crypto: done

client-core/src/service/root_hash_service.rs
leejw51crypto: done

client-core/src/service/root_hash_service.rs
leejw51crypto: done

chain-core/src/tx/data/input.rs
leejw51crypto: done

client-core/src/transaction_builder/default_transaction_builder.rs
leejw51crypto: currently, RUST iterator index is also usize,
also other output functions are in usize,
so it seems ok as usize

devashishdxt:
LGTM. Wait for @tomtau to approve before merging.

devashishdxt:
Since we are making this change only to support encoding and decoding of `usize`, will it make more sense just to implement `Encode` and `Decode` for types containing `usize` instead of changing `usize` to `u64` everywhere? @tomtau @leejw51crypto any thoughts?

tomtau:
> Since we are making this change only to support encoding and decoding of `usize`, will it make more sense just to implement `Encode` and `Decode` for types containing `usize` instead of changing `usize` to `u64` everywhere? @tomtau @leejw51crypto any thoughts?

do you mean to keep `usize` and have a manual implementation of `Encode` and `Decode`?

devashishdxt:
> do you mean to keep `usize` and have a manual implementation of `Encode` and `Decode`?

Yes. This will also help with error handling and avoiding integer overflows. Since Rust does not perform any overflow checks in release mode, I think manually implementing `Encode` and `Decode` with overflow checks for these special cases will give us more control and safety as compared to casting `u64` to `usize` at multiple places.

tomtau:
> > do you mean to keep `usize` and have a manual implementation of `Encode` and `Decode`?
> 
> Yes. This will also help with error handling and avoiding integer overflows. Since Rust does not perform any overflow checks in release mode, I think manually implementing `Encode` and `Decode` with overflow checks for these special cases will give us more control and safety as compared to casting `u64` to `usize` at multiple places.

There were two cases where `usize` was used:

* unnecessary (e.g. nonce): here, it's fine to change to u64
* lengths / indices: here, we'll want manual `Encode`/`Decode` with size checking

For the latter one, it'll require more thought case-by-case (or datatype-by-datatype), so I'll delay that for now -- the simplest solution for the problem here ("latest parity/scale-codec failed to compile") is to keep using auto-derived implementations.
Later PRs could then fix individual problems that this may cause ("unreasonably big tx input/output length", "Merkle Tree length may overflow on 32-bit archs"...)

leejw51crypto:
ok, got it

tomtau:
seems upstream merge attempt brought back old code? https://github.com/crypto-com/chain/pull/133/files#diff-38d25694877d43825df8be4ab3c0a683
@devashishdxt @leejw51crypto

leejw51crypto:
merged with latest master

PR #134, opened 2019-06-11
Problem: (CRO-128) 1-of-n schnorr signatures are not supported by transaction builder

devashishdxt:
**Solution**: 
- Added support for 1-of-n schnorr signatures in transaction builder.
- Divided transaction builder into three components:
  - `Signer`: For transaction signing
  - `UnspentTransactions`: For UTXO selection
  - `TransactionBuilder`: For building transaction
- There is also a slight change in algorithm used for creating transactions:
  - We now no longer have a `DummySigner` which signs a transaction with dummy values for calculating fees. (I'll have to give more thought on implementing a `DummySigner` because 1-of-n signatures consist of a schnorr signature and a merkle proof which can be of variable size. So, we cannot deterministically estimate the size and structure of a 1-of-n signature)

---
This change is [Reviewable](https://reviewable.io/reviews/crypto-com/chain/134)

### DIFF
--- a/client-cli/src/command.rs
+++ b/client-cli/src/command.rs
@@ -11,6 +11,7 @@ use client_common::balance::BalanceChange;
 use client_common::storage::SledStorage;
 use client_common::tendermint::{Client, RpcClient};
 use client_common::Result;
+use client_core::signer::DefaultSigner;
 use client_core::transaction_builder::DefaultTransactionBuilder;
 use client_core::wallet::{DefaultWalletClient, WalletClient};
 use client_index::index::DefaultIndex;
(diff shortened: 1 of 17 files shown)


devashishdxt:
> as this is a "high-level" wallet API, I'm thinking it may be more client code friendly to have some optional restricted choice, e.g.:
> 
> ```
> input_selection: Option<InputSelectionStrategy>,
> ```
> 
> where if `None`, it'll use some default operations.
> if` Some`, then `InputSelectionStrategy` will be some enum that will map to one or more operations

What do you think should be different input selection strategies?

devashishdxt:
> Strategy 1 (default): what it has been doing before (highest value first?)
> Strategy 2: random
> Strategy 3: highest value first
> Strategy 4: lowest value first
> 
> @calvinco any thoughts?

I've added `InputSelectionStrategy` in external API with three variants:
- `HighestValueFirst`: Default
- `LowestValueFirst`
- `Random`

Note: Currently, these strategies does not group transactions by their address type (i.e. unspent inputs with redeem addresses are not given priority over the ones with 1-on-n addresses. Only criteria for sorting is their value).

Open Question: Should we prioritise redeem addresses over 1-of-n addresses?

PR #140, opened 2019-06-13
Problem: Coin edge case overflow panics (CRO-168)

leejw51crypto:
when overflow occurs, it goes into Coin:new 
instead of panic

---
This change is [Reviewable](https://reviewable.io/reviews/crypto-com/chain/140)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -280,6 +280,7 @@ dependencies = [
  "serde 1.0.92 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde_json 1.0.39 (registry+https://github.com/rust-lang/crates.io-index)",
  "sha3 0.8.2 (registry+https://github.com/rust-lang/crates.io-index)",
+ "static_assertions 0.3.3 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
@@ -1958,6 +1959,11 @@ version = "3.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
+name = "static_assertions"
+version = "0.3.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
 name = "stream-cipher"
 version = "0.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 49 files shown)


devashishdxt:
> @leejw51crypto @devashishdxt does this solve any problem?
> Currently, one _cannot construct a `Coin`_ (from calling `new`, JSON serde deserialisation or Parity Codec binary decoding) where the value is larger than `MAX_COIN`.
> 
> `MAX_COIN == 100_000_000_000__0000_0000`
> 
> `u64::MAX == 184_467_440_737__0955_1615`
> 
> so, the upper machine bound is almost twice as what's enforced in Coin, so the only edge case (which is extremely unlikely from the real coin distribution) I can think of is that both values would be very close to `MAX_COIN`

Even though this case is highly unlikely, we don't have to manually do `checked` operations if we use
```
[profile.release]
overflow-checks = true
```
in our `Cargo.toml`. The only place this won't check for overflows is when we use `as` keyword for typecasting, which, ideally shouldn't be used in security critical applications.

[Example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=8f8ca56cae8e66869f548eddb6fc2a34)

tomtau:
@devashishdxt the difference is, I guess, that the Cargo configuration would panic, whereas the explicit `checked_add` in this patch would return an error value.

I think panic is fine for this extreme edge case -- @devashishdxt @leejw51 ?

devashishdxt:
> @devashishdxt the difference is, I guess, that the Cargo configuration would panic, whereas the explicit `checked_add` in this patch would return an error value.
> 
> I think panic is fine for this extreme edge case -- @devashishdxt @leejw51 ?

I think panicking in this scenario is fine because, this means that someone is able to make two instances of `Coin` which exceed our total supply. 
I believe this is a good enough reason to panic and stop execution immediately.

tomtau:
> > @devashishdxt the difference is, I guess, that the Cargo configuration would panic, whereas the explicit `checked_add` in this patch would return an error value.
> > I think panic is fine for this extreme edge case -- @devashishdxt @leejw51 ?
> 
> I think panicking in this scenario is fine, because, this means that someone is able to make two instances of `Coin` which exceed our total supply.
> I believe this is a good enough reason to panic and stop execution immediately.

@devashishdxt panic is definitely desired / failsafe, but I'm now thinking whether this could lead to DoS -- e.g. I'll create a seemingly valid transaction (spends valid unspent inputs etc.), but I'll put two outputs with `Coin::max` ... then it'll crash any node that tries to validate it?

@leejw51crypto could you perhaps add a test case that panicked in the old code and is now safe?

leejw51crypto:
> @devashishdxt the difference is, I guess, that the Cargo configuration would panic, whereas the explicit `checked_add` in this patch would return an error value.
> 
> I think panic is fine for this extreme edge case -- @devashishdxt @leejw51 ?

yes,
panic is much better than processing in wrong way

PR #142, opened 2019-06-13
Problem: (CRO-132) No way to create transactions to withdraw unbonded amount from account

devashishdxt:
**Solution**: 
- Created `NetworkOpsClient` with a default implementation to create and sign transaction to withdraw unbonded amount.
- In future, all the functions related to network operations on Chain will be added to `NetworkOpsClient`.

NOTE: Currently, I've hard-coded value of nonce to be zero. This will change in future once we have a way to query nonce value from Chain.

---
This change is [Reviewable](https://reviewable.io/reviews/crypto-com/chain/142)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -379,6 +379,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "client-network"
+version = "0.1.0"
+dependencies = [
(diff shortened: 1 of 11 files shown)


devashishdxt:
> `client-core` instead of `chain-core`?

Fixed.

PR #144, opened 2019-06-14
Problem: coin overflow check (CRO-168)

leejw51crypto:
Solution: add unit test using quickcheck

### DIFF
--- a/chain-core/src/init/coin.rs
+++ b/chain-core/src/init/coin.rs
@@ -241,6 +241,39 @@ mod test {
     use quickcheck::quickcheck;
 
     quickcheck! {
+        // test whether oveflow error occur
+        fn coin_add() -> bool {
+            let a = Coin::max();
+            let b = Coin::max();
+            let sum = a+ b;
+            sum.is_err()
+        }
+
+        // test whether overflow error not occur
+          fn coin_add2() -> bool {
+            let a = Coin::max();
+            let b = Coin::new(0).unwrap();
+            let sum = (a+ b).unwrap();
+            sum == a
+        }
+
+        // test whether underflow error occur
+        fn coin_sub() -> bool {
+            let a = Coin::new(0).unwrap();
+            let b = Coin::max();
+            let sub = a- b;
+            sub.is_err()
+        }
+
+        // test whether underflow error not occur
+        fn coin_sub2() -> bool {
+            let a = Coin::max();
+            let b = Coin::new(0).unwrap();
+            let sub = (a-b).unwrap();
+            sub == a
+        }
+
+
         // test a given u32 is always a valid value for a `Coin`
         fn coin_from_u32_always_valid(v: u32) -> bool {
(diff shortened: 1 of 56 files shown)


leejw51crypto:
add unit tests

chain-core/src/init/coin.rs
```diff
@@ -241,6 +241,39 @@ mod test {
     use quickcheck::quickcheck;
 
     quickcheck! {
+        // test whether oveflow error occur
```
tomtau: Quickcheck is meant for property tests where the argument to the test functions are fed with random inputs, so it's a bit confusing / redundant to have unit tests here.

it'll be better to move these up (from the quickcheck macro) as regular units in Rust:
```
    #[test]
    fn coin_overflow_add() {
    ...
    assert!(sum.is_err())
    }
```

and perhaps give the more meaningful names than `coin_add`, `coin_add2`...

For example: `add_2_coin_max_should_produce_error`

chain-core/src/init/coin.rs
```diff
@@ -241,6 +241,39 @@ mod test {
...
+            let a = Coin::max();
+            let b = Coin::max();
+            let sum = a+ b;
+            sum.is_err()
+        }
+
+        // test whether overflow error not occur
+          fn coin_add2() -> bool {
+            let a = Coin::max();
+            let b = Coin::new(0).unwrap();
```
tomtau: could this actually cause overflows?
The "lowest" overflowing addition is: Coin::max() + Coin::new(84_467_440_737__0955_1616), I think?

chain-core/src/init/coin.rs
```diff
@@ -241,6 +241,39 @@ mod test {
...
+            let a = Coin::new(0).unwrap();
+            let b = Coin::max();
+            let sub = a- b;
+            sub.is_err()
+        }
+
+        // test whether underflow error not occur
+        fn coin_sub2() -> bool {
+            let a = Coin::max();
+            let b = Coin::new(0).unwrap();
```
tomtau: could this cause underflow?

chain-core/src/init/coin.rs
leejw51crypto: this is not overflow.
this is assert for "max + 0" works fine

chain-core/src/init/coin.rs
leejw51crypto: done

chain-core/src/init/coin.rs
leejw51crypto: no

chain-core/src/init/coin.rs
tomtau: ok, you may want to add a test case with `Coin::new(84_467_440_737__0955_1616)`

chain-core/src/init/coin.rs
tomtau: so perhaps this test is not necessary?

PR #151, opened 2019-06-17
Problem: (CRO-132) New account/staking details are not handled properly in client-index

devashishdxt:
Solution: Added support for all the types of transactions (payment and network ops) in client-index

---
This change is [Reviewable](https://reviewable.io/reviews/crypto-com/chain/151)

### DIFF
--- a/client-rpc/src/wallet_rpc.rs
+++ b/client-rpc/src/wallet_rpc.rs
@@ -198,7 +198,7 @@ mod tests {
     use chain_core::tx::TxAux;
     use client_common::balance::{BalanceChange, TransactionChange};
     use client_common::storage::MemoryStorage;
-    use client_common::{Error, ErrorKind, Result};
+    use client_common::{Error, ErrorKind, Result, Transaction};
     use client_core::signer::DefaultSigner;
     use client_core::transaction_builder::DefaultTransactionBuilder;
     use client_core::wallet::DefaultWalletClient;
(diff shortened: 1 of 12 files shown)


client-rpc/src/wallet_rpc.rs
```diff
@@ -237,15 +237,15 @@ mod tests {
             Ok(Vec::new())
         }
 
-        fn transaction(&self, _: &TxId) -> Result<Option<Tx>> {
-            Ok(Some(Tx {
+        fn transaction(&self, _: &TxId) -> Result<Option<Transaction>> {
+            Ok(Some(Transaction::TransferTransaction(Tx {
```
leejw51crypto: how about just using Result<Tx>

tomtau: It's already using `Result<Option<...>>`?

leejw51crypto: Result with Option seems redundant

devashishdxt: This function is mainly used by UI. @calvinaco any comments?

tomtau: @leejw51crypto it depends on the semantics -- it could be that `Ok(None)` indicates there wasn't any transaction found, but everything else was OK (connection was OK, TXID was a correct length, I/O was OK...)

tomtau: But it could be that one of the errors could be "TransactionNotFound", and the whole return type could be `Result<Transaction>` instead... probably simpler -- perhaps in a different PR?

PR #167, opened 2019-06-28
Problem: add current tx nonce query for the current staked state in network operation tx generation(CRO-176)

leejw51crypto:
add get nonce to default network ops client
add get account to rpc client


leejw51crypto:
because client-common uses chain-core,  rpc client support high level api ,  rpc client can use api easily.

> overall it's ok, but there are a couple of things to finalize:
> 
>     * [ ]  change the RPC client's API (I'm not sure if it should be more high-level or low-level: @leejw51crypto @devashishdxt , any thoughts?)
> 
>     * [ ]  a datatype for abci_query results instead of manual parsing
> 
>     * [ ]  mock rpc client in tests?



> overall it's ok, but there are a couple of things to finalize:
> 
>     * [ ]  change the RPC client's API (I'm not sure if it should be more high-level or low-level: @leejw51crypto @devashishdxt , any thoughts?)
> 
>     * [ ]  a datatype for abci_query results instead of manual parsing
> 
>     * [ ]  mock rpc client in tests?



> overall it's ok, but there are a couple of things to finalize:
> 
>     * [ ]  change the RPC client's API (I'm not sure if it should be more high-level or low-level: @leejw51crypto @devashishdxt , any thoughts?)
> 
>     * [ ]  a datatype for abci_query results instead of manual parsing
> 
>     * [ ]  mock rpc client in tests?

client-common/src/tendermint/client.rs
```diff
@@ -17,4 +18,7 @@ pub trait Client: Send + Sync {
 
     /// Makes `broadcast_tx_sync` call to tendermint
     fn broadcast_transaction(&self, transaction: &[u8]) -> Result<()>;
+
+    /// Get staked-state from the staked stake address
+    fn get_account(&self, staked_state_address: &[u8]) -> Result<StakedState>;
```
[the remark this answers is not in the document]
leejw51crypto: because client-common uses chain-core, rpc client support high level api , rpc client can use api easily.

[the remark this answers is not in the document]
tomtau: but say `broadcast_transaction` doesn't take `&TxAux`, but takes `&[u8]` -- any thoughts @devashishdxt ?

client-network/src/network_ops/default_network_ops_client.rs
```diff
@@ -147,6 +180,7 @@ mod tests {
     fn check_create_deposit_bonded_stake_transaction() {
         let name = "name";
         let passphrase = &SecUtf8::from("passphrase");
+        let tendermint_url = "http://localhost:26657/";
```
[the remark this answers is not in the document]
leejw51crypto: MockClient is used in client-index, i'll use that.

client-index/src/index/default_index.rs
```diff
@@ -469,6 +470,16 @@ mod tests {
         fn broadcast_transaction(&self, _: &[u8]) -> Result<()> {
             Ok(())
         }
+
+        fn get_account(&self, staked_state_address: &[u8]) -> Result<StakedState> {
+            Ok(StakedState::new(
```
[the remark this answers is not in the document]
leejw51crypto: yes that's a mock

client-common/src/tendermint/client.rs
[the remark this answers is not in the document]
devashishdxt: `broadcast_transaction` takes a byte array because tendermint RPC expects a byte array. This interface is only to wrap tendermint RPC calls in Rust functions and is not intended to perform any domain specific things. I think, `fn query(&self, data: &[u8], path: &str) -> Result<QueryResult>;` (as suggested by @tomtau) is the right approach here. All other domain specific things can be handled in `NetworkOpsClient` implementation.

client-common/src/tendermint/rpc_client.rs
```diff
@@ -65,4 +65,22 @@ impl Client for RpcClient {
...
             .map(|_| ())
     }
+
+    fn query(&self, path: &str, data: &str) -> Result<QueryResult> {
+        // path, data, height, prove
+        let params = [json!(path), json!(data), json!(null), json!(null)];
+        self.call("abci_query", &params)
+    }
+    fn get_account(&self, staked_state_address: &[u8]) -> Result<StakedState> {
+        self.query("account", hex::encode(staked_state_address).as_str())
```
devashishdxt: As mentioned above, I don't think this implementation belongs in `RpcClient`. We can move all the custom login in `NetworkOpsClient` implementation. Any thoughts @tomtau?

client-network/Cargo.toml
```diff
@@ -12,3 +12,9 @@ chain-tx-validation = { path = "../chain-tx-validation" }
 client-common = { path = "../client-common" }
 client-core = { path = "../client-core" }
 secstr = "0.3.2"
+base64 = "0.10"
+chrono = { version = "0.4", features = ["serde"] }
+parity-codec = { features = ["derive"], version = "4.1.1" }
+
+[dev-dependencies]
+secp256k1zkp = { git = "https://github.com/crypto-com/rust-secp256k1-zkp.git", rev = "9dce794d1d8ce03ffa48823acf198fae15e896d3", features = ["serde", "zeroize", "rand", "recovery", "endomorphism"] }
```
devashishdxt: Change the commit hash to latest for this repository.

client-network/src/network_ops.rs
```diff
@@ -32,6 +43,7 @@ pub trait NetworkOpsClient {
         from_address: &ExtendedAddr,
         value: Coin,
         attributes: StakedStateOpAttributes,
+        nonce: Nonce,
```
devashishdxt: I don't know if we want to take `nonce` as input or calculate it while creating a transaction (using `get_staked_state_account()`). Any thoughts @tomtau?

client-network/src/network_ops.rs
```diff
@@ -15,6 +17,15 @@ use secstr::SecUtf8;
...
 /// Interface for performing network operations on Crypto.com Chain
 pub trait NetworkOpsClient {
+    /// fetch current StakedState
+    fn get_staked_state_account(
+        &self,
+        to_staked_account: StakedStateAddress,
+    ) -> Result<StakedState>;
+
+    /// fetch current StakedState nonce
+    fn get_staked_state_nonce(&self, to_staked_account: StakedStateAddress) -> Result<Nonce>;
```
devashishdxt: I don't think this function adds any value. `get_staked_state_account()` already returns `StakedState` object which contains nonce along with other details.

client-common/src/tendermint/client.rs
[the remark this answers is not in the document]
leejw51crypto: agree, i will move get_account to network ops client.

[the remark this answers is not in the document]
leejw51crypto: let's keep rpc client only related to tendermint

client-network/src/network_ops.rs
leejw51crypto: nonce is fetched from core.
so it can be contained, but api has two round trips from core.

client-network/src/network_ops.rs
leejw51crypto: get_staked_state_nonce is for convenience.
what do you think?

devashishdxt: If this is a convenience method, I don't think it belongs in the interface. Also, I think below code is perfectly reasonable.
```
let nonce = client.get_staked_state_account()?.nonce;
```
So, I don't think we're adding any value by creating a new function just to avoid accessing field using `.nonce`.

client-network/src/network_ops.rs
devashishdxt: Why will it have two round trips?

client-network/Cargo.toml
```diff
@@ -12,3 +12,10 @@ chain-tx-validation = { path = "../chain-tx-validation" }
 client-common = { path = "../client-common" }
 client-core = { path = "../client-core" }
 secstr = "0.3.2"
+base64 = "0.10"
+chrono = { version = "0.4", features = ["serde"] }
+parity-codec = { features = ["derive"], version = "4.1.1" }
+hex = ""
```
tomtau: version?

client-network/src/network_ops/default_network_ops_client.rs
```diff
@@ -8,41 +10,77 @@ use chain_core::tx::data::attribute::TxAttributes;
...
+    fn get_staked_state_account(
+        &self,
+        to_staked_account: StakedStateAddress,
+    ) -> Result<StakedState> {
+        match to_staked_account {
+            StakedStateAddress::BasicRedeem(a) => self.get_account(&a.0),
+        }
+    }
+
+    fn get_staked_state_nonce(&self, to_staked_account: StakedStateAddress) -> Result<Nonce> {
```
tomtau: as suggested by @devashishdxt , this method is probably not necessary -- nonce can be retrieved in one line:
```
let nonce = self.get_staked_state_account(to_staked_account)?.nonce;
```

client-network/src/network_ops/default_network_ops_client.rs
```diff
@@ -78,10 +116,11 @@ where
         from_address: &ExtendedAddr,
         value: Coin,
         attributes: StakedStateOpAttributes,
+        nonce: Nonce,
```
tomtau: can be removed from high-level APIs

client-network/src/network_ops/default_network_ops_client.rs
leejw51crypto: got it

client-network/src/network_ops.rs
leejw51crypto: got it

client-network/src/network_ops/default_network_ops_client.rs
leejw51crypto: yes , it's removed.
will be automatically fetched from core

leejw51crypto:
for signing txs, it fetch nonce from core,
and signs it.

client-network/src/network_ops/default_network_ops_client.rs
```diff
@@ -80,22 +112,25 @@ where
...
-                    .ok_or_else(|| Error::from(ErrorKind::PrivateKeyNotFound))?;
-                let signature = private_key
-                    .sign(transaction.id())
-                    .map(StakedStateOpWitness::new)?;
-                Ok(TxAux::UnbondStakeTx(transaction, signature))
-            }
+            ExtendedAddr::BasicRedeem(ref redeem_address) => self
+                .get_staked_state_account(StakedStateAddress::BasicRedeem(*redeem_address))
+                .and_then(|account| {
+                    let nonce = account.nonce;
```
tomtau: this signing procedure is almost the same as for `WithdrawUnbondedTx`, so I guess the common parts could be abstracted out into a function?

PR #177, opened 2019-07-02
Problem: (CRO-197) Latest ABCI (for tendermint 0.32) doesn't compile

devashishdxt:
Solution: Changed abci version to 0.6 and made all necessary changes.

- tendermint 0.32 also involves changes in the response structure of `/block_results` RPC call which will be fixed as a part of separate PR.

### DIFF
--- a/chain-abci/src/app/mod.rs
+++ b/chain-abci/src/app/mod.rs
@@ -196,7 +196,10 @@ impl abci::Application for ChainNodeApp {
             kvpair.key = Vec::from(&b"txid"[..]);
             // TODO: "Keys and values in tags must be UTF-8 encoded strings" ?
             kvpair.value = Vec::from(&txaux.tx_id()[..]);
-            resp.tags.push(kvpair);
+            let mut event = Event::new();
+            event.field_type = "delivered_txs".to_string();
+            event.attributes.push(kvpair);
+            resp.events.push(event);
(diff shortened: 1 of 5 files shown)


chain-abci/src/app/mod.rs
```diff
@@ -196,7 +196,10 @@ impl abci::Application for ChainNodeApp {
             kvpair.key = Vec::from(&b"txid"[..]);
             // TODO: "Keys and values in tags must be UTF-8 encoded strings" ?
             kvpair.value = Vec::from(&txaux.tx_id()[..]);
-            resp.tags.push(kvpair);
+            let mut event = Event::new();
+            event.field_type = "delivered_txs".to_string();
```
tomtau: I guess these field types can be anything or at least I didn't find any convention: https://tendermint.com/docs/spec/abci/abci.html#events ?

perhaps extra these strings / event types into constants -- I'm assuming they may need to be used in client, so perhaps put them in chain-core (could be in a separate PR)?

tomtau: And as there are now more key-value levels, the event type could be named based on what it contains: e.g. "valid_txs" in this case, "block_filter" in the other?

devashishdxt: Extracted these event types in an enum in `chain_core`

PR #178, opened 2019-07-02
Problem: latest ethbloom doesn't compile (CRO-191)

lezzokafka:
Solution: updated abci app test to use a fixed sized array that works with the new ethbloom API

### DIFF
--- a/chain-abci/tests/abci_app.rs
+++ b/chain-abci/tests/abci_app.rs
@@ -557,7 +557,9 @@ fn valid_commit_should_persist() {
     assert_eq!(1, cresp.events.len());
     assert_eq!(1, cresp.events[0].attributes.len());
     assert_eq!(1, app.delivered_txs.len());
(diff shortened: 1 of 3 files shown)


chain-abci/tests/abci_app.rs
```diff
@@ -557,7 +557,9 @@ fn valid_commit_should_persist() {
     assert_eq!(1, cresp.events.len());
     assert_eq!(1, cresp.events[0].attributes.len());
     assert_eq!(1, app.delivered_txs.len());
-    let bloom = Bloom::from(&cresp.events[0].attributes[0].value[..]);
+    let mut bloom_array = [0u8; 256];
+    bloom_array.copy_from_slice(&cresp.events[0].attributes[0].value);
```
devashishdxt: Add `assert_eq!(256, &cresp.events[0].attributes[0].value.len())` before copying?

PR #187, opened 2019-07-09
Problem: no standard "textual" addresses(CRO-165)

leejw51crypto:
Solution: Add textual api & unit-test using bech32


chain-core/Cargo.toml
```diff
@@ -20,6 +20,7 @@ serde_json = "1.0"
 parity-codec = { features = ["derive"], version = "4.1.1" }
 base64 = "0.10"
 static_assertions = "0.3.3"
+bech32= ""
```
tomtau: add a version?

chain-core/src/init/address.rs
```diff
@@ -135,6 +145,25 @@ impl RedeemAddress {
     }
 }
 
+impl CroAddress<RedeemAddress> for RedeemAddress {
+    fn to_cro(&self) -> Result<String, ()> {
+        let checked_data: Vec<u5> = self.0.to_vec().to_base32();
+        Bech32::new("crms".into(), checked_data)
```
tomtau: this hardcodes the network -- perhaps `to_cro` should take the network as a parameter? could be an enum:
```
pub enum Network {
   Mainnet,
   Testnet,
}
```

chain-core/src/init/address.rs
```diff
@@ -135,6 +145,25 @@ impl RedeemAddress {
...
+            .map_err(|_e| ())
```
tomtau: this throws away errors: https://docs.rs/bech32/0.6.0/bech32/enum.Error.html
for encoding, I guess there would only be an error if we made a mistake in the human-readable part? I guess there could be a compile-time assertion and this could be safely unwrapped?

chain-core/src/init/address.rs
```diff
@@ -135,6 +145,25 @@ impl RedeemAddress {
...
+            .and_then(|b| {
+                let bytes = b.to_string().as_bytes().to_vec();
+                String::from_utf8(bytes).map_err(|_e| ())
+            })
+    }
+
+    fn from_cro(encoded: &str) -> Result<Self, ()> {
+        encoded
+            .parse::<Bech32>()
+            .map_err(|_e| ())
```
tomtau: for parsing, there could be many possible errors: https://docs.rs/bech32/0.6.0/bech32/enum.Error.html

perhaps create an enum error type that would list / wrap all the possible errors:
```
enum CroAddressError {
   Bech32Error(bech32::Error),
   ...
}
```
and this could be returned?

chain-core/src/init/address.rs
devashishdxt: Or maybe perform conditional compilation based on compile time arguments?

chain-core/src/init/address.rs
```diff
@@ -26,6 +27,15 @@ use tiny_keccak::Keccak;
 
 use crate::common::{H256, HASH_SIZE_256};
 
+// CRMS: mainnet staked-state
+// CRMT: mainnet transfer
+// CRTS: testnet staked-state
+// CRTT: testnet transfer
+pub trait CroAddress<T> {
```
devashishdxt: Any particular reason for having this trait and not using Rust standard traits, `Display` and `FromStr`, for doing the conversion?
The advantage of implementing `Display` and `FromString` is that you get a lot of convenience functions for free. For example, you can use `.to_string()` and `.parse()` functions.

leejw51crypto: currently, Display and FromString works for hexa address
Textual Address can vary , also can have multiple forms like container.

chain-core/src/init/address.rs
leejw51crypto: because chaing network is very frequent, let's go with runtime options.

chain-core/src/init/address.rs
leejw51crypto: ok

chain-core/src/init/address.rs
tomtau: I think these standard traits would be possible, i.e. change their current `impl` in corresponding addresses.
The only drawback would be that it'd require conditional compilation (as they don't take extra arguments) -- https://github.com/crypto-com/chain/pull/187#discussion_r301376681

This requirement would then propagate downstream -- e.g. it wouldn't be possible / easy to manage testnet and mainnet wallets inside client-* at the same time

leejw51crypto:
> I think we also need to implement `Serialize` and `Deserialize` for different addresses so that json conversion happens properly.



> I think we also need to implement `Serialize` and `Deserialize` for different addresses so that json conversion happens properly.

ok

chain-core/src/init/address.rs
leejw51crypto: how about standard traits for CroAddress?

tomtau:
@leejw51crypto you can later also add a document to https://github.com/crypto-com/chain-docs that describes the address convention

leejw51crypto:
> @leejw51crypto you can later also add a document to https://github.com/crypto-com/chain-docs that describes the address convention

ok

chain-core/src/init/address.rs
leejw51crypto: broken checksum happens often.
cannot be unwrapped or expected

leejw51crypto: agree

chain-core/src/init/address.rs
devashishdxt: One more advantage of using standard traits is that `client-cli` and `client-rpc` already uses them to display data. If we use a custom trait, we'll have to search for every usage of `.to_string()` and `.parse()` for any address type and change it to use `CroAddress`.

tomtau: One thing to watch out there though is that the hex-encoded variant is still needed for RedeemAddress in InitConfig

devashishdxt: > One thing to watch out there though is that the hex-encoded variant is still needed for RedeemAddress in InitConfig

Why can't we change that to the new format?

tomtau: We can -- but InitConfig is, from a large part, a textual dump of Ethereum ERC20 state and all the Eth account-related information published in official materials.

tomtau: hence the hex encoding would preferable stay in InitConfig, as it'll be easier for verification

leejw51crypto: OK,
let's use standard trait.
I'll work on it.

tomtau:
@leejw51crypto as discussed yesterday with @devashishdxt:
- use the new address format for value transfers / ExtendedAddr (CRMT, CRTT)
- keep using the old Eth-style hex dump address format for staking operations / RedeemAddress (until the genesis / Eth-interop story is finalized)

leejw51crypto:
> @leejw51crypto as discussed yesterday with @devashishdxt:
> 
>     * use the new address format for value transfers / ExtendedAddr (CRMT, CRTT)
> 
>     * keep using the old Eth-style hex dump address format for staking operations / RedeemAddress (until the genesis / Eth-interop story is finalized)

RedeemAddr, ExtendedAddr
understands 0x... addresses.
i restored back previous hex reading codes.

chain-core/src/init/address.rs
leejw51crypto: done,


used conditional compilation.
chain-core = { path = "../chain-core" , features = ["mainnet"]}
chain-core = { path = "../chain-core" , features = ["testnet"]}

otherwise, rust don't like mutable global state.
1.
so need to hand over globla state to all classes
2. 
or use unstable block, to access some global state,

both are possible, but need some discussion.
so let's proceed with conditional compilation for now.

chain-core/Cargo.toml
```diff
@@ -8,6 +8,8 @@ edition = "2018"
 
 [features]
 default = ["serde"]
+mainnet = []
```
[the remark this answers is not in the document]
tomtau: Or one perhaps more evil way for dynamic network choice would be:
```
#[cfg(feature = "serde")]
static mut CURRENT_NETWORK: network::Network = network::Network::Testnet;
```
(as it mainly matters in serde?)

devashishdxt:
>     1. network feature-guarding of chain-core -- we can discuss on Slack or in person what to do

@tomtau @leejw51crypto Let's discuss this tomorrow. I have a few questions and concerns around this.

chain-core/Cargo.toml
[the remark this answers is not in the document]
tomtau: a bit more cleaner (but more invasive) change would be to add a generic type parameter that would bubble up cross chain-core

chain-abci/tests/abci_app.rs
```diff
@@ -179,7 +179,7 @@ fn previously_stored_hash_should_match() {
 fn init_chain_for(address: RedeemAddress) -> ChainNodeApp<MockClient> {
```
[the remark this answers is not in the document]
leejw51crypto: got it,
i will change as hex test

chain-core/src/tx/data/address.rs
```diff
@@ -18,34 +21,110 @@ pub enum ExtendedAddr {
...
+}
+
+impl FromStr for ExtendedAddr {
+    type Err = CroAddressError;
 
-        if decoded.len() != HASH_SIZE_256 {
-            return Err(hex::FromHexError::InvalidStringLength);
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        if s.starts_with("0x") {
+            ExtendedAddr::from_hex(s).map_err(|_e| CroAddressError::ConvertError)
```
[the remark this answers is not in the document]
leejw51crypto: ok

chain-core/Cargo.toml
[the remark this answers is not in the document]
leejw51crypto: yes, agree for basic approach.
rust don't like getting global mutable state.
1. make a global class, and hand over to classes, which needs to access global state
2. use mutable global state, use unstable block or function to access

both are valid options.
what do you think?

leejw51crypto:
redeem address for hex,
extended address for bech32

leejw51crypto:
ok, i'll fix it now.

client-core/src/key/public_key.rs
```diff
@@ -148,6 +148,7 @@ mod tests {
         let address = RedeemAddress::from(&public_key);
 
         let address = address.to_string();
+        println!("{}", address);
```
devashishdxt: This will resolve merge conflict.
```suggestion
```

chain-core/src/tx/data/address.rs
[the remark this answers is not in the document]
tomtau: What I meant is that there's no need for hex encoding in `ExtendedAddr` -- is there any reason for keeping it there (tests?)?

client-core/src/key/public_key.rs
leejw51crypto: thanks, i'll fix it

chain-core/src/init/address.rs
```diff
@@ -135,6 +169,58 @@ impl RedeemAddress {
     }
 }
 
+impl CroAddress<RedeemAddress> for RedeemAddress {
```
tomtau: As mentioned her: https://github.com/crypto-com/chain/pull/187/files#r302037650 RedeemAddress could be hex-only for start

chain-core/Cargo.toml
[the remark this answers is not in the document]
tomtau: the feature flags should be removed

chain-core/src/init/mod.rs
```diff
@@ -16,3 +18,9 @@ pub const MAX_COIN: u64 = 100_000_000_000__0000_0000;
 pub const MAX_COIN_DECIMALS: u64 = 1_0000_0000;
 /// 100 billion
 pub const MAX_COIN_UNITS: i64 = 100_000_000_000;
+
+#[cfg(feature = "testnet")]
+pub const CURRENT_NETWORK: network::Network = network::Network::Testnet;
```
tomtau: no need for these

chain-core/Cargo.toml
tomtau: the feature flags should be removed

leejw51crypto: ok, i agree

chain-core/src/init/address.rs
leejw51crypto: ok
i'll modify as follow
RedeemAddress: hex only
ExtendedAddress: bech32 only

chain-core/src/init/mod.rs
leejw51crypto: ok, i'll remove this feature.
so our binary will be one, networkid will work as runtime argument.

PR #190, opened 2019-07-09
Problem: (CRO-212) Network operations (staking related operations) are not exposed in CLI

devashishdxt:
Solution: Added CLI commands for performing network operations


client-cli/src/command/transaction_command.rs
```diff
@@ -26,83 +60,219 @@ pub enum TransactionCommand {
...
+fn new_withdraw_transaction<N: NetworkOpsClient>(
+    network_ops_client: &N,
+    name: &str,
+    passphrase: &SecUtf8,
+    chain_id: &str,
+) -> Result<TxAux> {
+    let attributes =
+        TxAttributes::new(decode(chain_id).context(ErrorKind::DeserializationError)?[0]);
+    let from_address = ask_staking_address()?;
+    let outputs = ask_outputs()?;
```
tomtau: @devashishdxt I think for the current withdraw, I guess it would be desirable:

1. to automatically put the output lock time based on `unbonded_from`
2. print some warning (or ask for confirmation) about the implicit fee (I guess the transaction type should be better called "withdraw all unbonded")

leejw51crypto: how about entering twice for fee?

tomtau: what exactly do you have in mind for "entering twice"?
I think one solution would be a confirmation printout at the end, something like:
```
fee to be paid: <...> (minimal fee: <...>)
Do you wish to broadcast this transaction [Y/n]: 
```

devashishdxt: @tomtau: Is there any minimum fee for withdraw transactions? If yes, is it calculated by same `FeePolicy` as for transfer transactions?

I’m thinking of not asking for an amount from user but calculating it based on fee.

devashishdxt:
> looks good, I only have two suggestions for the withdraw operation, as given the current logic, its outcome may a bit unexpected (producing invalid transactions or paying unnecessary high fees)

After discussing with @tomtau, I've decided to implement automatic calculation of withdraw amount based on `FeePolicy`. This change will be done in a separate PR.

PR #200, opened 2019-07-16
Problem: network ops (e.g. staking ops) not exposed in client-rpc (cro-203)

leejw51:

PR #215, opened 2019-07-21
Problem: JSON-RPC wallet_transactions response format is not parse friendly (cro-109)

leejw51crypto:


chain-core/src/init/coin.rs
```diff
@@ -18,7 +18,7 @@ use std::convert::TryFrom;
 use std::{fmt, mem, ops, result, slice};
 
 #[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone, Copy, Encode)]
-pub struct Coin(u64);
+pub struct Coin(pub u64);
```
tomtau: preferable this wouldn't be exposed, as it'd allow changing the internal value that would break Coin's invariants (e.g. one could make it more than the maximum supply).

it can be kept as `pub struct Coin(u64);` -- if you need the internal u64, just use `.into()` / `u64::from(...)`

client-rpc/src/wallet_rpc.rs
```diff
@@ -6,16 +6,27 @@ use secstr::SecUtf8;
...
 use client_core::{MultiSigWalletClient, WalletClient};
 
 use crate::server::{rpc_error_from_string, to_rpc_error};
 
+#[derive(Serialize, Deserialize)]
+pub struct RowTx {
+    kind: String,
+    transaction_id: String,
+    address: String,
+    height: String,
```
tomtau: height is `i64` from Tendermint's protobufs and it'll be "0, 1, 2, 3 ...", so I guess it'd be parse-friendly for a long time

chain-core/src/init/coin.rs
leejw51crypto: ok

client-rpc/src/wallet_rpc.rs
leejw51crypto: yes, javascript handles all i64 or u64 in strings.

leejw51crypto:
coin is sealed as before.
used u64::from instead of directly using tuple0

client-rpc/src/wallet_rpc.rs
```diff
@@ -202,13 +213,32 @@ where
...
         self.sync()?;
 
-        match self.client.history(&request.name, &request.passphrase) {
-            Ok(transaction_change) => Ok(transaction_change),
-            Err(e) => Err(to_rpc_error(e)),
-        }
+        self.client
+            .history(&request.name, &request.passphrase)
+            .map_err(|e| to_rpc_error(e))
+            .map(|transaction_changes| {
```
calvinaco: Should we port this to the serialization method of `TransactionChange`? Then all future APIs with this structure in response will have the parse-friendly serialization.

calvinaco:
This response format is good for me. Client can easily integrate with it without concern of lose in precision.

client-rpc/src/wallet_rpc.rs
leejw51crypto: all future api will be the same format with this tx.
you can use the same decoding api afterwards.

client-rpc/src/wallet_rpc.rs
```diff
@@ -202,13 +213,32 @@ where
...
+    fn transactions(&self, request: WalletRequest) -> Result<Vec<RowTx>> {
         self.sync()?;
 
-        match self.client.history(&request.name, &request.passphrase) {
-            Ok(transaction_change) => Ok(transaction_change),
-            Err(e) => Err(to_rpc_error(e)),
-        }
+        self.client
+            .history(&request.name, &request.passphrase)
+            .map_err(|e| to_rpc_error(e))
```
tomtau:
```suggestion
            .map_err(to_rpc_error)
```

tomtau:
@leejw51crypto I merged #200 , so this PR should sync up (as both modify `client-rpc/src/wallet_rpc.rs`)

leejw51crypto:
i'll fix it

leejw51crypto:
merged done

calvinaco:
LGTM

PR #216, opened 2019-07-21
Problem: inconsistent usage of chain-id and network-id names(cro-218)

leejw51crypto:
now it can be configured as runtime option in chain-abci


chain-abci/src/main.rs
```diff
@@ -32,6 +33,13 @@ fn main() {
     let genesis_app_hash = matches.value_of("genesis_app_hash").unwrap();
     let chain_id = matches.value_of("chain_id").unwrap();
 
+    init_chain_id(chain_id);
+    println!(
```
tomtau: can use `info!` instead of `println!`

chain-core/src/init/network.rs
```diff
@@ -1,5 +1,82 @@
-#[derive(Debug)]
+use std::sync::Once;
```
tomtau: I'm not sure if this "std::sync::*" is present in sgx sdk -- the TX validation wouldn't use the human-readable representation, so I think this network module could be feature-guarded in `chain-core/src/init/mod.rs`

chain-core/src/init/network.rs
```diff
@@ -1,5 +1,82 @@
...
+static INIT_NETWORK: Once = Once::new();
+static INIT_NETWORK_ID: Once = Once::new();
```
tomtau: I'm thinking whether there's a need for two "Once" -- it makes things more flexible, so perhaps yes

chain-core/src/init/network.rs
```diff
@@ -1,5 +1,82 @@
...
+pub fn get_full_network_name() -> &'static str {
+    unsafe {
+        match NETWORK {
+            Network::Mainnet => "mainnet",
+            Network::Testnet => "testnet",
+        }
+    }
+}
+
+static mut NETWORK: Network = Network::Mainnet;
```
tomtau: these two mutable static could be possible wrapped in a private module, e.g.: 

```
mod chosen_network {
  static mut NETWORK: Network = Network::Mainnet;
  static mut NETWORK_ID: u8 = 0 as u8;
}
```

to make the encapsulation more explicit (that it can only be read and initialised using the public functions in network)

chain-core/src/init/network.rs
tomtau: it would also require feature guards elsewhere in chain-core where the network's functions are used

chain-core/src/tx/data/address.rs
```diff
@@ -27,14 +29,8 @@ pub enum ExtendedAddr {
...
-        match crate::init::CURRENT_NETWORK {
-            crate::init::network::Network::Testnet => {
-                bech32::encode("crtt", checked_data).expect("bech32 crmt encoding")
-            }
-            crate::init::network::Network::Mainnet => {
-                bech32::encode("crmt", checked_data).expect("bech32 crmt encoding")
-            }
-        }
+
+        bech32::encode(get_bech32_human_part(), checked_data).expect("bech32 crmt encoding")
```
tomtau: `.expect("bech32 encoding")`

I guess this shouldn't ever panic?

client-rpc/src/main.rs
```diff
@@ -30,12 +30,12 @@ pub(crate) struct Options {
     port: u16,
 
     #[structopt(
-        name = "chain-id",
+        name = "network-id",
         short,
         long,
-        help = "Chain ID (Last two hex digits of chain-id)"
+        help = "Network ID (Last two hex digits of network-id)"
```
tomtau: I guess the comment should be `"Network ID (Last two hex digits of chain_id)"`?
if network-id = the single byte convention;  chain-id = the full network name string in Tendermint's genesis.json

chain-core/src/init/network.rs
```diff
@@ -1,5 +1,82 @@
...
+static mut NETWORK_ID: u8 = 0 as u8;
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    #[test]
+    fn init_chain_id_should_setup_correctly() {
```
devashishdxt: Add a test for checking `Network::Testnet` initialization.

chain-core/src/init/network.rs
leejw51crypto: ok

chain-core/src/init/network.rs
leejw51crypto: ok, i used bech32 for feature guard

chain-core/src/init/network.rs
leejw51crypto: ok

chain-core/src/tx/data/address.rs
leejw51crypto: hardly not panic,
unwrap will be enough

client-rpc/src/main.rs
leejw51crypto: ok

chain-core/src/init/network.rs
leejw51crypto: by Once,cannnot perform testnet unit-test

chain-core/src/tx/data/address.rs
tomtau: `expect` is fine (it's the same as unwrap, but will print a more meaningful panic) -- I just mean that there could be a comment above explaining why it wouldn't panic (or perhaps a compile-time assertion that proves it wouldn't panic https://github.com/nvzqz/static-assertions-rs#usage )

leejw51crypto:
fixing conflict

leejw51crypto:
fixed collision

chain-core/src/tx/data/address.rs
leejw51crypto: got it.
add more message

PR #218, opened 2019-07-22
Problem: (CRO-219) Potential inconsistent client state due to concurrent updates

devashishdxt:
Solution:

- Combined `balance`, `transaction_history` and `unspent_transactions` in a single keyspace.
- Use a simplified variant of [memento](https://refactoring.guru/design-patterns/memento) design pattern to buffer all database updates and execute them in the end.


client-index/src/service/address_service.rs
```diff
@@ -0,0 +1,218 @@
...
+use client_common::balance::TransactionChange;
+use client_common::{ErrorKind, Result, Storage};
+
+const KEYSPACE: &str = "index_address";
+
+/// All the details related to an address
+#[derive(Debug, Encode, Decode)]
+pub struct AddressDetails {
+    /// Unpent transactions corresponding to an address
+    pub unspent_transactions: Vec<(TxoPointer, TxOut)>,
```
tomtau: I guess similarly to the note below, this would ideally change to BTreeMap?

client-index/src/service/transaction_service.rs
```diff
@@ -33,13 +42,38 @@ where
...
+            .map(|_| ())
+    }
+
+    /// Retrieves transaction output corresponding to given pointer
+    pub fn get_output(&self, input: &TxoPointer) -> Result<TxOut> {
+        let transaction = self
+            .get(&input.id)?
+            .ok_or_else(|| Error::from(ErrorKind::TransactionNotFound))?;
+
+        let outputs = match transaction {
```
tomtau: I'm thinking whether https://github.com/crypto-com/chain/blob/44d98640c679bd49b7dff13607267c2b8a6f4fad/chain-core/src/tx/mod.rs#L24 could be used to simplify this

client-index/src/service/address_service.rs
devashishdxt: Added a comment to change it when `BTreeSet` is implemented in `parity_codec`.

leejw51crypto:
lgtm

PR #220, opened 2019-07-22
Problem: (CRO-215) No data structure to hold details of a block header

devashishdxt:
Solution: Added `BlockHeader` data type in `client-common`

- Also, changed tendermint RPC query data from `&str` to `&[u8]`

### DIFF
--- a/client-common/src/block_header.rs
+++ b/client-common/src/block_header.rs
@@ -1,0 +1,16 @@
+use chrono::{DateTime, Utc};
+
+use chain_core::tx::data::TxId;
+use chain_tx_filter::BlockFilter;
+
+/// Structure for representing a block header on Crypto.com Chain
+pub struct BlockHeader {
+    /// Block height
+    pub block_height: u64,
+    /// Block time
+    pub block_time: DateTime<Utc>,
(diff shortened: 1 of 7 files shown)


client-common/src/block_header.rs
```diff
@@ -0,0 +1,16 @@
+use chrono::{DateTime, Utc};
+
+use chain_core::tx::data::TxId;
+use chain_tx_filter::BlockFilter;
+
+/// Structure for representing a block header on Crypto.com Chain
+pub struct BlockHeader {
+    /// Block height
+    pub block_height: u64,
```
tomtau: I'm thinking whether it's worth importing this: https://github.com/crypto-com/chain/blob/master/chain-core/src/state/tendermint.rs#L9 so there's a single alias / way of representing block height?
currently, it's i64, because that's how it's represented in the TM protobuf

devashishdxt: This is temporary. [tendermint.rs](https://crates.io/crates/tendermint) have their own trait around height. We'll have to use that in client when we integrate change our RPC implementation.

PR #222, opened 2019-07-22
Problem: (CRO-215) Transaction synchronization code is tightly coupled with transaction index

devashishdxt:
Solution: This is first of a series of PRs which will slowly extract out all the transaction synchronization code from transaction index


devashishdxt:
Code in `TransactionHandler` is almost exactly similar to current code in `DefaultIndex`. In future PRs, I'll be deleting the duplicate code from `DefaultIndex` and using `TransactionHandler` for synchronization.

PR #224, opened 2019-07-23
Problem: (CRO-215) Transaction synchronization code is tightly coupled with transaction index

devashishdxt:
**Solution**: This is second of a series of PRs which will slowly extract out all the transaction synchronization code from transaction index.

**Note**: Current implementation of `BlockHandler` only handles transaction which are encrypted (`Transfer` and `Withdraw`). Implementation for `DepositStake` transactions will be a part of future PRs.

### DIFF
--- a/client-cli/src/command.rs
+++ b/client-cli/src/command.rs
@@ -196,7 +196,6 @@ impl Command {
 
     fn get_balance<T: WalletClient>(wallet_client: T, name: &str) -> Result<()> {
         let passphrase = ask_passphrase()?;
-        wallet_client.sync()?;
         let balance = wallet_client.balance(name, &passphrase)?;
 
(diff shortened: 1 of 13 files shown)


client-cli/src/command.rs
```diff
@@ -196,7 +196,6 @@ impl Command {
 
     fn get_balance<T: WalletClient>(wallet_client: T, name: &str) -> Result<()> {
         let passphrase = ask_passphrase()?;
-        wallet_client.sync()?;
```
leejw51crypto: why removed sync ?

devashishdxt: Most of the client logic will change in future. Now that new have view keys, client will have to maintain view key of current user somewhere and then pass it to `sync()`. Currently, we don't have that.

PR #226, opened 2019-07-23
Problem: (CRO-215) Duplicate code in DefaultIndex implementation

devashishdxt:
Solution: Removed all the duplicate code related to synchronization from `DefaultIndex`. Added remaining synchronization code in `ManualSynchronizer`.

### DIFF
--- a/client-core/src/wallet/default_wallet_client.rs
+++ b/client-core/src/wallet/default_wallet_client.rs
@@ -215,7 +215,7 @@ where
 
         let balances = addresses
             .iter()
-            .map(|address| self.index.balance(address))
+            .map(|address| Ok(self.index.address_details(address)?.balance))
             .collect::<Result<Vec<Coin>>>()?;
 
         Ok(sum_coins(balances.into_iter()).context(ErrorKind::BalanceAdditionError)?)
@@ -226,7 +226,7 @@ where
 
         let history = addresses
             .iter()
-            .map(|address| self.index.transaction_changes(address))
+            .map(|address| Ok(self.index.address_details(address)?.transaction_history))
             .collect::<Result<Vec<Vec<TransactionChange>>>>()?
             .into_iter()
             .flatten()
@@ -244,7 +244,7 @@ where
 
         let mut unspent_transactions = Vec::new();
         for address in addresses {
-            unspent_transactions.extend(self.index.unspent_transactions(&address)?);
+            unspent_transactions.extend(self.index.address_details(&address)?.unspent_transactions);
         }
 
         Ok(UnspentTransactions::new(unspent_transactions))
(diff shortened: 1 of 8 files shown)


client-core/src/wallet/default_wallet_client.rs
```diff
@@ -226,7 +226,7 @@ where
 
         let history = addresses
             .iter()
-            .map(|address| self.index.transaction_changes(address))
+            .map(|address| Ok(self.index.address_details(address)?.transaction_history))
             .collect::<Result<Vec<Vec<TransactionChange>>>>()?
```
leejw51crypto: can we do more flattening?

leejw51crypto: type TxChanges = vec<TransactionChange>;
type TxChangesRow = vec<TxChanges>;
or using type?

devashishdxt: One line [232](https://github.com/crypto-com/chain/pull/226/files/89926f269fadccf4abaf13fe7fb429e48fe38c59#diff-56d5986577a1416a78d11832b08af3c0R232) we `flatten()` this structure.
Do you think it can be improved further?

tomtau: Maybe `flat_map` could be called instead of `map`: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.flat_map ?

devashishdxt: We cannot use `flat_map` here because closure in `map` returns a `Result`, but, `flat_map`'s closure should return a value which implements `IntoIterator`.

@tomtau @leejw51crypto Do you know of a way to do this without `collect`ing the values?

tomtau: https://doc.rust-lang.org/std/result/enum.Result.html#impl-IntoIterator ?
```
let x: Result<u32, &str> = Ok(5);
let v: Vec<u32> = x.into_iter().collect();
assert_eq!(v, [5]);

let x: Result<u32, &str> = Err("nothing!");
let v: Vec<u32> = x.into_iter().collect();
assert_eq!(v, []);
```

should be ok?

leejw51crypto: and_then for second map?

devashishdxt: @tomtau `IntoIterator` implementation of `Result` throws away all the errors and just returns all the successful values which is not what we want here.

[Here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=22dfb25bbc7a46af017f2dc47694d8dd)'s an example. `version1` is using the approach currently used and `version2` implements `flat_map` approach.

Any comments or improvements?

tomtau: I've been thinking something like https://hackage.haskell.org/package/base-4.12.0.0/docs/Prelude.html#v:sequenceA

But yeah, the "magic" conversion is that there's that `FromIterator` implementation: https://doc.rust-lang.org/std/result/enum.Result.html#implementations

so it's probably via this:
```
impl<A, E, V> FromIterator<Result<A, E>> for Result<V, E>
where
    V: FromIterator<A>, 
```

I guess the `collect` is the way to go, as there doesn't seem to be a general Traversable equivalent

PR #227, opened 2019-07-23
Problem: (CRO-215) Client does not handle unencrypted transaction in a block (deposit and unbond)

devashishdxt:
Solution: Modified `BlockHandler` and `ManualSynchronizer` to handle unencrypted transactions

### DIFF
--- a/client-common/src/tendermint/types/block.rs
+++ b/client-common/src/tendermint/types/block.rs
@@ -8,7 +8,7 @@ use serde::Deserialize;
 
 use chain_core::tx::TxAux;
 
-use crate::{ErrorKind, Result};
+use crate::{Error, ErrorKind, Result, Transaction};
 
 #[derive(Debug, Deserialize)]
 pub struct Block {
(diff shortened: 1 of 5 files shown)


client-common/src/tendermint/types/block.rs
```diff
@@ -33,18 +33,38 @@ pub struct Header {
...
+                            TxAux::DepositStakeTx { tx, .. } => {
+                                Ok(Some(Transaction::DepositStakeTransaction(tx)))
+                            }
+                            TxAux::UnbondStakeTx(tx, _) => {
+                                Ok(Some(Transaction::UnbondStakeTransaction(tx)))
+                            }
+                            _ => Ok(None),
+                        }
+                    })
+                    .collect::<Result<Vec<Option<Transaction>>>>()?;
```
tomtau: could the `filter_map` directly on the first iter()?

client-index/src/synchronizer.rs
```diff
@@ -34,7 +37,12 @@ where
     }
 
     /// Synchronizes transaction index for given view key with Crypto.com Chain (from last known height)
-    pub fn sync(&self, view_key: &PublicKey, private_key: &PrivateKey) -> Result<()> {
+    pub fn sync(
```
tomtau: I'm wondering, as these parts are now being rewritten -- could the manual sync make use of batch requests? (perhaps in a different PR)
From TM 0.35, one could make use of https://www.jsonrpc.org/specification#batch https://godoc.org/github.com/tendermint/tendermint/rpc/client
so instead of one JSON-RPC call at a time, one could submit several in an array in one HTTP request

leejw51crypto: yes
using batch will increase sync speed dramatically

also other approaches which eth is using 
make headers download first,
multiple downloading simulatneously.
this is too complicated, we can try in next phase.

leejw51crypto: also sync is the most challenging part in blockchain networking.

client-common/src/tendermint/types/block.rs
devashishdxt: Done

client-index/src/synchronizer.rs
devashishdxt: This can and should be done. Will do this in future PRs.

PR #229, opened 2019-07-24
Problem: (CRO-215) TransactionBuilder is not integrated with TransactionCipher

devashishdxt:
Solution: Added `TransactionCipher` in `TransactionBuilder` and started using it for transaction encryption.

### DIFF
--- a/client-core/src/transaction_builder/default_transaction_builder.rs
+++ b/client-core/src/transaction_builder/default_transaction_builder.rs
@@ -4,15 +4,12 @@ use secstr::SecUtf8;
 use chain_core::init::coin::{sum_coins, Coin};
 use chain_core::tx::data::address::ExtendedAddr;
 use chain_core::tx::data::attribute::TxAttributes;
-use chain_core::tx::data::input::TxoIndex;
 use chain_core::tx::data::output::TxOut;
 use chain_core::tx::data::Tx;
 use chain_core::tx::fee::FeeAlgorithm;
-use chain_core::tx::PlainTxAux;
-use chain_core::tx::TxObfuscated;
 use chain_core::tx::{TransactionId, TxAux};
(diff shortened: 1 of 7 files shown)


client-core/src/transaction_builder/default_transaction_builder.rs
```diff
@@ -25,38 +22,45 @@ use crate::{SelectedUnspentTransactions, Signer, TransactionBuilder, UnspentTran
 /// 3. Select unspent transactions with `fees + output_value`.
 /// 4. Build transaction with selected unspent transactions (also add an extra output for change amount).
 /// 5. Sign transaction with private keys corresponding to selected unspent transactions.
-/// 6. Calculate `new_fees`.
-/// 7. If `new_fees > fees`, then change `fees = new_fees` and goto step 3, otherwise return signed transaction.
+/// 6. Encrypt/obfuscate transaction.
+/// 7. Calculate `new_fees`.
+/// 8. If `new_fees > fees`, then change `fees = new_fees` and goto step 3, otherwise return signed transaction.
```
tomtau: I think this could potentially be simplified (i.e. no need to have several round trips -- the fee can be estimated correctly upfront) once there's a concrete implementation: if it's obfuscated by a stream cipher, it should be the same... if it's a block cipher, it'll be rounded by the block size.
I think it'll most likely be AES_128_GCM_SIV

devashishdxt: Yes.

Currently, we have to do this because 1-of-n signatures consist of a schnorr signature and a merkle proof which can be of variable size. So, we cannot deterministically estimate the size and structure of a 1-of-n signature. 

Once we have a transaction format where size can be estimated deterministically, we can change this.

PR #249, opened 2019-07-31
Problem: StakeStateAddress to_string adds a newline character (cro-228)

leejw51crypto:
Solution: use write instead of writeln


tomtau:
@leejw51crypto what is the problem this is trying to solve?

PR #251, opened 2019-08-01
Problem: (CRO-254) No network for devnet/local testing + reflect a new name

devashishdxt:
Solution: Added devnet in Network enum and human readable part of bech32 address according to proposed convention

### DIFF
--- a/chain-core/src/init/address.rs
+++ b/chain-core/src/init/address.rs
@@ -42,7 +42,7 @@ impl ::std::error::Error for CroAddressError {}
 
 // CRO: mainnet transfer
 // TCRO: testnet transfer
-// DCRO: testnet transfer
+// DCRO: devnet/regnet transfer
 #[cfg(feature = "bech32")]
 pub trait CroAddress<T> {
     fn to_cro(&self) -> Result<String, CroAddressError>;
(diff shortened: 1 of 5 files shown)


chain-core/src/init/address.rs
```diff
@@ -33,14 +33,16 @@ use crate::common::{H256, HASH_SIZE_256};
...
 }
 
 #[cfg(feature = "bech32")]
 impl ::std::error::Error for CroAddressError {}
 
-// CRMT: mainnet transfer
-// CRTT: testnet transfer
+// CRO: mainnet transfer
+// TCRO: testnet transfer
+// DCRO: testnet transfer
```
tomtau: DCRO = devnet/regnet (for unit tests or local development testing)
TCRO = future deployed public Testnet

chain-core/src/tx/data/address.rs
```diff
@@ -99,7 +99,7 @@ mod test {
         let b = a.to_cro().unwrap();
         assert_eq!(
             b,
-            "crmt1pe7qg5gshrdl99m9q3ecpzvfr8zuk4h5qqgjyv6y24n80zye42asr8c7xt"
+            "cro1pe7qg5gshrdl99m9q3ecpzvfr8zuk4h5qqgjyv6y24n80zye42asjwegvg"
```
tomtau: the tests should ideally use "dcro"

leejw51crypto: ok

chain-core/src/init/address.rs
```diff
@@ -174,25 +177,22 @@ impl RedeemAddress {
...
-                let encoded = bech32::encode("crts", checked_data).expect("bech32 crms encoding");
-                Ok(encoded.to_string())
-            }
-            super::network::Network::Mainnet => {
-                let encoded = bech32::encode("crms", checked_data).expect("bech32 crms encoding");
-                Ok(encoded.to_string())
-            }
-        }
+        let encoded = bech32::encode(super::network::get_bech32_human_part(), checked_data)
+            .expect("bech32 crms encoding");
```
tomtau: Could this encoding return an error instead fo panicking (or would it ever panic / return an error?)?

devashishdxt: This will only panic when human readable part is invalid and I assume we already have a static assertion for that (@leejw51crypto)?

[Documentation](https://docs.rs/bech32/0.7.1/bech32/fn.encode.html)

PR #255, opened 2019-08-06
Problem: no easy way to "launch everything"(cro-259)

leejw51crypto:
very simple way to run ,stop all sub-components
currently support
tx-validation-app,
chain-abci,
tendermint

dev-utils init <- setup 
dev-utils run 
dev-utils stop 

automatically read tendermint genesis.json,
all files including tendermint should be in the same folder.
for ubuntu)
apt install psmisc


dev-utils/src/commands/run_command.rs
```diff
@@ -0,0 +1,91 @@
...
+        RunCommand {
+            chainid: "".to_string(),
+            app_hash: "".to_string(),
+        }
+    }
+
+    fn get_tendermint_filename(&self) -> String {
+        format!(
+            "{}/.tendermint/config/genesis.json",
+            dirs::home_dir().unwrap().to_str().unwrap()
```
calvinaco: Should we add a command option to customize genesis file path?

dev-utils/src/commands/run_command.rs
```diff
@@ -0,0 +1,91 @@
...
+                    "tcp://127.0.0.1:25933",
+                ];
+                self.run_program("./chain-abci", args)
+            })
+            .and_then(|_| {
+                println!("wait for abci booting");
+                thread::sleep(time::Duration::from_millis(3000));
+                Ok(())
+            })
+            .and_then(|_| self.run_program("./tendermint", vec!["node"]))
```
calvinaco: I may have misunderstood, does it assume `tendermint` program to sits in the same folder? Similarly for `tx-validator-app` as well.

dev-utils/src/commands/run_command.rs
```diff
@@ -0,0 +1,91 @@
+use failure::{format_err, Error};
+use std::fs;
+use std::process::Command;
+use std::{thread, time};
+#[derive(Debug)]
+pub struct RunCommand {
+    chainid: String,
```
devashishdxt: Change `chainid` to `chain_id`.

dev-utils/src/commands/run_command.rs
leejw51crypto: ok, i will add a path

dev-utils/src/commands/run_command.rs
leejw51crypto: yes,
i assume all files are packaged in the same folder.

dev-utils/src/commands/run_command.rs
leejw51crypto: ok

dev-utils/src/commands/run_command.rs
```diff
@@ -0,0 +1,91 @@
...
+                self.app_hash = json["app_hash"].as_str().unwrap().to_string();
+                println!("chainid={}", self.chainid);
+                println!("app_hash={}", self.app_hash);
+                Ok(())
+            })
+            .map_err(|_e| format_err!("read tendermint genesis error"))
+    }
+
+    pub fn run_program(&self, command: &str, arg: Vec<&str>) -> Result<(), Error> {
+        Command::new(command)
```
tomtau: How is it with the process stdout/stderr logs? I think it'll be useful to see what's going on, especially if it didn't work -- so one option is to dump/pipe the process logs into files, so one can inspect them later...
the other is not to use this custom launcher, but say https://github.com/DarthSim/overmind (this tool lets you specify concurrent processes in one "Procfile" and it'll run them similarly to Docker Compose)
@calvinaco any thoughts

leejw51crypto: it's spawned process, so it's the same with running in multiple consoles.
so each daemon will store logs internally, it will work as the same.

tomtau: but "normally", the default behaviour of the particular spawned processes (tendermint, our abci app + tx validation server) is that they'd print to stdout... will that be lost?

leejw51crypto: if it's necessary, we can pipe and copy , periodically collect, and remove oldest one.
i sugguest internal logging system instead of stdin/out.
because sometime , too many logs can burdern main system.(or precious main system disk space)

tomtau: this is a launcher for development purposes (if the developer's laptop is running out of disk space, I'm sure he can fix it :)).
what do you mean by "internal" logging? something like `journalctl` ? if so, then they'd need to be spawned e.g. as systemd services rather than plain processes

tomtau: For our own programs, the amount of logs can be controlled with `RUST_LOG` environment variable.
For tendermint, there's a "--log-level" flag:
```
      --log_level string   Log level (default "main:info,state:info,*:error")
```

leejw51crypto: i'm thinking just writing to a file.
chain-abci.log
when it reaches some size, then backup 
chain-abci-2019-01-02.log
and start again

or i often use sqlite for logging

dev-utils/src/commands/run_command.rs
leejw51crypto: for activation
export TENDERMINT_HOME=~/.tendermint

for de-activation
unset TENDERMINT_HOME

dev-utils/src/commands/run_command.rs
leejw51crypto: working on with overmind

tomtau: dumping logs to files is fine as well in development -- just as long as there's a way for a developer to easily inspect what was going on in each process, it's fine

leejw51crypto: ok

dev-utils/src/commands/init_command.rs
```diff
@@ -307,10 +310,35 @@ impl InitCommand {
...
+
+        Ok(())
+    }
+
     pub fn execute(&mut self) -> Result<(), Error> {
-        println!("initialize");
+        println!("initialize chain");
 
-        self.prepare_tendermint()
+        self.clear_disk()
```
tomtau: perhaps display some warning or ask for confirmation that it'll delete the stuff

leejw51crypto: ok, i'll add warning and confirmation

dev-utils/src/commands/run_command.rs
leejw51crypto: overmind Procfile added,
now you can begin all stuff by using
overmind s

leejw51crypto:
fixed issues

dev-utils/src/commands/init_command.rs
leejw51crypto: i added warning, double check for clearing folders

PR #259, opened 2019-08-06
Problem: dev-utils json processing is non-deterministic (CRO-228)

leejw51crypto:
item order in json list should be exactly the same for correct hash


leejw51crypto:
yes, the problem is
i converted app_state: InitConfig , to string.
and i turned to serde_json::Valuey by serde_json::from_string(self.app_state)

by InitConfig->string-> JsonValue,    json order in distribution changed.
because we build tree from distribution list tom compute hash.
so computed hash is different.

so i modified again
InitConfig->JsonValue  directly as original code.

it's working as before.

sometimes, order is the same, so no problem.
sometimes, order is different, so hash mis-match occurrs.

becuase it's separated issue from cro-259 (launcher), 
i make a branch from master, and made a pull request.

PR #261, opened 2019-08-07
Problem: (CRO-238) No passphrase confirmation in client cli during new wallet creation

devashishdxt:
Solution: Added passphrase confirmation during new wallet creation

### DIFF
--- a/client-cli/src/command.rs
+++ b/client-cli/src/command.rs
@@ -216,7 +216,7 @@ impl Command {
         name: &str,
         address: &StakedStateAddress,
     ) -> Result<()> {
-        let passphrase = ask_passphrase()?;
+        let passphrase = ask_passphrase(None)?;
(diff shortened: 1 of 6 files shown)


calvinlauco:
This could break the integration test, the fix should be just to echo the password twice in `integration_tests/prepare.sh` during create wallet

tomtau:
> This could break the integration test, the fix should be just to echo the password twice in `integration_tests/prepare.sh` during create wallet

seems the integration test has been broken for some time due to other changes

calvinlauco:
Oh, let me fix it after this is merged then.

devashishdxt:
> This could break the integration test, the fix should be just to echo the password twice in `integration_tests/prepare.sh` during create wallet

I've changed `create_wallet()` in `prepare.sh`.

PR #267, opened 2019-08-08
Problem: (CRO-247) Transaction broadcast is prone to DoS attacks

devashishdxt:
Solution: Added maximum size of binary encoded data for transaction and also put limits on number of inputs

### DIFF
--- a/chain-core/src/state/account.rs
+++ b/chain-core/src/state/account.rs
@@ -19,6 +19,14 @@ use secp256k1::recovery::{RecoverableSignature, RecoveryId};
 use std::convert::{From, TryFrom};
 use std::fmt;
 
+const MAX_INPUTS_LENGTH: usize = 64;
+
+/// Each input is 33 bytes
+///
+/// Assuming maximum inputs allowed are 64,
+/// So, maximum deposit transaction size (33 * 64) + 21 (address) + 1 (attributes) = 2134 bytes
+const MAX_DEPOSIT_TX_SIZE: usize = 2150; // 2150 bytes
+
 /// reference counter in the sparse patricia merkle tree/trie
 pub type Count = u64;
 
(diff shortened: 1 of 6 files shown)


chain-core/src/state/account.rs
```diff
@@ -19,6 +19,14 @@ use secp256k1::recovery::{RecoverableSignature, RecoveryId};
 use std::convert::{From, TryFrom};
 use std::fmt;
 
+const MAX_INPUTS_LENGTH: usize = 64;
```
leejw51crypto: for exchange,
input length can be very large.

leejw51crypto: it will be filtered out by size check,
so how about using very large count ?

leejw51crypto: or no limit

chain-core/src/state/account.rs
```diff
@@ -186,14 +194,41 @@ impl StakedStateOpAttributes {
...
     pub attributes: StakedStateOpAttributes,
 }
 
+impl Decode for DepositBondTx {
+    fn decode<I: Input>(input: &mut I) -> Result<Self, Error> {
+        let size = input
+            .remaining_len()?
+            .ok_or_else(|| "Unable to calculate size of input")?;
+
+        if size > MAX_DEPOSIT_TX_SIZE {
```
leejw51crypto: because utxo table is vector, it can be filered prior to deserialization

leejw51crypto: what do you think ?  @tomtau

chain-core/src/state/account.rs
tomtau: this is a limit for 1 transaction -- if more inputs are needed, one can submit more transactions? @leejw51crypto 
@calvinaco any thoughts on reasonable limits?

chain-core/src/state/account.rs
tomtau: do you mean to find the length as an offset in the binary?

chain-core/src/state/account.rs
leejw51crypto: very large utxo txs will be filtered by the size.
so limiting utxo count doesn't have much effect.

for the certain circumstances, for massive test,
large utxo in tx is necessary.
for stress test etc.

leejw51crypto: we can use many txs though.

chain-core/src/state/account.rs
leejw51crypto: yes,
if vector is really big, 
total message size is also big,
then it will be already filtered out.

leejw51crypto: if we apply limit of individual vectors,
then program can be more complicated.

PR #279, opened 2019-08-14
Problem: (CRO-313) Missing create multisig address rpc

calvinaco:
Solution: Added `multiSig_createAddress` RPC

### DIFF
--- a/client-core/src/service/root_hash_service.rs
+++ b/client-core/src/service/root_hash_service.rs
@@ -188,7 +188,23 @@ mod tests {
                     &passphrase
                 )
                 .expect_err("Created invalid multi-sig address")
-                .kind()
+                .kind(),
+            "Should throw error when required signature is larger than public key size"
+        );
+
+        assert_eq!(
+            ErrorKind::InvalidInput,
+            root_hash_service
+                .new_root_hash(
+                    public_keys.clone(),
+                    public_keys[0].clone(),
+                    5,
+                    3,
+                    &passphrase
+                )
+                .expect_err("Created invalid multi-sig address")
+                .kind(),
+            "Should throw error when required signature is larger than total public keys"
         );
 
         assert_eq!(
@@ -202,7 +218,8 @@ mod tests {
                     &passphrase
                 )
                 .expect_err("Created multi-sig address without self public key")
-                .kind()
+                .kind(),
+            "Should throw error when self public key is not one of the public keys"
         );
 
         assert_eq!(
@@ -210,7 +227,8 @@ mod tests {
             root_hash_service
                 .new_root_hash(vec![], public_keys[0].clone(), 0, 0, &passphrase)
                 .expect_err("Created invalid multi-sig address")
-                .kind()
+                .kind(),
+            "Should throw error when required signature is 0"
         );
 
         let root_hash = root_hash_service
--- a/client-core/src/wallet/default_wallet_client.rs
+++ b/client-core/src/wallet/default_wallet_client.rs
@@ -183,8 +183,10 @@ where
         m: usize,
         n: usize,
     ) -> Result<ExtendedAddr> {
-        // To verify if the passphrase is correct or not
-        self.transfer_addresses(name, passphrase)?;
+        let wallet_public_keys = self.public_keys(&name, &passphrase)?;
+        if !wallet_public_keys.contains(&self_public_key) {
+            return Err(Error::from(ErrorKind::MultiSigInvalidSelfPubKey));
+        }
 
         let root_hash =
             self.root_hash_service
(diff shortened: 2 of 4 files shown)


client-rpc/src/rpc/multisig_rpc.rs
```diff
@@ -81,6 +91,33 @@ impl<T> MultiSigRpc for MultiSigRpcImpl<T>
 where
     T: WalletClient + MultiSigWalletClient + 'static,
 {
+    fn create_address(
+        &self,
+        request: WalletRequest,
+        public_keys: Vec<String>,
+        self_public_key: String,
+        required_signatures: usize,
```
tomtau: should it do some check on `required_signatures` vs `public_keys.len()`?
Or is that handled downstream in `new_multisig_transfer_address` ?

tomtau: or seems it's handled: https://github.com/crypto-com/chain/pull/279/files#diff-95f634f70a8d297a68cbd93dc8696949R195 ?

calvinaco: Yes, most of the errors is handled downstream and is tested, I have added one more test but it's just to make it more complete.

The only check I added in the wallet_client layer is the checking of self public key belongs to the wallet.

client-core/src/service/root_hash_service.rs
```diff
@@ -188,7 +188,23 @@ mod tests {
                     &passphrase
                 )
                 .expect_err("Created invalid multi-sig address")
-                .kind()
+                .kind(),
+            "Should throw error when required signature is larger than public key size"
+        );
+
```
leejw51: assert!(public_keys.size()>0);

client-core/src/wallet/default_wallet_client.rs
```diff
@@ -1134,6 +1136,28 @@ mod tests {
             PublicKey::from(&PrivateKey::new().unwrap()),
         ];
```
leejw51: assert!(public_keys.size()>0);

client-rpc/src/rpc/multisig_rpc.rs
```diff
@@ -5,13 +5,23 @@ use jsonrpc_derive::rpc;
...
 
 #[rpc]
 pub trait MultiSigRpc: Send + Sync {
+    #[rpc(name = "multiSig_createAddress")]
+    fn create_address(
+        &self,
+        request: WalletRequest,
+        public_keys: Vec<String>,
+        self_public_key: String,
+        required_signatures: usize,
```
leejw51: usize is ambious type,
how about u32

client-rpc/src/rpc/multisig_rpc.rs
calvinaco: https://github.com/calvinaco/chain/blob/feature/cro-313-create-multisig-address-rpc/client-core/src/service/root_hash_service.rs#L49

Checking here but the error needs more polishing in the future.

client-core/src/service/root_hash_service.rs
calvinaco: Do you mean to add more test here?

client-core/src/wallet/default_wallet_client.rs
calvinaco: The `public_keys` size check is performed in the underlying function
https://github.com/calvinaco/chain/blob/feature/cro-313-create-multisig-address-rpc/client-core/src/service/root_hash_service.rs#L49

client-rpc/src/rpc/multisig_rpc.rs
calvinaco: Yes, I think a more concrete integer size maybe appropriate. Any reason to use usize before @devashishdxt

devashishdxt: There is no specific reason. The only reason is that Rust's `Vec` length is measured in `usize`. I don't think there is any reason to change it to anything else since we already have a limit on number of leaf nodes a merkle tree can have.

client-core/src/wallet/default_wallet_client.rs
```diff
@@ -183,8 +183,10 @@ where
         m: usize,
         n: usize,
     ) -> Result<ExtendedAddr> {
-        // To verify if the passphrase is correct or not
-        self.transfer_addresses(name, passphrase)?;
+        let wallet_public_keys = self.public_keys(&name, &passphrase)?;
+        if !wallet_public_keys.contains(&self_public_key) {
```
devashishdxt: This check is already performed [here](https://github.com/crypto-com/chain/blob/master/client-core/src/service/root_hash_service.rs#L53). Remove the redundant check here and we can improve error reporting in underlying function.

calvinaco: If I understand correctly, this check is talking about the public key list argument contains the self public key? The check I added is trying to make sure the self key belongs to the wallet.

devashishdxt: Yeah. I misunderstood.

PR #288, opened 2019-08-19
Missing: (CRO-318) Create raw transaction RPC

calvinaco:
Solution: Added create raw transaction RPC


client-rpc/src/rpc/transaction_rpc.rs
```diff
@@ -0,0 +1,168 @@
...
+}
+
+#[rpc]
+pub trait TransactionRpc: Send + Sync {
+    #[rpc(name = "transaction_createRaw")]
+    fn create_raw(
+        &self,
+        inputs: Vec<TxoPointer>,
+        outputs: Vec<TxOut>,
+        view_keys: Vec<String>,
```
tomtau: isn't there a serde-implementation: https://github.com/crypto-com/rust-secp256k1-zkp/blob/master/src/key.rs#L370 ? 
but I guess that one would be an array of numbers instead of some string?

calvinaco: I have to experiment it and will give the experiment results here.

calvinaco:
I am gonna add custom serialization and deserialization to the involved types and fix the compilation error.

calvinaco:
I have added custom serializer/deserializer, now the response becomes
```
{
    "jsonrpc": "2.0",
    "result": {
        "tx": {
            "attributes": {
                "allowed_view": [
                    {
                        "access": "AllData",
                        "view_key": "034dc059a07ae65cff5754859110db63f1e52913dc61f2903a67bf181c596e68df"
                    }
                ],
                "chain_hex_id": "AB"
            },
            "inputs": [
                {
                    "id": "0000000000000000000000000000000000000000000000000000000000000000",
                    "index": 0
                }
            ],
            "outputs": [
                {
                    "address": "dcro16dgqj94gdlkj7kfygd9xx99fra55f409vrsr5ham2kj70tp28nas6emfg4",
                    "valid_from": null,
                    "value": "750"
                }
            ]
        },
        "tx_id": "cb058c146ba4a6798e59ba41b34f725e5f21762699418e04a2ebeb2010c6369d"
    },
    "id": "transaction_createRaw"
}
```

client-rpc/src/rpc/transaction_rpc.rs
calvinaco: I tried and it will require an array to deserialize and serialize as array of numbers as well. I have implemented custom de/serializer here.

chain-core/src/tx/data/attribute.rs
```diff
@@ -9,11 +12,62 @@ use crate::tx::data::access::TxAccessPolicy;
...
+        serde(deserialize_with = "deserialize_chain_hex_id")
+    )]
     pub chain_hex_id: u8,
     pub allowed_view: Vec<TxAccessPolicy>,
     // TODO: other attributes, e.g. versioning info
 }
 
+#[cfg(feature = "serde")]
+fn serialize_chain_hex_id<S>(
+    chain_hex_id: &u8,
```
tomtau: Do the custom serialisers need it by reference?

There's a clippy warning:
```
error: this argument (1 byte) is passed by reference, but would be more efficient if passed by value (limit: 8 byte)
3841  --> chain-core/src/tx/data/attribute.rs:27:19
3842   |
384327 |     chain_hex_id: &u8,
3844   |                   ^^^ help: consider passing by value instead: `u8`
3845   |
3846   = note: `-D clippy::trivially-copy-pass-by-ref` implied by `-D warnings`
3847   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#trivially_copy_pass_by_ref
3848
3849error: aborting due to previous error
```

If it's required by the serialiser / it's not possible to pass by value, then you can add an annotation for clippy to ignore the `clippy::trivially-copy-pass-by-ref` warning in this function

client-rpc/src/rpc/transaction_rpc.rs
```diff
@@ -0,0 +1,153 @@
...
+}
+
+impl TransactionRpc for TransactionRpcImpl {
+    fn create_raw(
+        &self,
+        inputs: Vec<TxoPointer>,
+        outputs: Vec<TxOut>,
+        view_keys: Vec<PublicKey>,
+    ) -> Result<RawTransaction> {
+        // let view_keys = view_keys
```
tomtau: I suppose this can be deleted?

chain-core/src/tx/data/attribute.rs
calvinaco: Yes, yesterday I tried to use pass by value and it seems it does not work. I will see if it is required and ignore the warning if needed.

client-rpc/src/rpc/transaction_rpc.rs
calvinaco: Yes, I will delete it now. Thanks

PR #293, opened 2019-08-19
Problem: client not continuously synced (cro-195)

leejw51:
how to run)
./client-rpc --network-id ab

enter wallets)
enter wallet a
enter wallet b
enter nothing

then it will sync multiple wallets


client-rpc/src/rpc/websocket_rpc.rs
```diff
@@ -0,0 +1,126 @@
...
+use futures::future::Future;
+use futures::sink::Sink;
+use futures::stream::Stream;
+use futures::sync::mpsc;
+use mpsc::Sender;
+use std::thread;
+use websocket::result::WebSocketError;
+use websocket::{ClientBuilder, OwnedMessage};
+
+pub const CONNECTION: &str = "ws://localhost:26657/websocket";
```
tomtau: This would ideally be configurable rather than hard-coded -- e.g. via a command-line argument or as an environment variable.

client-rpc/src/server.rs
```diff
@@ -100,6 +102,91 @@ impl Server {
...
+                }
+            }
+            Err(_b) => default.to_string(),
+        }
+    }
+
+    pub fn start_websocket(&mut self, storage: SledStorage) -> Result<()> {
+        println!("web socket");
+        let mut wallet_infos: WalletInfos = vec![];
+        let tendermint_client = RpcClient::new(&self.tendermint_url);
```
tomtau: Is this one needed if all the methods could be called over websockets?

leejw51: it's necessary to receive callbacks,
also other restful api don't have to use websockets for processing requests.

leejw51:
i'll

> I'll wait for the RPC changes before reviewing this. Also, can you check if [jsonrpc-ws-client](https://crates.io/crates/jsonrpc-ws-client) will help you to get rid of some boilerplate code related to websockets in this PR?

i will add rpc first.
i need to control flow-controls in low level.

client-rpc/src/rpc/websocket_rpc.rs
leejw51: i will add as runtime argument

client-rpc/src/server.rs
tomtau: they are not restful -- both websockets and plain HTTP requests use JSON-RPC. So it's just that there are two ways of transport.
But this could be addressed in a different PR.

leejw51:
rpc added)
you can add wallet runtime.
when there is no wallet, it will wait.
you can add wallet in runtime.

request is as follows)
--------------------------------------------------
{
	"method": "sync_continuous",
	"jsonrpc": "2.0",
	"params": [
		{
		"name": "a",
		"passphrase": "1"
	}],
	"id": "sync_continuous"
}

leejw51:
removed cli

client-rpc/src/rpc/websocket_rpc.rs
leejw51: ok

client-rpc/src/rpc/websocket_core.rs
```diff
@@ -0,0 +1,382 @@
...
+
+    // tx channel for this thread
+    pub fn get_queue(&self) -> Sender<OwnedMessage> {
+        self.my_sender.clone()
+    }
+
+    // because everything is done via channel
+    // no mutex is necessary
+    // wallet can be added in runtime
+    pub fn add_wallet(&mut self, name: String, passphrase2: String) -> JsonResult<()> {
```
tomtau: prefer to take `passphrase: SecUtf8` -- passphrases shouldn't be passed around as plain strings

client-rpc/src/rpc/websocket_core.rs
```diff
@@ -0,0 +1,382 @@
...
+        };
+
+        self.wallets.push(info.clone());
+        println!("wallets length {}", self.wallets.len());
+        Ok(())
+    }
+
+    // Value is given from websocket_rpc
+    // received
+    fn do_parse(&mut self, value: Value) -> Option<()> {
```
tomtau: Instead of this manual parsing, it'll be easier to define the datatypes that capture the information received from RPC, use serde deserializer auto-derivation and just use `serde_json::...` -- just like it's done elsewhere in client-rpc, e.g.: https://github.com/crypto-com/chain/blob/156bd0299f76b3567016079b82a4860324f0a45b/client-rpc/src/server.rs#L167

client-rpc/src/rpc/websocket_rpc.rs
```diff
@@ -0,0 +1,154 @@
...
+        let _child = thread::spawn(move || {
+            core.start();
+        });
+
+        assert!(self.core.is_some());
+    }
+
+    // activate tokio websocket 
+    pub fn run_network(&mut self) {
+        println!("Connecting to {}", self.websocket_url);
```
tomtau: I'm not sure if the client-rpc uses the env_logger, but if it does, it'll be better to print these using an appropriate log level (e.g. `info!` here?)

client-rpc/src/rpc/websocket_rpc.rs
```diff
@@ -0,0 +1,154 @@
...
+use client_index::BlockHandler;
+use futures::future::Future;
+use futures::sink::Sink;
+use futures::stream::Stream;
+use futures::sync::mpsc;
+use std::thread;
+use websocket::result::WebSocketError;
+use websocket::{ClientBuilder, OwnedMessage};
+
+// this handles low level network connection
```
tomtau: The convention in Rust is that `/// comment` is used for top-level (on consts, structs, functions...) comments that would be picked up by `cargo docs` if you generate the HTML documentation.
`// comment` is used for inline comments inside code blocks

client-rpc/src/server.rs
```diff
@@ -100,6 +104,38 @@ impl Server {
...
+            wallet_infos,
+            tendermint_client,
+            storage.clone(),
+            block_handler,
+            wallet_client,
+        );
+        assert!(web.core.is_some());
+        self.websocket_queue = Some(web.core.as_mut().unwrap().clone());
+
+        let _child = thread::spawn(move || {
```
devashishdxt: Remove `let` here if it isn't used.

client-rpc/src/rpc/websocket_rpc.rs
```diff
@@ -0,0 +1,154 @@
...
+
+        let _child = thread::spawn(move || {
+            core.start();
+        });
+
+        assert!(self.core.is_some());
+    }
+
+    // activate tokio websocket 
+    pub fn run_network(&mut self) {
```
devashishdxt: I think the whole code in this function is boilerplate and can be removed if we use [jsonrpc-ws-client](https://crates.io/crates/jsonrpc-ws-client).

I think it is a good idea to use crates which are well tested and used by many. Maybe we can do this in a future PR. Any comments @tomtau?

client-rpc/src/rpc/websocket_core.rs
```diff
@@ -0,0 +1,382 @@
...
+            .unwrap()
+    }
+
+    // write block to internal database
+    fn do_save_block_to_chain(&mut self, value: &Value, kind: &str) {
+        if self.wallets.is_empty() {
+            return;
+        }
+        // restore as object
+        let height: u64 = value["block"]["header"]["height"]
```
devashishdxt: Use `Block` struct in `client-common` instead of parsing manually.

client-rpc/src/rpc/websocket_rpc.rs
tomtau: client-rpc already uses jsonrpc-core-client (or does it? it has it in cargo dependencies at least) which abstracts over the transport (tls, http or ws) -- so it'll make sense to use jsonrpc-ws-client.

tomtau: but yeah, using jsonrpc-ws-client should reduce some of the concurrency boilerplate

leejw51: core and rpc networking communicates with channel,
that means calling function is not allowed, only serialized strings.

client-rpc/src/rpc/websocket_core.rs
leejw51: ok

client-rpc/src/rpc/websocket_core.rs
leejw51: i agree, too.
but used this for inter thread communiation.

api cannot be called directly in other threads.

client-rpc/src/rpc/websocket_rpc.rs
leejw51: ok

client-rpc/src/rpc/websocket_rpc.rs
leejw51: ok

client-rpc/Cargo.toml
```diff
@@ -21,6 +22,10 @@ structopt = "0.2"
 chrono = { version = "0.4", features = ["serde"] }
 hex = "0.3.2"
 secstr = { version = "0.3.2", features = ["serde"] }
+websocket = "0.23.0"
+futures = "0.1.28"
+tokio = "0.1.22"
+quest = "0.3"
```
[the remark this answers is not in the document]
leejw51: if current low level websocket is used, i need this crates,
also that crates are shared by other modules.

client-rpc/src/rpc/sync_rpc.rs
```diff
@@ -28,6 +33,7 @@ where
 {
     client: T,
     synchronizer: ManualSynchronizer<S, C, H>,
+    websocket_queue: Mutex<Option<std::sync::mpsc::Sender<OwnedMessage>>>,
```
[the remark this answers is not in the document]
leejw51: without mutex, i cannot sedn queue by compiling error.

client-rpc/src/rpc/sync_rpc.rs
```diff
@@ -54,6 +60,23 @@ where
             .sync_all(&staking_addresses, &view_key, &private_key)
             .map_err(to_rpc_error)
     }
+
+    fn sync_continuous(&self, request: WalletRequest) -> Result<String> {
+        let data = json!({
```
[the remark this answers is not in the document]
leejw51: ok

client-rpc/src/rpc/sync_rpc.rs
```diff
@@ -54,6 +60,23 @@ where
...
+            "wallet": request,
+        });
+        let ret = "OK".to_string();
+        {
+            let sendqoption = self.websocket_queue.lock().unwrap();
+            assert!(sendqoption.is_some());
+            let sendq = sendqoption.as_ref().unwrap();
+            sendq
+                .send(OwnedMessage::Text(serde_json::to_string(&data).unwrap()))
+                .unwrap();
```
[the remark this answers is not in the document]
leejw51: it's not processed

client-rpc/src/rpc/sync_rpc.rs
```diff
@@ -67,9 +90,14 @@ where
         SyncRpcImpl {
             client,
             synchronizer,
+            websocket_queue: Mutex::new(None),
         }
     }
 
+    pub fn set_websocket_queue(&mut self, q: std::sync::mpsc::Sender<OwnedMessage>) {
```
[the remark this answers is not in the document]
leejw51: it doesn't exist in runtime.

client-rpc/src/server.rs
leejw51: ok

client-rpc/src/rpc/websocket_core.rs
leejw51: ok

leejw51:
overall, continuous sync is connected session.
it's easier in handle in lower level with full tokio send, recv queue.
so in various situations, dealing erros in socket level is easier, i think.

what do you think? @tomtau  @devashishdxt

devashishdxt:
> overall, continuous sync is connected session.
> it's easier in handle in lower level with full tokio send, recv queue.
> so in various situations, dealing errors in socket level is easier, i think.
> 
> for example
> 
> 1. when server is not reacting
> 2. not graceful closing of sockets
> 3. and other stranage network errors
> 
> if we use high level api, usual socket level api is protected, hard to control.
> 
> what do you think? @tomtau @devashishdxt

Yes. If we need low level control. It is ok to use this crate and manage all the communication ourselves.

leejw51:
i'm fixing issues first.

https://github.com/paritytech/jsonrpc
while doging that, I will test json-rpc first,
i need channel to share with core and rpc.
i'll check whether that's feasible.(maybe possible)

client-rpc/Cargo.toml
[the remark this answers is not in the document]
tomtau: even `quest`?

client-rpc/src/rpc/websocket_rpc.rs
tomtau: https://crates.io/crates/jsonrpc-ws-client handles that for you

client-rpc/src/rpc/websocket_core.rs
```diff
@@ -0,0 +1,378 @@
...
+            .unwrap()
+    }
+
+    // write block to internal database
+    fn do_save_block_to_chain(&mut self, block: Block, kind: &str) {
+        if self.wallets.is_empty() {
+            return;
+        }
+        // restore as object
+        let height: u64 = block.block.header.height.parse().unwrap();
```
devashishdxt:
```suggestion
        let height: u64 = block.height().unwrap();
```

client-rpc/src/rpc/sync_rpc.rs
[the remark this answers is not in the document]
leejw51: added return message

client-rpc/src/rpc/sync_rpc.rs
```diff
@@ -1,22 +1,27 @@
...
+use websocket::OwnedMessage;
 #[rpc]
 pub trait SyncRpc: Send + Sync {
     #[rpc(name = "sync")]
     fn sync(&self, request: WalletRequest) -> Result<()>;
 
     #[rpc(name = "sync_all")]
     fn sync_all(&self, request: WalletRequest) -> Result<()>;
+
+    #[rpc(name = "sync_continuous")]
```
calvinaco: Perhaps `sync_addWallet` will be more descriptive of what it is doing?

client-rpc/src/rpc/websocket_core.rs
```diff
@@ -0,0 +1,376 @@
...
+    // tx channel for this thread
+    pub fn get_queue(&self) -> Sender<OwnedMessage> {
+        self.my_sender.clone()
+    }
+
+    // because everything is done via channel
+    // no mutex is necessary
+    // wallet can be added in runtime
+    pub fn add_wallet(&mut self, name: String, passphrase: SecUtf8) -> JsonResult<()> {
+        log::info!("add_wallet ***** {} {}", name, passphrase);
```
calvinaco: I think passphrase should not be printed here? Although it is protected by SecUtf8 already

client-rpc/src/rpc/websocket_core.rs
```diff
@@ -0,0 +1,376 @@
...
+                current,
+                self.max_height
+            );
+            return;
+        }
+        log::info!("save block height={} kind={}", height, kind);
+        let _ = self.write_block(height, &block);
+    }
+
+    // low level block processing
```
tomtau: `//` vs `///`

client-rpc/src/rpc/websocket_core.rs
leejw51: ok

leejw51:
> there are a lot of `unwrap` -- could they be replaced by proper error returning/handling (if appropriate) or `expect` (so that if the error isn't recoverable, it at least panics with some message saying what went wrong?)

ok
i will change error handling or expect

client-rpc/src/rpc/sync_rpc.rs
leejw51: ok

client-rpc/Cargo.toml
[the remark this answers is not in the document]
leejw51: quest will be removed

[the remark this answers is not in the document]
leejw51: it is used for entering wallet pass-phrase , but it is no longer used.

client-rpc/src/rpc/websocket_rpc.rs
leejw51: i'm making test code for this.

leejw51: request makes channel for each request, and
subscription handling is not documented well.
so after researching issues, i'll comment the more detail.

leejw51:
> > overall, continuous sync is connected session.
> > it's easier in handle in lower level with full tokio send, recv queue.
> > so in various situations, dealing errors in socket level is easier, i think.
> > for example
> > 
> > 1. when server is not reacting
> > 2. not graceful closing of sockets
> > 3. and other stranage network errors
> > 
> > if we use high level api, usual socket level api is protected, hard to control.
> > what do you think? @tomtau @devashishdxt
> 
> Yes. If we need low level control. It is ok to use this crate and manage all the communication o

> > overall, continuous sync is connected session.
> > it's easier in handle in lower level with full tokio send, recv queue.
> > so in various situations, dealing errors in socket level is easier, i think.
> > for example
> > 
> > 1. when server is not reacting
> > 2. not graceful closing of sockets
> > 3. and other stranage network errors
> > 
> > if we use high level api, usual socket level api is protected, hard to control.
> > what do you think? @tomtau @devashishdxt
> 
> Yes. If we need low level control. It is ok to use this crate and manage all the communication ourselves.

using jsonrpc on websocket is good idea.  i agree.
i will checking with subscription, and calls .
and overall stability.

client-rpc/src/rpc/websocket_core.rs
leejw51: ok

leejw51:
fixed unwrap and others.
now working on json rpc.

jsonrpc working flow is a little different,
i may re-code flow structure

leejw51:
ok
i'll fix on clippy.

for jsonrpc client websocket,
i succeed in using call method,
but subscription stream doesn't work as i thought.

source code is as below)
```
use futures::future::lazy;
use futures::future::Future;
use futures::sink::Sink;
use futures::stream::Stream;
use futures::sync::mpsc;

use futures::sync::oneshot;

use futures::future::ok;
use futures::sync::mpsc::Receiver;
use futures::sync::mpsc::Sender;
use futures::try_ready;
use futures::{Async, Poll};
use jsonrpc_client_transports::transports::ws::connect;
use jsonrpc_client_transports::RawClient;
use jsonrpc_client_transports::RpcChannel;
use jsonrpc_client_transports::RpcError;
use jsonrpc_client_transports::SubscriptionStream;
use jsonrpc_core::types::params::Params;
use serde_json::json;
use serde_json::map::Map;
use serde_json::Value;
use std::fmt;
use std::io::Read;
use std::thread;
use std::time;
use tokio::io;
use tokio::net::{tcp::ConnectFuture, TcpStream};
use tokio::runtime::Runtime;
use websocket::result::WebSocketError;
use websocket::OwnedMessage;
const CONNECTION: &'static str = "ws://localhost:26657/websocket";
#[derive(Clone)]
struct MyClient(RawClient);

impl From<RpcChannel> for MyClient {
    fn from(channel: RpcChannel) -> Self {
        MyClient(channel.into())
    }
}

pub struct ProcessFuture<T> {
    stream: T,
}
impl<T> ProcessFuture<T> {
    fn new(stream: T) -> ProcessFuture<T> {
        ProcessFuture { stream }
    }
}
impl<T> Future for ProcessFuture<T>
where
    T: Stream,
{
    type Item = ();
    type Error = T::Error;

    fn poll(&mut self) -> Poll<(), Self::Error> {
        println!("polling");
        loop {
            match try_ready!(self.stream.poll()) {
                Some(_value) => println!("got value"),

                None => {}
            };
        }
        Ok(Async::Ready(()))
    }
}

fn main() {
    let (channel_tx, channel_rx): (Sender<Value>, Receiver<Value>) = mpsc::channel(0);
    let mut channel_sink = channel_tx.clone().wait();
    let mut rt = Runtime::new().unwrap();
    let a = connect::<MyClient>(CONNECTION);
    let b: MyClient = rt.block_on(a.unwrap()).unwrap();
    println!("connected..");

    let mut map = Map::new();
    map.insert("query".to_string(), "tm.event='NewBlock'".into());
    let fut = b.0.subscribe("subscribe", Params::Map(map), "NewBlock", "");
    let stream: SubscriptionStream = rt.block_on(fut).unwrap();

    println!("subscribed ok!");
    let w = lazy(|| {
        for i in 0..4 {
            tokio::spawn(lazy(move || {
                loop {
                    thread::sleep(time::Duration::from_millis(1000));
                    //println!("Hello from task {}", i);
                }
                Ok(())
            }));
        }

        Ok(())
    });
    let runner = stream
        .into_future()
        .map(move |(result, _)| {
            println!("subscription stream map {:?}", result);
        })
        .map_err(|_| {
            panic!("Expected message not received.");
        })
        .join(w);
    rt.block_on(runner);
    //tokio::run();
    println!("wait================================");
    loop {
        thread::sleep(time::Duration::from_millis(1000));
    }
    println!("done");
}

```

calvinaco:
> One more warning:
> 
> ```
> error: unused `futures::stream::select::Select` in tuple element 0 that must be used
> 3870   --> client-rpc/src/rpc/websocket_rpc.rs:163:9
> 3871    |
> 3872163 |         runtime.block_on(runner).expect("tokio block_on");
> 3873    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> 3874    |
> 3875    = note: `-D unused-must-use` implied by `-D warnings`
> 3876    = note: streams do nothing unless polled
> 3877
> 3878error: aborting due to previous error
> ```
> 
> Plus, the client continuous sync seems to break the integration test @calvinaco: https://travis-ci.org/crypto-com/chain/jobs/577662715#L4921

I am re-running the test. and see if the problem persists.

leejw51crypto:
ok

PR #299, opened 2019-08-20
Problem: (CRO-339) Missing staking state RPC

calvinaco:
Solution: Added staking state RPC


client-rpc/src/rpc/staking_rpc.rs
```diff
@@ -100,6 +103,12 @@ where
             .map_err(to_rpc_error)
     }
 
+    fn state(&self, request: WalletRequest, address: StakedStateAddress) -> Result<StakedState> {
```
leejw51: how about get_state ?

leejw51: or get_staked_state

PR #310, opened 2019-08-22
Problem: (CRO-224) Missing network ops integration tests

calvinaco:
Solution: Added network ops integration tests


calvinaco:
Seems there is some minor problem when the env is brought up by Docker, I will investigate.

PR #313, opened 2019-08-23
Problem: (CRO-359) Client does not fast forward when app hash is same

devashishdxt:
**Solution**:
1. Client will now fast forward if current app hash matches last app hash stored in client.
2. Client will fast forward each batch of requests (default is 20) if the batch does not change app hash (it will still make `block` RPC call).

**Note**: Because this PR changes internal storage structure of `last_block_height` and `last_app_hash`, every client will have to do a `sync --force`.

### DIFF
--- a/client-index/src/synchronizer.rs
+++ b/client-index/src/synchronizer.rs
@@ -3,7 +3,7 @@ use itertools::Itertools;
 
 use chain_core::state::account::StakedStateAddress;
 use chain_tx_filter::BlockFilter;
-use client_common::tendermint::types::Block;
+use client_common::tendermint::types::{Block, BlockResults, Status};
 use client_common::tendermint::Client;
 use client_common::{BlockHeader, PrivateKey, PublicKey, Result, Storage, Transaction};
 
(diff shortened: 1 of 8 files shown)


client-index/src/synchronizer.rs
```diff
@@ -48,37 +48,44 @@ where
...
         let last_block_height = self.global_state_service.last_block_height(view_key)?;
-        let current_block_height = self.client.status()?.last_block_height()?;
+        let current_block_height = status.last_block_height()?;
 
         // Send batch RPC requests to tendermint in chunks of `batch_size` requests per batch call
         for chunk in ((last_block_height + 1)..=current_block_height)
             .chunks(batch_size.unwrap_or(DEFAULT_BATCH_SIZE))
             .into_iter()
         {
+            if self.fast_forward_status(staking_addresses, view_key, private_key, &status)? {
```
tomtau: would it be possible to give a command-line option, e.g. `--fast` or `--slow`, so that one can still do the previous block-by-block check without the fast-forward?

devashishdxt: Is there any reason why someone would want to do that?

tomtau: for sanity

PR #314, opened 2019-08-23
Problem: (CRO-359 CRO-319) Synchronization progress is not visible in client-cli

devashishdxt:
Solution: Added a progress bar


Cargo.lock
```diff
@@ -1630,6 +1636,18 @@ dependencies = [
  "winapi 0.3.7 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
+[[package]]
+name = "pbr"
+version = "1.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "kernel32-sys 0.2.2 (registry+https://github.com/rust-lang/crates.io-index)",
```
tomtau: a bit worried about the dependency bloat -- but there's a separate issue for that

PR #327, opened 2019-08-28
Problem: (CRO-297) Missing list public keys RPC

calvinaco:
Solution: Added wallet_listPublicKeys RPC


client-rpc/src/rpc/wallet_rpc.rs
```diff
@@ -34,6 +34,15 @@ pub trait WalletRpc: Send + Sync {
     #[rpc(name = "wallet_list")]
     fn list(&self) -> Result<Vec<String>>;
 
+    #[rpc(name = "wallet_listPublicKeys")]
+    fn list_public_keys(&self, request: WalletRequest) -> Result<Vec<PublicKey>>;
```
devashishdxt: I'd recommend to convert `public_keys` to `String` before returning (similar to what we do before returning addresses).

PR #331, opened 2019-08-29
Problem: no option to choose the network initialisation (testnet/devnet...) in client-cli / rpc(cro-367)

leejw51crypto:
network initialization option
mainnet
testnet
devnet


chain-core/src/init/network.rs
```diff
@@ -71,8 +71,8 @@ pub fn get_bech32_human_part() -> &'static str {
 
 mod chosen_network {
     use super::*;
-    pub static mut NETWORK: Network = Network::Devnet;
-    pub static mut NETWORK_ID: u8 = 0 as u8;
+    pub static mut NETWORK: Network = Network::Testnet;
+    pub static mut NETWORK_ID: u8 = 0x42 as u8;
```
devashishdxt: Is there any specific reason to change these defaults?

client-rpc/src/main.rs
```diff
@@ -37,6 +37,15 @@ pub(crate) struct Options {
     )]
     network_id: String,
 
+    #[structopt(
+        name = "network-type",
+        short = "i",
+        long,
+        default_value = "test",
+        help = "Network Type (main, test, deve)"
```
devashishdxt:
```suggestion
        help = "Network Type (main, test, dev)"
```

client-rpc/src/main.rs
```diff
@@ -37,6 +37,15 @@ pub(crate) struct Options {
     )]
     network_id: String,
 
+    #[structopt(
+        name = "network-type",
+        short = "i",
+        long,
+        default_value = "test",
```
devashishdxt: I think the default value should be `dev`

client-rpc/src/server.rs
```diff
@@ -49,6 +50,13 @@ impl Server {
     pub(crate) fn new(options: Options) -> Result<Server> {
         let network_id =
             hex::decode(&options.network_id).context(ErrorKind::SerializationError)?[0];
+        let network_type = options.network_type;
+        match &network_type[..4] {
+            "main" => init_network(Network::Mainnet),
```
tomtau: `init_network` isn't public anymore.

What about:

1. make `options.network_id` optional argument
2.  use the constants from chain_core and `init_chain_id`: 
- `"main" => init_chain_id(MAINNET_CHAIN_ID)`
- `"test" => init_chain_id(TESTNET_CHAIN_ID)`
- for devnet, require `options.network_id` and do `init_chain_id(format!("devnet-{}", options.network_id))`
3. for the RPC server, use `get_network_id()` from chain-core (after it's initialised)

What do you think?

chain-core/src/init/network.rs
leejw51crypto: no specific reason
devnet is also good.
for devnet
how about 0xab?
@tomtau

client-rpc/src/main.rs
leejw51crypto: ok

chain-core/src/init/network.rs
tomtau: just keep it 0, it'll mean that it wasn't initialised

tomtau: for devnet, anyone can choose any number

client-rpc/src/server.rs
leejw51crypto: not bad
so it's like this

--network-type main 
or
--network-type test
or  automaticall dev mode, if network type is not specified
--network-id=ab

chain-core/src/init/network.rs
leejw51crypto: ok, i will change back

PR #345, opened 2019-09-04
Problem: (CRO-374) CLI uses base units for amounts

devashishdxt:
**Solution**: Changed client to accept amount in CRO

**Note**: As a long term solution to this, we should change `FromStr` implementation of `Coin`.

### DIFF
--- a/client-cli/src/command/transaction_command.rs
+++ b/client-cli/src/command/transaction_command.rs
@@ -7,7 +7,6 @@ use structopt::StructOpt;
 use unicase::eq_ascii;
 
 use chain_core::common::{Timespec, HASH_SIZE_256};
-use chain_core::init::coin::Coin;
 use chain_core::state::account::{StakedStateAddress, StakedStateOpAttributes};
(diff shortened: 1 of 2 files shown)


leejw51crypto:
how about force using .0 or .00
decimal point for numbers?

to remove mis-understanding completely

devashishdxt:
> how about force using .0 or .00
> decimal point for numbers?
> 
> to remove mis-understanding completely

@tomtau What do you think?

tomtau:
> > how about force using .0 or .00
> > decimal point for numbers?
> > to remove mis-understanding completely
> 
> @tomtau What do you think?

How is it in the Angular UI sample wallet? I think either way is fine as long as it's consistent among human interfaces

devashishdxt:
> > > how about force using .0 or .00
> > > decimal point for numbers?
> > > to remove mis-understanding completely
> > 
> > 
> > @tomtau What do you think?
> 
> How is it in the Angular UI sample wallet? I think either way is fine as long as it's consistent among human interfaces

Current implementation is same as Angular UI

PR #346, opened 2019-09-05
Problem: autosync unlock request doesn't check if the wallet was already added to auto sync (cro-395)

leejw51crypto:
make inserting wallet as upsert to prevent adding redundant wallets


client-index/src/auto_sync_core.rs
```diff
@@ -217,7 +217,16 @@ where
     pub fn get_queue(&self) -> Sender<OwnedMessage> {
         self.my_sender.clone()
     }
-
+    fn update_wallet(&mut self, info: &WalletInfo) -> Result<()> {
+        for i in 0..self.wallets.len() {
```
devashishdxt: Instead of checking duplicates in a `Vec`. Can we have a `HashSet` or `BTreeSet` of `WalletInfo`?

leejw51crypto: order is important, so need to use vec or list

devashishdxt: Why is order important?

leejw51crypto:
agree, i'll change to btreemap

client-index/src/auto_sync_core.rs
leejw51crypto: yes , BTreeMap will be good

leejw51crypto:
sorry, it will work now

PR #351, opened 2019-09-05
Problem: autosync unlock request doesn't check if the wallet was already added to auto sync(cro-395)

leejw51crypto:


leejw51crypto:
working on this

PR #352, opened 2019-09-05
Problem: no retry mechanism in auto sync (cro-394)

leejw51crypto:


leejw51crypto:
fixed

client-index/src/auto_synchronizer.rs
```diff
@@ -92,45 +82,55 @@ impl AutoSynchronizer {
...
+                            OwnedMessage::Ping(d) => Some(OwnedMessage::Pong(d)),
+                            _ => None,
+                        })
+                        .select(channel_rx.map_err(|_| WebSocketError::NoDataAvailable))
+                        .forward(sink)
+                });
+            match runtime.block_on(runner) {
+                Ok(_a) => {}
+                Err(b) => {
+                    println!("connection fail {}", b);
```
tomtau: `warn!` or `error!` for logging would be preferable, as that would highlight that something went wrong

client-index/src/auto_synchronizer.rs
```diff
@@ -92,45 +82,55 @@ impl AutoSynchronizer {
...
+                    std::thread::sleep(std::time::Duration::from_millis(1000));
```
tomtau: this will spam the error log -- the retry time could be extended on each retry attempt and/or there could be a maximum number of retries (after which it'll crash the whole client-rpc process)

client-index/src/auto_synchronizer.rs
leejw51crypto: ok, i'll fix

PR #365, opened 2019-09-10
Problem: (CRO-414) Creating transfer address also creates a staking address

devashishdxt:
Solution: Separated key storage for staking and transfer public keys

Note: This commit changes client's storage data model. We should tag master branch before pushing this change. Old data will not work with these changes.


devashishdxt:
> if this breaks the client, perhaps bump the version numbers to `0.1.0` and create a changelog entry --
> 
> ```
> *Unreleased*
> 
> ## v0.1.0
> 
> ### Features
> ...
> ### Bugfixes
> ...
> ```

It does not break client. It is not compatible with old storage data. Should we create a new release for that also?
Also, changes for grouping transaction history with transaction ID will bring a lot of incompatibility with past client versions. My suggestion is to bump the version only when all the breaking changes are in master.

Any comments @tomtau @leejw51 @calvinaco ?

tomtau:
> > if this breaks the client, perhaps bump the version numbers to `0.1.0` and create a changelog entry --
> > ```
> > *Unreleased*
> > 
> > ## v0.1.0
> > 
> > ### Features
> > ...
> > ### Bugfixes
> > ...
> > ```
> 
> It does not break client. It is not compatible with old storage data. Should we create a new release for that also?
> Also, changes for grouping transaction history by transaction ID will bring a lot of incompatibility with past client versions. My suggestion is to bump the version only when all the breaking changes are in master.
> 
> I have added above changes to `CHANGELOG` but haven't bumped up the version.
> 
> Any comments @tomtau @leejw51 @calvinaco ?

Well, it breaks the storage data format / public contract -- a non-breaking change would be to have some migration command / way to import old data.
It's fine to bump the version numbers now and later spawn a release branch from it. If there are some "v0.0-compatible" changes after this one, they can be cherry-picked to the "v0.0" branch and there could be "v0.0.2" if that's desired.

CHANGELOG.md
```diff
@@ -5,6 +5,13 @@
 The release is an incomplete alpha version meant to be deployed on the first iteration of the public testnet.
 There are no guarantees on future API and binary compatibility at this stage.
 
+## v0.0.2 (Unreleased)
```
tomtau: 0.1.0

CHANGELOG.md
```diff
@@ -5,6 +5,13 @@
...
+
+### Features
+
+### Bug Fixes
+* Creating a transfer address also creates a staking address
```
tomtau:
```suggestion
* [365](https://github.com/crypto-com/chain/pull/365) Creating a transfer address also creates a staking address
```

devashishdxt:
> > > if this breaks the client, perhaps bump the version numbers to `0.1.0` and create a changelog entry --
> > > ```
> > > *Unreleased*
> > > 
> > > ## v0.1.0
> > > 
> > > ### Features
> > > ...
> > > ### Bugfixes
> > > ...
> > > ```
> > 
> > 
> > It does not break client. It is not compatible with old storage data. Should we create a new release for that also?
> > Also, changes for grouping transaction history by transaction ID will bring a lot of incompatibility with past client versions. My suggestion is to bump the version only when all the breaking changes are in master.
> > I have added above changes to `CHANGELOG` but haven't bumped up the version.
> > Any comments @tomtau @leejw51 @calvinaco ?
> 
> Well, it breaks the storage data format / public contract -- a non-breaking change would be to have some migration command / way to import old data.
> It's fine to bump the version numbers now and later spawn a release branch from it. If there are some "v0.0-compatible" changes after this one, they can be cherry-picked to the "v0.0" branch and there could be "v0.0.2" if that's desired.

Changed the version to `0.1.0`

client-rpc/src/rpc/wallet_rpc.rs
```diff
@@ -126,20 +128,20 @@ where
         self.client.wallets().map_err(to_rpc_error)
     }
 
-    fn list_public_keys(&self, request: WalletRequest) -> Result<Vec<PublicKey>> {
+    fn list_public_keys(&self, request: WalletRequest) -> Result<BTreeSet<PublicKey>> {
```
calvinaco: Will the BTreeSet serializable to JSON as an array?

devashishdxt: Yes

PR #371, opened 2019-09-11
Problem: too many logs for disconnection of client-rpc auto-sync(cro-404)

leejw51:
fix sync bug


leejw51:
for block explorer, this is quick fix.

PR #372, opened 2019-09-11
Problem: (CRO-423) Integration test does not index all transaction tag

calvinaco:
Solution: Configure Tendermint to index all transaction tags

### DIFF
--- a/integration-tests/prepare.sh
+++ b/integration-tests/prepare.sh
@@ -64,7 +64,7 @@ function init_tendermint() {
 
 # @argument Tendermint directory
 function index_all_tags() {
-    print_step "Disabling empty blocks for ${1}"
+    print_step "Enable tag indexing for ${1}"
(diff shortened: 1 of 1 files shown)


integration-tests/prepare.sh
```diff
@@ -58,6 +58,14 @@ function init_tendermint() {
     if [ ! -z "${CI}" ]; then
         sudo chmod -R 777 ./tendermint
     fi
+
+    index_all_tags "tendermint"
+}
+
+# @argument Tendermint directory
+function index_all_tags() {
+    print_step "Disabling empty blocks for ${1}"
```
tomtau:
```suggestion
    print_step "Enable tag indexing for ${1}"
```

calvinaco: > is tag indexing required for integration test?

Not at the moment. I use it to test between other projects I am working on. So I enable it to facilities testing and development.

PR #373, opened 2019-09-12
Problem:(CRO-424) Auto-sync doesn't work correctly after adding wallet

leejw51:
Solution:
in process_text of auto_synchronizer,
it should return Ok, Err instead of directly close_connection.
because close_connection is called in caller "process_text".


leejw51crypto:
sorry, for complicated merging and incorrect-auto-sync.
I'll use 
```
git pull --rebase upstream master.
```
also i'll write more detail in description.
thanks!

calvinaco:
👍🏻 Thanks

PR #386, opened 2019-09-17
(CRO-378) Problem: GUI requires client-rpc

leejw51crypto:
Solution: use neon-binding to merge client-rpc binary


client-rpc/src/lib.rs
```diff
@@ -0,0 +1,120 @@
...
+        name = "websocket-url",
+        short,
+        long,
+        default_value = "ws://localhost:26657/websocket",
+        help = "Url for connecting with tendermint websocket RPC"
+    )]
+    websocket_url: String,
+}
+
+pub fn main() {
```
devashishdxt: I don't think we should name this function `main`

leejw51crypto: agree, "main" removed

leejw51crypto:
agree, i'll move to main and export it

PR #387, opened 2019-09-17
Problem: (CRO-333) Transactions are not grouped by transaction id in client storage

devashishdxt:
Solution: Grouped transactions by transaction id on per wallet basis. Also, updated CLI.

---
This change is [Reviewable](https://reviewable.io/reviews/crypto-com/chain/387)


devashishdxt:
> could the integration test be fixed before / for this PR?

@tomtau @calvinaco Is there a ticket opened for that?

devashishdxt:
___
*[client-core/src/handler/default_transaction_handler.rs, line 55 at r2](https://reviewable.io/reviews/crypto-com/chain/387#-Loy2qOsAB8l2AB-4FLW:-Loy4ngKCDors3ZHEAN1:b-xzt1tg) ([raw file](https://github.com/crypto-com/chain/blob/9a629ca875441b9bdc196b5c21e0258f2804f04b/client-core/src/handler/default_transaction_handler.rs#L55)):*
<details><summary><i>Previously, tomtau (Tomas Tauber) wrote…</i></summary><blockquote>

what will happen if some of the inputs are missing in the wallet?
</blockquote></details>

This step tries to gather details of all the inputs present in a transaction. If details corresponding to any input is not present, it means that current wallet's view-key was not present in the source transaction. So, the details will be set to `None`.

devashishdxt:
___
*[client-core/src/handler/default_transaction_handler.rs, line 55 at r2](https://reviewable.io/reviews/crypto-com/chain/387#-Loy2qOsAB8l2AB-4FLW:-Loy5u7k5jKv63SNXF1f:b-6br9b1) ([raw file](https://github.com/crypto-com/chain/blob/9a629ca875441b9bdc196b5c21e0258f2804f04b/client-core/src/handler/default_transaction_handler.rs#L55)):*
<details><summary><i>Previously, tomtau (Tomas Tauber) wrote…</i></summary><blockquote>

So, in the later step, the calculated fee may be incorrect or it'll fail?
</blockquote></details>

Fee is only calculated when all the inputs belong to current wallet, i.e., it is an `OUTGOING` transaction. Fee is not calculated when a transaction is `INCOMING`, i.e., inputs do not belong to current wallet but outputs does.

devashishdxt:
___
*[client-core/src/handler/default_transaction_handler.rs, line 55 at r2](https://reviewable.io/reviews/crypto-com/chain/387#-Loy2qOsAB8l2AB-4FLW:-Loy6JGm8BtLC3PlKkPx:bebg2fq) ([raw file](https://github.com/crypto-com/chain/blob/9a629ca875441b9bdc196b5c21e0258f2804f04b/client-core/src/handler/default_transaction_handler.rs#L55)):*
<details><summary><i>Previously, devashishdxt (Devashish Dixit) wrote…</i></summary><blockquote>

Fee is only calculated when all the inputs belong to current wallet, i.e., it is an `OUTGOING` transaction. Fee is not calculated when a transaction is `INCOMING`, i.e., inputs do not belong to current wallet but outputs does.
</blockquote></details>

`OUTGOING` = Some coins are spent
`INCOMING` = Some coins are received

devashishdxt:
___
*[client-core/src/handler/default_transaction_handler.rs, line 55 at r2](https://reviewable.io/reviews/crypto-com/chain/387#-Loy2qOsAB8l2AB-4FLW:-Loy6aV1BZDOVYdKXj4Z:b-jvt154) ([raw file](https://github.com/crypto-com/chain/blob/9a629ca875441b9bdc196b5c21e0258f2804f04b/client-core/src/handler/default_transaction_handler.rs#L55)):*
<details><summary><i>Previously, devashishdxt (Devashish Dixit) wrote…</i></summary><blockquote>

`OUTGOING` = Some coins are spent
`INCOMING` = Some coins are received
</blockquote></details>

You can see implementation of `calculate_balance_change` for details of different cases.

PR #390, opened 2019-09-18
[Problem] Unable to build client-cli

lezzokafka:
Solution: Dependance change for secstr


devashishdxt:
> is this really a problem? recent CI builds https://travis-ci.org/crypto-com/chain/builds/586165026 are fine

Its a problem if you only try to build `client-cli`

PR #392, opened 2019-09-18
Problem:(CRO-403) auto-sync connection state not signalled to the API caller

leejw51crypto:
Solution: add sync info
add state

update connect information

refactoring

### DIFF
--- a/client-core/src/synchronizer/auto_sync.rs
+++ b/client-core/src/synchronizer/auto_sync.rs
@@ -19,7 +19,9 @@ use client_common::tendermint::Client;
 use client_common::Result;
 use client_common::Storage;
 
-use super::auto_sync_data::{AddWalletCommand, AutoSyncDataShared, RemoveWalletCommand};
+use super::auto_sync_data::{
+    AddWalletCommand, AutoSyncDataShared, AutoSyncInfo, RemoveWalletCommand,
+};
 use super::auto_synchronizer::AutoSynchronizer;
 use crate::BlockHandler;
 
@@ -42,7 +44,7 @@ impl AutoSync {
         storage: S,
         block_handler: H,
     ) {
-        let mut web = AutoSynchronizer::new(url);
+        let mut web = AutoSynchronizer::new(url, self.data.clone());
         web.run(client, storage, block_handler, self.data.clone());
         let websocket_queue = web.get_send_queue();
 
@@ -69,6 +71,21 @@ impl AutoSync {
         });
 
         self.send_json(data)
+    }
+
+    /// Get sync information
+    pub fn sync_info(&self) -> Result<AutoSyncInfo> {
+        let mut ret = AutoSyncInfo::default();
+
+        {
+            let data = self.data.lock().expect("get progress autosync lock");
+            ret.current_height = data.current_height;
+            ret.max_height = data.max_height;
+            ret.wallet = data.wallet.clone();
+            ret.connected = data.connected;
+            ret.state = data.state.clone();
+        }
+        Ok(ret)
     }
 
     /// Removes a wallet from auto-sync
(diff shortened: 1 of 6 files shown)


leejw51crypto:
output is as follows)

{
    "jsonrpc": "2.0",
    "result": {
        "connected": true,
        "current_height": 2774,
        "max_height": 4110,
        "state": "GetBlocks",
        "wallet": "b"
    },
    "id": "sync_info"
}

client-core/src/synchronizer/auto_sync_data.rs
```diff
@@ -20,6 +20,10 @@ pub struct AutoSyncData {
     pub max_height: u64,
     /// current syncing wallet name
     pub wallet: String,
+    /// connected via websocket
+    pub connected: bool,
+    /// sync state : getting block, waiting, etc.
+    pub state: String,
```
tomtau: what if `WebsocketState` is extended with one more variant "Disconnected", and here it could just be:
```
pub state: WebsocketState
```

instead of bool+String?

tomtau: And add `#[derive(Serialize)]` to `WebsocketState`

client-core/src/synchronizer/auto_sync.rs
```diff
@@ -71,6 +73,21 @@ impl AutoSync {
         self.send_json(data)
     }
 
+    /// Get sync information
+    pub fn sync_info(&self) -> Result<AutoSyncInfo> {
```
tomtau: `-> AutoSyncInfo`

it seems It won't fail (if the lock is poisoned, it'll panic which is ok)

client-core/src/synchronizer/auto_sync_data.rs
leejw51crypto: state is for core.
so that can be

connect_state: ConnectState, <- connecting, connected, disconnected
state: CoreState  <- get blocks, waiting, ready

client-core/src/synchronizer/auto_sync.rs
```diff
@@ -71,6 +73,21 @@ impl AutoSync {
...
+        let mut ret = AutoSyncInfo::default();
+
+        {
+            let data = self.data.lock().expect("get progress autosync lock");
```
tomtau: instead of copying everything here, what if there's `info: AutoSyncInfo` field in `self.data`, so one just do something like:
```
let data = self.data.lock().expect("get progress autosync lock");
data.info.clone()
```
?

client-core/src/synchronizer/auto_sync.rs
leejw51crypto: ok

client-core/src/synchronizer/auto_sync.rs
leejw51crypto: ok

client-core/src/synchronizer/auto_sync_data.rs
tomtau: Core state only applies if it's connected, so something like:
```
pub enum ConnectionState {
  Connecting,
  Connected(CoreState),
  Disconnected
}
```
?

leejw51: ok , got it

client-core/src/synchronizer/auto_sync_data.rs
```diff
@@ -20,6 +20,17 @@ pub struct AutoSyncData {
     pub max_height: u64,
     /// current syncing wallet name
     pub wallet: String,
```
calvinaco: Is it possible to also include a field of all the wallet being synced as well? I think this information is useful for the `sync_info` RPC

leejw51crypto: ok

PR #395, opened 2019-09-19
Problem: (CRO-403) auto-sync connection state not signalled to the API caller

leejw51crypto:
Solution: add sync_info api
add unlocked wallets in autosync info

add state


client-core/src/synchronizer/auto_sync_data.rs
```diff
@@ -20,6 +53,16 @@ pub struct AutoSyncData {
     pub max_height: u64,
     /// current syncing wallet name
     pub wallet: String,
+    /// all wallets
+    pub wallets_all: Vec<String>,
```
calvinaco: I suggest to change to `syncing_wallet/current_wallet` and `unlocked_wallets`.

leejw51crypto: ok, i'll change to 
current_wallet,
unlocked_wallets

PR #396, opened 2019-09-19
Problem: (CRO-443) ClientRPC and ClientCLI accept different chain ID format

calvinaco:
Solution: Standardize them to always provide Chain ID option


client-rpc/src/program.rs
```diff
@@ -93,18 +83,13 @@ pub fn find_string(args: &[String], target: &str) -> Option<usize> {
...
     env_logger::init();
-    // "~/Electron", ".", "--network-id", "ab", "--network-type", "test"]
+    // "~/Electron", ".", "--chain-id", "ab", "--network-type", "test"]
     let args: Vec<String> = env::args().collect();
     log::info!("args={:?}", args);
     let mut options = Options::from_iter(vec![""].iter());
-    if let Some(a) = find_string(&args, "--network-id") {
-        options.network_id = args[a + 1].clone()
+    if let Some(a) = find_string(&args, "--chain-id") {
+        options.chain_id = args[a + 1].clone()
```
calvinaco: @leejw51crypto I have changed this part which related to Electron, will it affect it?

leejw51crypto: chain_id = test...ab
network_id=  ab
network_type = test

leejw51crypto: so is  chain_id full id?,  or just changing name?

leejw51crypto: ok

calvinaco: chain_id is the full id, for reference

leejw51crypto: i'll modify electron as another pull request

leejw51crypto: yes

calvinaco:
> cargo fmt

- cargo fmt fixed
- I forgot to update the docker version of the integration tests, fix applied

calvinaco:
> `cargo fmt` + it seems the integration test fails: https://travis-ci.org/crypto-com/chain/jobs/586893015#L4780?

- cargo fmt fixed
- I forgot to update the docker version of the integration tests, fix applied

tomtau:
@calvinaco integration test still fails with `curl: (7) Failed to connect to 127.0.0.1 port 26659: Connection refused`

leejw51crypto:
also , remove --network-type

calvinaco:
> @calvinaco integration test still fails with `curl: (7) Failed to connect to 127.0.0.1 port 26659: Connection refused`

Another fix submitted, should be able to fix the test.

calvinaco:
> also , remove --network-type

Removed in `run_electron()`

tomtau:
@calvinaco `     Tendermint RPC error: account lookup failed: account not found
4818`

calvinaco:
> @calvinaco ` Tendermint RPC error: account lookup failed: account not found 4818`

Just saw it, it is weird because it does not happen when I run it locally yesterday. I am going to re-run the test locally.

PR #402, opened 2019-09-20
Problem: (CRO-391) Two transport layers for tendermint JSON-RPC client

devashishdxt:
Solution: Use websocket for JSON-RPC everywhere

Note: Currently, `AutoSynchronizer` creates its own websocket connection with tendermint. So, there'll be two separate websocket connections with tendermint.

### DIFF
--- a/client-cli/src/main.rs
+++ b/client-cli/src/main.rs
@@ -51,7 +51,7 @@ pub(crate) fn storage_path() -> String {
 #[inline]
 pub(crate) fn tendermint_url() -> String {
     std::env::var("CRYPTO_CLIENT_TENDERMINT")
-        .unwrap_or_else(|_| "http://localhost:26657/websocket".to_owned())
+        .unwrap_or_else(|_| "ws://localhost:26657/websocket".to_owned())
 }
 
 #[inline]
--- a/client-cli/src/main.rs
+++ b/client-cli/src/main.rs
@@ -51,7 +51,7 @@ pub(crate) fn storage_path() -> String {
 #[inline]
 pub(crate) fn tendermint_url() -> String {
     std::env::var("CRYPTO_CLIENT_TENDERMINT")
-        .unwrap_or_else(|_| "http://localhost:26657/".to_owned())
+        .unwrap_or_else(|_| "http://localhost:26657/websocket".to_owned())
 }
 
 #[inline]
--- a/client-common/src/tendermint/websocket_rpc_client.rs
+++ b/client-common/src/tendermint/websocket_rpc_client.rs
@@ -1,0 +1,377 @@
+#![cfg(feature = "websocket-rpc")]
+mod types;
+mod websocket_rpc_loop;
+
+pub use types::ConnectionState;
+
+use std::collections::HashMap;
+use std::iter;
+use std::sync::mpsc::{sync_channel, Receiver, SyncSender};
+use std::sync::{Arc, Mutex};
+use std::thread::{self, JoinHandle};
+use std::time::Duration;
+
+use rand::distributions::Alphanumeric;
+use rand::{thread_rng, Rng};
+use serde::Deserialize;
+use serde_json::{json, Value};
+use websocket::sender::Writer;
+use websocket::stream::sync::TcpStream;
+use websocket::OwnedMessage;
+
+use self::types::*;
+use crate::tendermint::types::*;
+use crate::tendermint::Client;
+use crate::{Error, ErrorKind, Result, ResultExt};
+
+const RESPONSE_TIMEOUT: Duration = Duration::from_secs(10);
+
+const WAIT_FOR_CONNECTION_SLEEP_INTERVAL: Duration = Duration::from_millis(200);
+const WAIT_FOR_CONNECTION_COUNT: u16 = 50;
+
+/// Tendermint RPC Client (uses websocket in transport layer)
+#[derive(Clone)]
+pub struct WebsocketRpcClient {
+    handle: Arc<Mutex<Option<JoinHandle<()>>>>,
+    connection_state: Arc<Mutex<ConnectionState>>,
+    websocket_writer: Arc<Mutex<Writer<TcpStream>>>,
+    channel_map: Arc<Mutex<HashMap<String, SyncSender<JsonRpcResponse>>>>,
+}
+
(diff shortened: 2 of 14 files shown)


client-cli/src/main.rs
```diff
@@ -51,7 +51,7 @@ pub(crate) fn storage_path() -> String {
 #[inline]
 pub(crate) fn tendermint_url() -> String {
     std::env::var("CRYPTO_CLIENT_TENDERMINT")
-        .unwrap_or_else(|_| "http://localhost:26657/".to_owned())
+        .unwrap_or_else(|_| "http://localhost:26657/websocket".to_owned())
```
tomtau: is the URL correct ? in auto sync, it was `ws://localhost:26657/websocket`?

tomtau: if the URL works with "http://" in the beginning -- for non-breaking changes (so that the client can be used in the same way as before), perhaps it can be assumed to be as before (without the "/websocket" path) and the client would then construct the URL with the "/websocket" path?

client-common/src/tendermint/websocket_rpc_client.rs
```diff
@@ -0,0 +1,366 @@
...
+    fn request(&self, method: &str, params: &[Value]) -> Result<Value> {
+        let (id, channel_receiver) = self.send_request(method, params)?;
+        self.receive_response(method, params, &id, channel_receiver)
+    }
+
+    /// Sends RPC requests for a batch.
+    ///
+    /// # Note
+    ///
+    /// This does not use batch JSON-RPC requests but makes multiple single JSON-RPC requests in parallel.
```
tomtau: why not to use the batch request?

client-cli/src/main.rs
devashishdxt: It works with `http://`. But, I think the correct prefix is `ws://`. Changed it in code as well as in integration tests.

client-common/src/tendermint/websocket_rpc_client/websocket_rpc_loop.rs
```diff
@@ -0,0 +1,229 @@
...
+
+use super::{ConnectionState, JsonRpcResponse};
+
+const MONITOR_RETRY_INTERVAL: Duration = Duration::from_secs(2);
+
+/// Creates a new websocket connection with given url
+pub fn new_connection(url: &str) -> Result<(Reader<TcpStream>, Writer<TcpStream>)> {
+    ClientBuilder::new(url)
+        .chain(|| (ErrorKind::InvalidInput, format!("Malformed url: {}", url)))?
+        .connect_insecure()
```
tomtau: perhaps "TODO" for https?

client-common/src/tendermint/websocket_rpc_client/websocket_rpc_loop.rs
```diff
@@ -0,0 +1,229 @@
...
+///
+/// # How it works
+///
+/// - Connects to websocket server at given `url` and splits the connection in `reader` and `writer`.
+/// - Spawns a thread and runs `websocket_rpc_loop` in the thread which continues until the thread panics.
+/// - For each websocket message received:
+///   - Parse the message into JSON-RPC response.
+///   - Pop the response channel from `channel_map` corresponding to response's `request_id`.
+///   - Send the response to the channel.
+#[allow(clippy::type_complexity)]
```
tomtau: what flags the type complexity? the channel map?

devashishdxt: Nothing. I added this by earlier and modified signature after that. Removed.

client-common/src/tendermint/websocket_rpc_client.rs
tomtau: what happens? does it crash?
I think the JSON-RPC handler is the same for both transports

devashishdxt: I returns error saying that it is unable to parse request json. Anyway, it does not matter on websockets because the amount of data transferred in batch request vs multiple requests in exactly same (as there is no overhead of doing a handshake for each request).

calvinaco:
One small question, not largely related to this feature. Does this websocket lib support disabling ssl verification?

tomtau:
@calvinaco there's not TLS: https://github.com/crypto-com/chain/pull/402#discussion_r326540619

calvinaco:
> @calvinaco there's not TLS: [#402 (comment)](https://github.com/crypto-com/chain/pull/402#discussion_r326540619)

Oh, I thought we will be using this lib for TLS in the future too.

tomtau:
> > @calvinaco there's not TLS: [#402 (comment)](https://github.com/crypto-com/chain/pull/402#discussion_r326540619)
> 
> Oh, I thought we will be using this lib for TLS in the future too.

if possible, we will use it with TLS

tomtau:
@calvinaco I tried (in the web editor) merge and resolve conflicts: https://github.com/crypto-com/chain/pull/402/commits/f6d763165cf1bce248451191c94fbebe63a9fc14

Could you check if it's correct? I may have made a mistake, as the integration test now fails: https://travis-ci.org/crypto-com/chain/jobs/588280792#L4760

calvinaco:
> @calvinaco I tried (in the web editor) merge and resolve conflicts: [f6d7631](https://github.com/crypto-com/chain/commit/f6d763165cf1bce248451191c94fbebe63a9fc14)
> 
> Could you check if it's correct? I may have made a mistake, as the integration test now fails: https://travis-ci.org/crypto-com/chain/jobs/588280792#L4760

It looks ok to me the merge, I am double-checking this PR and see which part could possibly causes the test failed.

calvinaco:
Actually the integration test of the merging commit passed:
https://travis-ci.org/crypto-com/chain/builds/588264893?utm_source=github_status&utm_medium=notification

I am restarting the test and see if the problem persist, then I will follow up base on that.

calvinaco:
My guess is the integration tests do not have enough setup time in CI, and also insufficient wait time after a transaction is broadcasted.

https://github.com/crypto-com/chain/pull/413 tries to fix the above stated issues. Please try to apply it after it is merged.

I have run a simplified one at my own repository with this branch and all the tests can pass.

PR #413, opened 2019-09-23
Problem: (CRO-448) Integration tests suite in CI is unstable

calvinaco:
Solution:
- Lengthened the setup time for docker
- Allow more time for transaction to broadcast and be included in a block
- Use debug build to speed up build process
- Integration tests local mode - Build chain-tx-enclave at prepare phase, instead of start-up phase
- Added ClientCLI in integration tests local mode

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -108,7 +108,7 @@ jobs:
         - . ./env.sh
         - docker-compose up -d || travis_terminate 1;
         - |
-          sleep 20
+          sleep 30
           docker-compose ps
           docker-compose logs -t --tail="all"
         - ./run-test.sh || travis_terminate 1;
--- a/.travis.yml
+++ b/.travis.yml
@@ -108,7 +108,7 @@ jobs:
         - . ./env.sh
         - docker-compose up -d || travis_terminate 1;
         - |
-          sleep 10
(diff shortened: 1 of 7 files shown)


leejw51crypto:
how about adding sleep before syncing
sleep(60*1000); 60 seconds?
too long?

tomtau:
> how about adding sleep before syncing
> sleep(60*1000); 60 seconds?
> too long?

could it instead check the state of what it's waiting for?

leejw51crypto:
yes, polling can be used to check to proceed to next step

calvinaco:
> > how about adding sleep before syncing
> > sleep(60*1000); 60 seconds?
> > too long?
> 
> could it instead check the state of what it's waiting for?

Yes, there are two types of timeout we need now:
1. Wait for docker-compose to setup
2. Wait for transaction to broadcast and be included

The former I can try to query each of the services and see if setup is completed.
The latter requires some existing ClientRPC call to return the transaction hash and I can loop the Tendermint query for the tx status until positive result.

Effort-wise for former I think timeout is sufficient. But the latter is a good improvement to have. I think we can have this fix using timeout to improve the stability first? Then I can work on a more proper check in the test.

PR #420, opened 2019-09-25
[CRO-375] Problem: No integration test for auto-sync

lezzokafka:
Solution: Adding integration test for auto-sync.

### Scratch of this test: 

Similar to ["wallet-transaction.test.ts"](https://github.com/crypto-com/chain/blob/master/integration-tests/client-rpc/test/wallet-transaction.test.ts).

- Generate a new receiver wallet 
- Manual sync once at the beginning, it is to prevent potential timeout caused by the racing condition.
- Get info about Default wallet and receiver wallet  (`senderWalletTransactionListBeforeSend`,  `receiverWalletTransactionListBeforeReceive` etc.)
- Send funds from Default wallet to the receiver wallet
- **Enable the auto-sync**
- Wait for the expected sync to be completed
- Check if the info of the wallets are as expected


leejw51crypto:
good job

lezzokafka:
Humm... I didnt touch that part if the test tho, will look into it.

PR #421, opened 2019-09-25
Problem: autosync don't work correctly when status reply is not delivered

leejw51crypto:
Solution: change wallet when sync is 100% , also add timeout for get status

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -68,11 +68,13 @@ jobs:
   include:
     - <<: *rust
       rust: stable
+      if: (branch != staging.tmp) AND (branch != trying.tmp)
     - <<: *rust
       rust: beta
       if: (branch = staging) OR (branch = trying)
     - <<: *rust
       rust: nightly
+      if: (branch != staging.tmp) AND (branch != trying.tmp)
     - name: Integration Test
       language: rust
       sudo: required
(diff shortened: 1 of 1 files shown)


client-core/src/synchronizer/auto_sync_core.rs
```diff
@@ -447,7 +449,15 @@ where
...
+
+        if self.get_current_height() < self.max_height {
+            self.send_request_block()
+        } else {
+            log::info!("polling_get_blocks, sync is complete, go to next wallet");
+            self.change_to_wait();
+            self.change_wallet();
+            Ok(())
+        }
     }
```
leejw51crypto: feat by @lezzokafka

calvinaco:
Can you explain a bit more about this PR?

leejw51crypto:
yes, in autosync,
we found that in some situation
in GetBlocks state,
wallet doesn't change.

so i added in polling, if sync is complete,
i activated changing wallet.

leejw51crypto:
sure, i'll try to unit test finite states.

tomtau:
bors try

tomtau:
@calvinaco any insight why this one failed: https://travis-ci.org/crypto-com/chain/jobs/589845794#L5849 ?

calvinaco:
The integration tests already passed. So it is `docker-compose logs -t --tail="all"` causing problem. My suspicious is the logs is too big but I need to retry it one more time to verify first.

I will wait for bors to do its job first.

PR #435, opened 2019-09-30
Problem: (CRO-461) No initial network parameters for jailing configuration

devashishdxt:
Solution: Added network parameters for `jail_duration` and `max_allowed_faulty_blocks` in `genesis.json`

### DIFF
--- a/chain-core/src/init/config.rs
+++ b/chain-core/src/init/config.rs
@@ -17,14 +17,18 @@ use std::collections::{BTreeMap, HashSet};
 #[derive(Debug, PartialEq, Eq, Clone)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub struct InitNetworkParameters {
-    // Initial fee setting
-    // -- TODO: perhaps change to be against T: FeeAlgorithm
-    // TBD here, the intention would be to "freeze" the genesis config, so not sure generic field is worth it
+    /// Initial fee setting
+    /// -- TODO: perhaps change to be against T: FeeAlgorithm
+    /// TBD here, the intention would be to "freeze" the genesis config, so not sure generic field is worth it
     pub initial_fee_policy: LinearFee,
-    // minimal? council node stake
+    /// minimal? council node stake
     pub required_council_node_stake: Coin,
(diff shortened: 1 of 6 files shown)


chain-core/src/init/config.rs
```diff
@@ -17,14 +17,18 @@ use std::collections::{BTreeMap, HashSet};
...
-    // minimal? council node stake
+    /// minimal? council node stake
     pub required_council_node_stake: Coin,
-    // stake unbonding time (in seconds)
+    /// stake unbonding time (in seconds)
     pub unbonding_period: u32,
+    /// Minimum jailing time for accounts with faulty validations (in seconds)
+    pub jail_duration: u32,
+    /// Maximum number of faulty blocks allowed for an account before it gets jailed
+    pub max_allowed_faulty_blocks: u16,
```
tomtau: `faulty` is a bit ambiguous -- does it refer to byzantine behaviour or liveness?

devashishdxt: `faulty` is for `faulty_validations` as mentioned in [docs](https://crypto-com.github.io/getting-started/staking.html#slashing).

devashishdxt: Can you suggest a better name for this?

tomtau: `missed_block_block_threshold`?

chain-core/src/init/config.rs
```diff
@@ -17,14 +17,18 @@ use std::collections::{BTreeMap, HashSet};
...
-    // minimal? council node stake
+    /// minimal? council node stake
     pub required_council_node_stake: Coin,
-    // stake unbonding time (in seconds)
+    /// stake unbonding time (in seconds)
     pub unbonding_period: u32,
+    /// Minimum jailing time for accounts with faulty validations (in seconds)
+    pub jail_duration: u32,
+    /// Maximum number of blocks with faulty/missed validations allowed for an account before it gets jailed
+    pub missed_block_threshold: u16,
```
tomtau: It should also have the block signing window length as another argument in which the threshold passes

tomtau:
bors try

PR #438, opened 2019-09-30
Problem:(CRO-468) autosync don't work correctly when status reply is not delivered

leejw51crypto:
Soluton: modify "polling get blocks" to check correctly, and added unit-test


leejw51crypto:
sorry, for confusion.
rebased and sumit again.

PR #439, opened 2019-09-30
Problem: (CRO-461) Staked state does not store jailing information

devashishdxt:
Solution: Added `jailed_until` in `StakedState`

### DIFF
--- a/chain-abci/src/storage/account/mod.rs
+++ b/chain-abci/src/storage/account/mod.rs
@@ -115,6 +115,7 @@ mod test {
                 unbonded: Coin::from(unbonded),
                 unbonded_from,
                 address,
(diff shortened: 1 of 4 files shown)


devashishdxt:
> all chain-tx-validation code that touches staked state (deposit, unbond, withdraw) should check that the account is not jailed

This PR only adds `jailed_until` in `StakedState`. Validation changes will be in following PRs.

tomtau:
> > all chain-tx-validation code that touches staked state (deposit, unbond, withdraw) should check that the account is not jailed
> 
> This PR only adds `jailed_until` in `StakedState`. Validation changes will be in following PRs.

ok, it already has todo notes in some places

PR #443, opened 2019-10-02
Problem: (CRO-461) Can create network transaction when account is jailed

devashishdxt:
Solution: Added a check in transaction verification to check for jailed accounts

### DIFF
--- a/chain-abci/src/storage/tx.rs
+++ b/chain-abci/src/storage/tx.rs
@@ -9,7 +9,9 @@ use chain_core::tx::fee::Fee;
 use chain_core::tx::TransactionId;
 use chain_core::tx::TxAux;
 use chain_core::tx::TxObfuscated;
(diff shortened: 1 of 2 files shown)


devashishdxt:
> it'll be good to add these checks also to `verify_bonded_deposit_core` and `verify_unbonded_withdraw_core` + add tests

@tomtau: Both `verify_bonded_deposit_core` and `verify_unbonded_withdraw_core` are executed in enclave. If we can check if an account is jailed before sending the enclave request, won't it be a lot quicker?

tomtau:
> > it'll be good to add these checks also to `verify_bonded_deposit_core` and `verify_unbonded_withdraw_core` + add tests
> 
> @tomtau: Both `verify_bonded_deposit_core` and `verify_unbonded_withdraw_core` are executed in enclave. If we can check if an account is jailed before sending the enclave request, won't it be a lot quicker?

it's more of a sanity check, enclaves can't trust any input

PR #440, opened 2019-09-30
Problem: (CRO-474) UnbondTx has same TxId across different wallet

calvinlauco:
Solution: Added StakedAddress to UnbondTx

PR #444, opened 2019-10-02
Problem: Integration tests relies on unreliable timeout for state changes

calvinaco:
Solution: Used more reliable mechanism to check for state changes and improved error message

- Transaction: Pull Tendermint for the new transaction
  - Changed staking RPCs to return txid
- Staking state: Pull staking state until expected result
- Added contextual error message to all RPC calls by introducing `asyncMiddleman()`
- Refactored Staking test cases and added unbond txid duplication check


calvinaco:
The added integration tests would need to wait for https://github.com/crypto-com/chain/pull/440 to be merged before working.

calvinaco:
Closed in favor of https://github.com/crypto-com/chain/pull/447

PR #450, opened 2019-10-03
Problem: (CRO-454) Jailed account may get added to validator set

devashishdxt:
Solution: Added a check for setting validator power of jailed accounts to zero.

Note: Code added in this PR is logically unreachable. I'm only adding these to double-check before modifying validator set.

### DIFF
--- a/chain-abci/src/app/mod.rs
+++ b/chain-abci/src/app/mod.rs
@@ -188,20 +188,25 @@ impl<T: EnclaveProxy> abci::Application for ChainNodeApp<T> {
             }
             match maccount {
                 Some(ref account) if self.validator_voting_power.contains_key(&account.address) => {
-                    let min_power = TendermintVotePower::from(
-                        self.last_state
-                            .as_ref()
-                            .expect("delivertx should have app state")
-                            .required_council_node_stake,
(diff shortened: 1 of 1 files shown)


devashishdxt:
> could a test for this be added to `tests/abci_app.rs`?

I don't think so. As I've mentioned in PR's description, this code is unreachable under normal operations.

tomtau:
> > could a test for this be added to `tests/abci_app.rs`?
> 
> @tomtau: I don't think so. As I've mentioned in PR's description, this code is unreachable under normal operations.

if it's unreachable, then `unreachable!("...")` or could be good to at least add a warning log if this happened (to distinguish from setting to zero by withdrawing the stake)

devashishdxt:
> > > could a test for this be added to `tests/abci_app.rs`?
> > 
> > 
> > @tomtau: I don't think so. As I've mentioned in PR's description, this code is unreachable under normal operations.
> 
> if it's unreachable, then `unreachable!("...")` or could be good to at least add a warning log if this happened (to distinguish from setting to zero by withdrawing the stake)

Added `unreachable!()` and a `log::error!()`.

PR #456, opened 2019-10-09
Problem: (CRO-452) Liveness tracker is not changed if a validator is added/removed from validator set

devashishdxt:
Solution: Change liveness tracker in `end_block` based on power change in block.

### DIFF
--- a/chain-abci/src/app/mod.rs
+++ b/chain-abci/src/app/mod.rs
@@ -8,6 +8,7 @@ use log::info;
 
 pub use self::app_init::{ChainNodeApp, ChainNodeState};
 use crate::enclave_bridge::EnclaveProxy;
+use crate::liveness::LivenessTracker;
 use crate::storage::account::AccountStorage;
 use crate::storage::account::AccountWrapper;
 use crate::storage::tx::StarlingFixedKey;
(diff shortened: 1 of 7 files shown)


chain-abci/src/app/mod.rs
```diff
@@ -297,6 +300,32 @@ impl<T: EnclaveProxy> abci::Application for ChainNodeApp<T> {
...
+                        .last_state
+                        .as_mut()
+                        .expect("Last app state not found, init chain was not called");
+
+                    let validator_liveness = &mut last_state.validator_liveness;
+
+                    let validator_address: TendermintValidatorAddress =
+                        into_tendermint_validator_pub_key(&self.validator_pubkeys[&address]).into();
+
+                    if i64::from(*new_power) == 0
```
tomtau: this could be named: 
```
let new_vote_power = TendermintVotePower::from(*new_power);
if new_vote_power == 0 {
...
} else if new_vote_power != 0 && !validator_liveness.contains_key(&validator_address) {
...
}
```
```

tomtau:
bors try

devashishdxt:
> looks ok, but would be good to add a test to cover this

I've added a basic test scenario. More details test cases can be tested in integration tests.

PR #463, opened 2019-10-10
Problem: (CRO-452) Byzantine validators are not jailed

devashishdxt:
Solution: Jail byzantine validators in begin block


chain-abci/src/app/mod.rs
```diff
@@ -148,6 +149,27 @@ impl<T: EnclaveProxy> abci::Application for ChainNodeApp<T> {
...
+                    .validator_liveness
+                    .get(&validator_address)
+                    .expect("Validator not found in liveness tracker")
+                    .address();
+                accounts_to_jail.push(account_address);
+            }
+        }
+
+        for staking_address in accounts_to_jail {
+            self.jail_account(staking_address)
```
tomtau: this could be moved up to L164 instead of building the extra intermediate vector

chain-abci/tests/abci_app.rs
```diff
@@ -788,9 +788,10 @@ fn get_account(account_address: &RedeemAddress, app: &ChainNodeApp<MockClient>)
     let account_key = to_stake_key(&StakedStateAddress::from(*account_address));
     let state = app.last_state.clone().expect("app state");
     println!("committed root hash: {:?}", &state.last_account_root_hash);
-    let items = app
-        .accounts
-        .get(&state.last_account_root_hash, &mut [account_key.clone()]);
+    let items = app.accounts.get(
+        &app.uncommitted_account_root_hash,
+        &mut [account_key.clone()],
+    );
```
tomtau: there should be now `get_one` method -- most tests were written for the earlier before this existed (and weren't updated), so for new code if it only looks up one value, it's preferable instead of wrapping in this `&mut [..]` and unwrapping the result from a map

chain-core/src/state/tendermint.rs
```diff
@@ -40,7 +40,7 @@ impl TendermintValidatorPubKey {
 }
 
 #[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Clone, Encode, Decode)]
-pub struct TendermintValidatorAddress([u8; 20]);
+pub struct TendermintValidatorAddress(pub [u8; 20]);
```
tomtau: is there a need to directly modify the internal value? If it's just getting it, them perhaps `impl Into<[u8; 20]>`?

PR #464, opened 2019-10-10
Probem:(CRO-486) building docker fails in intergration test

leejw51crypto:
Solution: replace ubuntu repo server with something else
restore newline


calvinaco:
What is missing from the Ubuntu repository?

leejw51:
it's geographical proxy issue.

PR #467, opened 2019-10-11
Problem: (CRO-453) Non-live validators are not jailed

devashishdxt:
Solution: After updating liveness of a validator based on `last_commit_info`. Jail if validator is not live.

### DIFF
--- a/chain-abci/src/app/mod.rs
+++ b/chain-abci/src/app/mod.rs
@@ -130,26 +130,26 @@ impl<T: EnclaveProxy> abci::Application for ChainNodeApp<T> {
             ),
         };
 
-        {
-            let last_state = self
+        let last_state = self
             .last_state
             .as_mut()
             .expect("executing begin block, but no app state stored (i.e. no initchain or recovery was executed)");
 
-            last_state.block_time = block_time;
+        last_state.block_time = block_time;
 
-            if block_height > 1 {
-                if let Some(last_commit_info) = req.last_commit_info.as_ref() {
-                    // liveness will always be updated for previous block, i.e., `block_height - 1`
-                    update_validator_liveness(last_state, block_height - 1, last_commit_info);
-                } else {
-                    panic!(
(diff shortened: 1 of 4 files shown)


chain-abci/src/app/mod.rs
```diff
@@ -130,27 +130,27 @@ impl<T: EnclaveProxy> abci::Application for ChainNodeApp<T> {
...
+                update_validator_liveness(last_state, block_height - 1, last_commit_info);
+            } else {
+                panic!(
+                    "No last commit info in begin block request for height: {}",
+                    block_height
+                );
             }
         }
 
+        let mut accounts_to_jail = Vec::new();
```
tomtau: should this be a vector and not a set -- but I guess jailing is idempotent, so it shouldn't matter if the validator was both non-live and byzantine?

chain-abci/src/app/mod.rs
```diff
@@ -164,11 +164,33 @@ impl<T: EnclaveProxy> abci::Application for ChainNodeApp<T> {
...
+            .as_ref()
+            .unwrap()
+            .jailing_config
+            .missed_block_threshold;
+
+        accounts_to_jail.extend(
+            self.last_state
+                .as_ref()
+                .unwrap()
+                .validator_liveness
```
tomtau: I'm thinking that this validator liveness tracking could be some struct wrapping the underlying map + containing the threshold -- it could then keep track of non-live validators after each `update_validator_liveness`... so here, it could directly return the non-live validators (instead of iterating over all validators and checking them)

chain-abci/src/app/mod.rs
devashishdxt: That is correct. Jailing is idempotent. In this case, `Vec` is preferred as compare to `BTreeSet` because it is contiguous memory and requires zero pointer jumps for iteration (better performance while iterating).

PR #468, opened 2019-10-11
Problem:(CRO-483) There is no dedicated ci builder

leejw51:
Solution: add drone.io  config
add prepare.sh

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -1,0 +1,17 @@
+kind: pipeline
+name: default
+
+steps:
+- name: test
+  image: chain
+  pull: never
+  commands:
+  - export RUST_BACKTRACE=1
+  - export RUSTFLAGS=-Ctarget-feature=+aes,+ssse3
+  - export LD_LIBRARY_PATH=$HOME/lib
+  - export LIBRARY_PATH=$HOME/lib
+  - export PATH=$HOME/.cargo/bin:$HOME/.local/bin:$PATH
+  - export PKG_CONFIG_PATH=$HOME/lib/pkgconfig
+  - cargo build
+  - cargo test
+  - echo "HOME FOLDER = " $HOME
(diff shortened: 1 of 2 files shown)


leejw51:
it took about 15 minutes on i7

leejw51:
```
```

leejw51:
if we use local docker image using "pull:never"
compile + unit-test total time is
3m 40s
'''
steps:
- name: test
  image: chain
  pull: never
'''

install_zeromq.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+FILE=/home/travis/lib/libzmq.so
```
tomtau: does this work on drone?

leejw51: yes, this script will be used in drone,too.

tomtau: but this assumes there's a "travis" user?

leejw51: yes, also drone.io uses it

leejw51:
drone io is workig now.
will move to azure soon.

.drone.yml
```diff
@@ -0,0 +1,17 @@
+kind: pipeline
+name: default
+
+steps:
+- name: test
+  image: chain
```
tomtau: is there a docker file for this?

.drone.yml
```diff
@@ -0,0 +1,17 @@
...
+  commands:
+  - export RUST_BACKTRACE=1
+  - export RUSTFLAGS=-Ctarget-feature=+aes,+ssse3
+  - export LD_LIBRARY_PATH=$HOME/lib
+  - export LIBRARY_PATH=$HOME/lib
+  - export PATH=$HOME/.cargo/bin:$HOME/.local/bin:$PATH
+  - export PKG_CONFIG_PATH=$HOME/lib/pkgconfig
+  - cargo build
+  - cargo test
+  - echo "HOME FOLDER= " $HOME
```
tomtau: this doesn't call those preparation bash scripts -- are they needed?

.drone.yml
leejw51crypto: yes , i already setup inside azure

PR #471, opened 2019-10-15
Problem: (CRO-482) jsonrpc-core/http-server 14.0 fails to compile

devashishdxt:
Solution: Changed `jsonrpc-core`/`jsonrpc-http-server` version to `14.0`


leejw51:
i'll working on hderive issue.

PR #478, opened 2019-10-17
Problem: (CRO-280) chain-abci init panics with raw genesis hash rather than hex-encoded

linfeng-crypto:
Solution:
1. encode the raw hash data into string 
2. if the raw hash is None, use "None"
3. make the log into a single line
4. use "{}" instead of "{:?}" in string format


tomtau:
@linfeng-crypto looks ok, just squash the commit

PR #484, opened 2019-10-18
Problem: (CRO-455) No initial network parameters for slashing configuration

devashishdxt:
Solution: Added network parameters for `slashing_config` in `genesis.json`

### DIFF
--- a/chain-core/src/init/config.rs
+++ b/chain-core/src/init/config.rs
@@ -1,8 +1,9 @@
 use parity_scale_codec::{Decode, Encode};
 #[cfg(feature = "serde")]
-use serde::{Deserialize, Serialize};
+use serde::{de, Deserialize, Deserializer, Serialize, Serializer};
 use std::fmt;
 use std::prelude::v1::{String, Vec};
+use std::str::FromStr;
 
 use crate::common::Timespec;
 use crate::init::address::RedeemAddress;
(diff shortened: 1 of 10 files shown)


chain-core/src/init/config.rs
```diff
@@ -42,6 +44,18 @@ pub struct JailingParameters {
...
 }
 
+#[derive(Debug, PartialEq, Eq, Clone, Copy, Encode, Decode)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub struct SlashingParameters {
+    /// Percentage of funds (bonded + unbonded) slashed when validator is not live (liveness is calculated by jailing
+    /// parameters)
+    pub liveness_slash_percent: u8,
+    /// Percentage of funds (bonded + unbonded) slashed when validator makes a byzantine fault
+    pub byzantine_slash_percent: u8,
```
tomtau: As one will need to do fixed-decimal calculations with these, wouldn't make sense to have a dedicated datatype directly -- e.g. wrap `Milli`: `pub struct MaxSlashRate(Milli);` where constructor / decoder of MaxSlashRate would check the value is in the range [0.0000, 1.0000) ?

chain-core/src/init/config.rs
```diff
@@ -42,6 +44,18 @@ pub struct JailingParameters {
...
+    /// Number of blocks to wait before slashing funds from an account
+    pub slash_wait_blocks: u32,
```
tomtau: `unbonding_period` is in unix time -- perhaps make it consistent and check the value <= `unbonding_period`: https://github.com/cosmos/cosmos-sdk/blob/master/x/slashing/spec/07_tombstone.md
ultimately the processing / slash period tracking-execution could be similar https://github.com/cosmos/cosmos-sdk/blob/master/x/slashing/spec/04_begin_block.md except that there's no delegation

PR #496, opened 2019-10-21
Problem: (CRO-500) client-cli fails to compile with cli-table 0.2

devashishdxt:
Solution: Updated `cli-table` version to 0.2 and made necessary code changes


devashishdxt:
bors retry

PR #497, opened 2019-10-21
Problem: (CRO-456) Non-live validators are not slashed

devashishdxt:
Solution: Add non-live validators to slashing queue

### DIFF
--- a/chain-abci/src/app/mod.rs
+++ b/chain-abci/src/app/mod.rs
@@ -151,6 +151,9 @@ impl<T: EnclaveProxy> abci::Application for ChainNodeApp<T> {
             }
         }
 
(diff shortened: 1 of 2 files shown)


chain-abci/src/app/mod.rs
```diff
@@ -151,6 +151,9 @@ impl<T: EnclaveProxy> abci::Application for ChainNodeApp<T> {
             }
         }
 
+        let slashing_time =
+            last_state.block_time + i64::from(last_state.slashing_config.slash_wait_period);
```
leejw51crypto: what about adding more condition here ?
in case, adding time value is bogus

PR #503, opened 2019-10-22
Problem: (CRO-495) no way for client to unjail account

leejw51crypto:
add unjail-tx to default network client

fix un-covered match

fix un-covered match in wallet-rpc

fix argument

add unit test

rename

### DIFF
--- a/client-core/src/cipher/default.rs
+++ b/client-core/src/cipher/default.rs
@@ -161,9 +161,7 @@ impl TransactionObfuscation for DefaultTransactionObfuscation {
             SignedTransaction::DepositStakeTransaction(tx, witness) => {
                 TxQueryInitRequest::Encrypt(Box::new(EncryptionRequest::DepositStake(tx, witness)))
             }
-            SignedTransaction::UnbondStakeTransaction(tx, witness) => {
-                return Ok(TxAux::UnbondStakeTx(tx, witness));
-            }
+
             SignedTransaction::WithdrawUnbondedStakeTransaction(tx, state, witness) => {
                 TxQueryInitRequest::Encrypt(Box::new(EncryptionRequest::WithdrawStake(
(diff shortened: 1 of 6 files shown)


leejw51:
ok

leejw51:
ok, i will add it

client-network/src/network_ops/default_network_ops_client.rs
```diff
@@ -221,6 +221,53 @@ where
...
+                    "Not able to find private key for given address in current wallet",
+                )
+            })?;
+
+        let signature = private_key
+            .sign(transaction.id())
+            .map(StakedStateOpWitness::new)?;
+
+        let signed_transaction = SignedTransaction::UnjailTransaction(transaction, signature);
+        let tx_aux = self.transaction_cipher.encrypt(signed_transaction)?;
```
tomtau: no need to call this, one can construct the TxAux directly

leejw51: removed

client-core/src/cipher/default.rs
```diff
@@ -169,6 +169,9 @@ impl TransactionObfuscation for DefaultTransactionObfuscation {
                     tx, state, witness,
                 )))
             }
+            SignedTransaction::UnjailTransaction(_, _) => {
+                unreachable!("unjailing does not need enclave")
```
tomtau: same for unbondtx -- currently, it's being returned as `return Ok(TxAux::UnbondStakeTx(tx, witness));` -- @devashishdxt may comment what's preferred there (where unreachable panic or identity function?)

leejw51: because SignedTransaction::UnbondStakeTransaction, SignedTransaction::UnjailTransaction are not used, i'll remove from enum.

PR #506, opened 2019-10-23
Problem: (CRO-503) Slashing rate is not proportional to number of failing validators

devashishdxt:
Solution: Made slashing rates = `(((3 * number of failing validators) / total vaildators) ^ 2) * slash rate`

### DIFF
--- a/chain-core/src/init/config.rs
+++ b/chain-core/src/init/config.rs
@@ -2,6 +2,7 @@ use parity_scale_codec::{Decode, Encode};
 #[cfg(feature = "serde")]
 use serde::{de, Deserialize, Deserializer, Serialize, Serializer};
 use std::fmt;
+use std::ops::Mul;
 use std::prelude::v1::{String, Vec};
 use std::str::FromStr;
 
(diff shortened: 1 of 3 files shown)


chain-core/src/init/config.rs
```diff
@@ -72,12 +73,32 @@ impl SlashRatio {
         }
     }
 
+    pub fn get_proportional(
+        self,
+        failing_validators: usize,
+        total_validators: usize,
+    ) -> SlashRatio {
+        let millis =
+            (((3.0 * failing_validators as f64) / total_validators as f64).powi(2) * 1000.0) as u64;
```
leejw51: 3.0 can be constant if we need to adjust

chain-core/src/init/config.rs
```diff
@@ -72,12 +73,32 @@ impl SlashRatio {
...
         self.0.as_millis()
     }
 }
 
+impl Mul for SlashRatio {
+    type Output = Self;
+
+    #[inline]
+    fn mul(self, rhs: SlashRatio) -> Self::Output {
+        SlashRatio::new(self.0 * rhs.0)
```
leejw51: can it be assertion?

chain-core/src/init/config.rs
```diff
@@ -72,12 +73,32 @@ impl SlashRatio {
...
     }
 
+    pub fn get_proportional(
+        self,
+        failing_validators: usize,
+        total_validators: usize,
+    ) -> SlashRatio {
+        let millis =
+            (((3.0 * failing_validators as f64) / total_validators as f64).powi(2) * 1000.0) as u64;
+        SlashRatio::new(Milli::from_millis(std::cmp::min(millis, 1000))).unwrap() * self
```
tomtau: this should use / preserve fixed-point operations rather than casting to floating points, otherwise it may not be deterministic (e.g. different rustc/llvm versions may optimizing floating point operations differently).
`x.powi(2)` can be rewritten as `x*x`

chain-core/src/init/config.rs
devashishdxt: I think having `expect` is similar to having runtime assertion.And in this case, we cannot have a compile time assertion.

chain-core/src/init/config.rs
tomtau: 3 comes from the fact that it can tolerate up to 1/3 of faulting nodes

chain-core/src/init/config.rs
```diff
@@ -72,12 +73,32 @@ impl SlashRatio {
         }
     }
 
+    pub fn get_proportional(
+        self,
+        failing_validators: usize,
+        total_validators: usize,
```
tomtau: I'm thinking whether this makes sense against the number of nodes if the voting power (determined by the bonded stake of validators) isn't equal among them.
Perhaps failing_validators_vote_power / total_vote_power?

devashishdxt:
> I guess this PR is currently blocked by: #507

That is correct. I'm closing this pull request for now and will reopen it once all the changes are done.

PR #529, opened 2019-10-28
Problem: (CRO-517) chain-abci doesn't send app_hash if there are no transactions

leejw51crypto:
Solution:  send app-hash even though there is no txs


leejw51crypto:
i will repost again

PR #530, opened 2019-10-28
Problem:(CRO-517) chain-abci doesn't send app_hash if there are no transactions

leejw51crypto:
Solution:  abci send apphash for empty blocks

### DIFF
--- a/chain-abci/src/app/commit.rs
+++ b/chain-abci/src/app/commit.rs
@@ -37,108 +37,114 @@ pub fn update_utxos_commit(
 }
 
 impl<T: EnclaveProxy> ChainNodeApp<T> {
+    pub fn process_txs(&mut self, inittx: &mut DBTransaction) {
+        for txaux in self.delivered_txs.iter() {
+            let txid: TxId = txaux.tx_id();
+            match &txaux {
+                TxAux::EnclaveTx(TxEnclaveAux::TransferTx {
+                    inputs,
+                    no_of_outputs,
+                    payload: TxObfuscated { txpayload, .. },
+                    ..
+                }) => {
+                    // FIXME: temporary hack / this shouldn't be here
+                    let plain_tx = PlainTxAux::decode(&mut txpayload.as_slice());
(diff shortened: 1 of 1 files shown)


leejw51crypto:
thanks @tomtau

chain-abci/src/app/commit.rs
```diff
@@ -43,7 +43,7 @@ impl<T: EnclaveProxy> ChainNodeApp<T> {
         let mut new_state = orig_state.expect("executing block commit, but no app state stored (i.e. no initchain or recovery was executed)");
         let mut resp = ResponseCommit::new();
         let mut inittx = self.storage.db.transaction();
-        if !self.delivered_txs.is_empty() {
+        {
```
calvinaco: Does it need a block scope?

chain-abci/src/app/commit.rs
```diff
@@ -43,7 +43,7 @@ impl<T: EnclaveProxy> ChainNodeApp<T> {
         let mut new_state = orig_state.expect("executing block commit, but no app state stored (i.e. no initchain or recovery was executed)");
         let mut resp = ResponseCommit::new();
         let mut inittx = self.storage.db.transaction();
-        if !self.delivered_txs.is_empty() {
```
tomtau: I'd suggest to keep this `!self.delivered_txs.is_empty()` check, but to move out the steps  on L123 to L142

tomtau:
bors try

chain-abci/src/app/commit.rs
tomtau: or remove the block scope as @calvinaco suggested https://github.com/crypto-com/chain/pull/530#discussion_r339448866

chain-abci/src/app/commit.rs
leejw51crypto: no, i'll remove it

chain-abci/src/app/commit.rs
leejw51crypto: ok, i'll modify

leejw51crypto:
tidy up

PR #532, opened 2019-10-29
Problem: (CRO-518) HD and Basic wallet code is not separated out

devashishdxt:
Solution: Separated HD wallet code in `HdKeyService`


client-cli/src/command/wallet_command.rs
```diff
@@ -1,11 +1,14 @@
-use client_common::{Error, ErrorKind, Result};
-use client_core::WalletClient;
-use quest::{ask, success};
+use bip39::{Language, Mnemonic};
```
tomtau: should this functionality be here, as it'll be needed in client-rpc + if the client-* is exposed as a library?

devashishdxt: I don’t think HD wallet code can be moved to `client-common` because it works with other parts in `client-core`. One possible change is to return `SecUtf8` instead of `Mnemonic` when creating a new wallet so that we can remove `bip39` dependency from `client-cli` and `client-rpc`.

leejw51: changing SecUtf8 makes sense

leejw51:
bors is pending

PR #534, opened 2019-10-29
 Problem:(CRO-514) No robust tests for Milli arithmetic operations

lezzokafka:
Solution: 
Adding testes with quickcheck 

 @tomtau  I added one more check for **a+b = 0 iff a=b=0**  -Line 343

### DIFF
--- a/chain-core/src/tx/fee/mod.rs
+++ b/chain-core/src/tx/fee/mod.rs
@@ -245,24 +245,27 @@ impl FeeAlgorithm for LinearFee {
 #[cfg(test)]
 mod test {
     use super::*;
+    use quickcheck::quickcheck;
 
-    fn test_milli_add_eq(v1: u64, v2: u64) {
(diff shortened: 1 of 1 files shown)


chain-core/src/tx/fee/mod.rs
```diff
@@ -311,4 +315,88 @@ mod test {
...
+      fn quickcheck_add_zero(n1: u64, n2: u64) -> bool {
+          //Check: Given a+b=0, it holds only when a=b=0
+           let v1 = Milli::from_millis(n1);
+           let v2 = Milli::from_millis(n2);
+           if v1+v2 == Milli::new(0,0)  {
+                v1 == Milli::new(0,0) && v2 == Milli::new(0,0)
+           }  else {
+               true
+           }
+        }
```
tomtau: there's only one case that matches that

PR #533, opened 2019-10-29
Problem: (CRO-511) Dockerfile and integration tests are inefficient

calvinaco:
This PR includes several improvements to speed up integration test setup and execution:
- Reuse built docker image to generate genesis and wallet
  - Before the PR `chain` code has to be built twice
- Generalize `chain` docker image and use pre-built Tendermint from DockerHub
- Optimize `chain` and `chain-tx-enclave` image to reduce layers to be re-built on code changes
- Use whitelisting of `.dockerignore` to reduce number of docker image to be re-built on code changes

PR #536, opened 2019-10-29
Problem: (CRO-501) Integration tests is not running in HW SGX

calvinlauco:
Solution: Add integration tests in Drone CI running in HW SGX

---
- Right now integration tests will run as SW mode in TravisCI and HW mode in Drone
- Remarks: This PR checked-out from https://github.com/crypto-com/chain/pull/533 and has to wait for it to merge first


leejw51:
drone.io is blocked

tomtau:
I approved -- when the pipeline changes, it won't be executed automatically

PR #537, opened 2019-10-30
Problem: (CRO-294) Client can generate invalid transactions

devashishdxt:
Solution: Added balance checks when creating network ops transactions

### DIFF
--- a/client-core/src/wallet/default_wallet_client.rs
+++ b/client-core/src/wallet/default_wallet_client.rs
@@ -395,6 +395,19 @@ where
         ))
     }
 
+    fn has_unspent_transactions(
+        &self,
+        name: &str,
+        passphrase: &SecUtf8,
+        inputs: &[TxoPointer],
+    ) -> Result<bool> {
+        // Check if wallet exists
+        self.wallet_service.view_key(name, passphrase)?;
+
+        self.wallet_state_service
+            .has_unspent_transactions(name, passphrase, inputs)
+    }
+
     #[inline]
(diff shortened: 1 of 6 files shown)


devashishdxt:
> looks ok, but is it only for staked state operations?
> e.g. if one sends TX that spends some tx inputs... if the client doesn't sync, those inputs would still be "unspent" in the internal state (rather than "potentially spent" with some TXID to check whether it was really committed), is that correct?

That is correct. Client won't know if some inputs are spent until it syncs the block in which they were spent.

tomtau:
> > looks ok, but is it only for staked state operations?
> > e.g. if one sends TX that spends some tx inputs... if the client doesn't sync, those inputs would still be "unspent" in the internal state (rather than "potentially spent" with some TXID to check whether it was really committed), is that correct?
> 
> That is correct. Client won't know if some inputs are spent until it syncs the block in which they were spent.

ok, I created CRO-524 for that

client-core/src/wallet/default_wallet_client.rs
```diff
@@ -395,6 +395,19 @@ where
...
     }
 
+    fn has_unspent_transactions(
+        &self,
+        name: &str,
+        passphrase: &SecUtf8,
+        inputs: &[TxoPointer],
+    ) -> Result<bool> {
+        // Check if wallet exists
+        self.wallet_service.view_key(name, passphrase)?;
```
calvinaco: A suggestion: I notice we are using `view_key` to check for wallet existence or passphrase correctness in wallet client, maybe we add a `wallet_service.has_wallet()` ?

devashishdxt:
bors retry

tomtau:
bors retry

PR #543, opened 2019-10-31
Problem: (CRO-530) Staked State does not show jailing information in client-cli

devashishdxt:
Solution: Added `Jailed Until` in cli display table of `staked-state`


tomtau:
bors retry

tomtau:
bors retry

PR #550, opened 2019-11-04
Problem: (CRO-533) Drone CI cannot run tests in parallel and tear down properly

calvinlauco:
Solution: Use branch name as docker-compose prefix and teardown properly

---

- Integration tests use branch name as prefix and random ports
- Always try to bring down docker on teardown
- Always try to kill `./tx-validation-app` and `./tx-query-app` on teardown


calvinaco:
bors try

calvinlauco:
> I think it won't still be possible to run tx-query exec pipelines in parallel due to: https://github.com/crypto-com/chain/blob/master/chain-tx-enclave/tx-query/app/src/test/mod.rs#L96
> 
> Perhaps dockerize those HW test exec pipelines and reuse the stuff in integration test?

Oh. In the integration tests case I assign a random free port on every test run. Can we change this part of code to use an environment variable/argument and assign a new one on each run
?

calvinlauco:
Oh yes maybe dockerize it and let it run in its own environment. Let me check the code first and see which is easier.

tomtau:
> > I think it won't still be possible to run tx-query exec pipelines in parallel due to: https://github.com/crypto-com/chain/blob/master/chain-tx-enclave/tx-query/app/src/test/mod.rs#L96
> > Perhaps dockerize those HW test exec pipelines and reuse the stuff in integration test?
> 
> Oh. In the integration tests case I assign a random free port on every test run. Can we change this part of code to use an environment variable/argument and assign a new one on each run
> ?

yeah, it could be changed

leejw51crypto:
consider using pskill, it's very simple

leejw51crypto:
i suggest docker on real enclave.

calvinaco:
bors try

calvinaco:
HI @tomtau Can you help approve the drone pipeline yml, thank you

calvinaco:
bors retry

calvinaco:
bors retry

PR #551, opened 2019-11-04
Problem:(CRO-536) backwards-incompatible hd-wallet derivation

leejw51crypto:
fix unit-test

### DIFF
--- a/client-core/src/service/hd_key_service.rs
+++ b/client-core/src/service/hd_key_service.rs
@@ -55,7 +55,9 @@ where
             ));
         }
 
-        let seed = Seed::new(mnemonic, passphrase.unsecure());
+        // TODO: advanced/optional recovery" seeding option
+        // give salt as another argument, make default as ""
+        let seed = Seed::new(mnemonic, "");
 
         let hd_key = HdKey {
             seed: seed.as_bytes().to_vec(),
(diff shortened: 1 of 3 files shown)


leejw51crypto:
by this PR,
wallet address is the same with previous.
let's add mnemonic salt as an another argument

client-core/src/service/hd_key_service.rs
```diff
@@ -55,7 +55,7 @@ where
             ));
         }
 
-        let seed = Seed::new(mnemonic, passphrase.unsecure());
```
leejw51crypto: because password can change, additional salt would be easier

client-core/src/service/hd_key_service.rs
```diff
@@ -131,7 +131,9 @@ where
         };
         let coin_type = get_bip44_coin_type();
 
-        let chain_path = ChainPath::from(format!("m/44'/{}'/{}'/0/{}", coin_type, account, index));
+        let chain_path_string = format!("m/44'/{}'/{}'/0/{}", coin_type, account, index);
```
leejw51crypto: for debugging info, i added chain path

client-core/src/wallet/default_wallet_client.rs
```diff
@@ -134,7 +134,7 @@ where
 
         let (view_key, _) =
             self.hd_key_service
-                .generate_keypair(name, passphrase, AddressType::Staking)?;
```
leejw51crypto: both are fine, chose Transfer for backwards compatible

client-core/src/service/hd_key_service.rs
```diff
@@ -131,7 +131,9 @@ where
...
+        log::info!("chain_path {}", chain_path_string);
```
tomtau: `debug!`?

client-core/src/service/hd_key_service.rs
```diff
@@ -55,7 +55,7 @@ where
             ));
         }
 
-        let seed = Seed::new(mnemonic, passphrase.unsecure());
+        let seed = Seed::new(mnemonic, "");
```
tomtau: you can also leave a TODO note for the "advanced/optional recovery" seeding option

client-core/src/service/hd_key_service.rs
leejw51crypto: ok

client-core/src/service/hd_key_service.rs
leejw51crypto: ok

PR #552, opened 2019-11-04
Problem: (CRO-494) not possible to create unjail transaction in client-cli

leejw51:
Solution: call unjail in cli

### DIFF
--- a/client-cli/src/command/transaction_command.rs
+++ b/client-cli/src/command/transaction_command.rs
@@ -1,11 +1,5 @@
 use std::str::FromStr;
 
-use hex::decode;
-use quest::{ask, text, yesno};
-use secstr::SecUtf8;
-use structopt::StructOpt;
-use unicase::eq_ascii;
-
 use chain_core::common::{Timespec, HASH_SIZE_256};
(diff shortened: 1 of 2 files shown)


leejw51crypto:
confirmed it is working fine with 4 nodes using HW enclave

lezzokafka:
This PR is more for creating an unjail Tx, which can become part of CRO-494; the test workflow can be something like :

- Run a blockchain with multiple validators
- Let one of the validator nodes go offline, let it reaches the `SIGNING_THRESHOLD` within the `BLOCK_SIGNING_WINDOW`;
- Check if the node is jailed ;
- Check unbounding is impossible when the node is jailed
- Check un-jailing is impossible within the `JAIL_DURATION` by sending un-jailing Tx.
At the end of `JAIL_DURATION` , 
- Check if the staking address was slashed by `LIVENESS_SLASH_RATIO`; 
- Check if can be un-jailed by an un-jail Tx
- Check if the node is back to the validator list (given that the staking amount after the slashing=> `COUNCIL_NODE_MIN_STAKE` )

leejw51crypto:
wallet bug

PR #554, opened 2019-11-05
Problem:(CRO 266) Not using dummy values for fee calculation

linfeng-crypto:
solution:
- create a Merkle tree (length is 2) for each selected unpent transactions to get the Merkle tree `Path`
- use Vec<u8> allocated with the same number of bytes as the `signature`
- use the above `Merkle tree path` + `signature` as the witness
- pad the payload to the multiples of 128 bits used in the  mock`TxEnclaveAux::TransferTx`

### DIFF
--- a/client-core/src/signer/dummy_signer.rs
+++ b/client-core/src/signer/dummy_signer.rs
@@ -1,0 +1,48 @@
+use crate::SelectedUnspentTransactions;
+use chain_core::common::MerkleTree;
+use chain_core::common::H264;
+use chain_core::tx::witness::tree::RawPubkey;
+use chain_core::tx::witness::{TxInWitness, TxWitness};
+use client_common::Result;
+use secp256k1::schnorrsig::SchnorrSignature;
+
+/// Default implementation of `Signer`
+#[derive(Debug, Clone)]
+pub struct DummySigner {}
+
+impl DummySigner {
+    /// Creates a mock merkletree
+    fn mock_merkletree(
+        &self,
+        raw_pubkey: RawPubkey,
+        tree_length: usize,
+    ) -> Result<MerkleTree<RawPubkey>> {
+        let tree = vec![raw_pubkey; tree_length];
+        Ok(MerkleTree::new(tree))
+    }
+
+    /// Signs with the mock key pair
+    fn sign_inside(&self, total_pubkeys_len: usize) -> Result<TxInWitness> {
+        let raw_pk = RawPubkey::from([0_u8; 33] as H264);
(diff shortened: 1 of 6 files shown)


client-core/src/transaction_builder/default_transaction_builder.rs
```diff
@@ -46,67 +47,62 @@ where
...
-
+            // use the dummy signer to sign the selected unspent transactions
+            let witness = dummy_signer.sign(total_pubkeys_len, &selected_unspent_txs)?;
+            let plain_payload_len = PlainTxAux::TransferTx(tx.clone(), witness).encode().len();
+            //  pad the payload to the multiples of 128bit(8bit*16)
+            let padded_payload = if plain_payload_len % 16_usize == 0 {
+                vec![0; plain_payload_len]
+            } else {
+                vec![0; plain_payload_len + 16_usize]
+            };
```
leejw51crypto: how about
asserting
assert!(0== padded_payload.len()%16);

tomtau: perhaps just:
```
let padded_payload =  vec![0; plain_payload_len + (plain_payload_len % 16_usize)];
``` ?

File: client-core/src/signer/dummy_signer.rs [hunk 2]
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    /// Signs with the mock key pair
+    fn sign_inside(&self, total_pubkeys_len: usize) -> Result<TxInWitness> {
+        // TODO: move to static
+        let secp = Secp256k1::new();
+        let sk = SecretKey::from_slice(&[0xcc; 32])
+            .expect("32 bytes for secret key, within curve order");
+        let pk = PublicKey::from_secret_key(&secp, &sk);
+        let raw_pk = RawPubkey::from(pk.serialize());
```
tomtau: this wouldn't work?
```
let raw_pk = RawPubkey::default();
```

client-core/src/transaction_builder/default_transaction_builder.rs
tomtau: why asserting?

tomtau: can be simplified with: https://github.com/crypto-com/chain/pull/554#discussion_r342399830

tomtau: it shouldn't be `+ 16_usize`, but `+` value between 1 and 15

tomtau:
bors try

tomtau:
bors retry

PR #555, opened 2019-11-05
Problem: (CRO-538) Caching is too huge and timeout exceeded in Travis

calvinlauco:
Solution: Disable caching in nightly and increase test allowed timeout


calvinaco:
bors try

PR #556, opened 2019-11-05
Problem: (CRO-494) not possible to create unjail transaction in client-cli

leejw51:
Solution: call unjail in cli
increase coverage

fix sync

fix save public, private key

### DIFF
--- a/client-cli/src/command/transaction_command.rs
+++ b/client-cli/src/command/transaction_command.rs
@@ -1,11 +1,5 @@
 use std::str::FromStr;
 
-use hex::decode;
-use quest::{ask, text, yesno};
-use secstr::SecUtf8;
-use structopt::StructOpt;
-use unicase::eq_ascii;
-
 use chain_core::common::{Timespec, HASH_SIZE_256};
 use chain_core::init::network::get_network_id;
 use chain_core::state::account::{StakedStateAddress, StakedStateOpAttributes};
@@ -18,6 +12,11 @@ use chain_core::tx::TxAux;
 use client_common::{Error, ErrorKind, PublicKey, Result, ResultExt};
 use client_core::WalletClient;
 use client_network::NetworkOpsClient;
+use hex::decode;
+use quest::{ask, text, yesno};
+use secstr::SecUtf8;
+use structopt::StructOpt;
+use unicase::eq_ascii;
 
 use crate::{ask_passphrase, coin_from_str};
 
@@ -27,6 +26,7 @@ pub enum TransactionType {
     Deposit,
     Unbond,
     Withdraw,
+    Unjail,
 }
 
 impl FromStr for TransactionType {
@@ -41,6 +41,8 @@ impl FromStr for TransactionType {
             Ok(TransactionType::Unbond)
         } else if eq_ascii(s, "withdraw") {
             Ok(TransactionType::Withdraw)
+        } else if eq_ascii(s, "unjail") {
+            Ok(TransactionType::Unjail)
         } else {
             Err(ErrorKind::DeserializationError.into())
         }
@@ -88,6 +90,7 @@ fn new_transaction<T: WalletClient, N: NetworkOpsClient>(
         TransactionType::Withdraw => {
             new_withdraw_transaction(wallet_client, network_ops_client, name, &passphrase)
         }
+        TransactionType::Unjail => new_unjail_transaction(network_ops_client, name, &passphrase),
     }?;
 
     wallet_client.broadcast_transaction(&transaction)?;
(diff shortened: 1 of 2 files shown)


leejw51crypto:
fixed viewkey

client-cli/src/main.rs
```diff
@@ -1,7 +1,7 @@
 #![deny(missing_docs, unsafe_code, unstable_features)]
 //! CLI for interacting with Crypto.com Chain
 mod command;
-
+#[allow(unused_imports)]
```
tomtau: prefer not to have this

client-cli/src/main.rs
```diff
@@ -60,14 +60,19 @@ pub(crate) fn chain_id() -> Option<String> {
...
     password()
         .map(Into::into)
         .chain(|| (ErrorKind::IoError, "Unable to read password"))
 }
 
+#[cfg(test)]
+pub(crate) fn ask_passphrase(_message: Option<&str>) -> Result<SecUtf8> {
+    Ok(SecUtf8::from("123456"))
+}
+
```
tomtau: I'm not sure if this helps with testing, as it'll skip the actual code to be tested and have a replacement -- most things in client are tested in isolation.
For cli testing, it's a bit tricky, but ideally it'd be done with passing in custom streams instead of using stdin/stdout

client-cli/src/command/transaction_command.rs
```diff
@@ -323,6 +338,14 @@ fn ask_staking_address() -> Result<StakedStateAddress> {
...
 }
 
+#[cfg(test)]
+fn ask_staking_address() -> Result<StakedStateAddress> {
+    Ok(
+        StakedStateAddress::from_str("0x83fe11feb0887183eb62c30994bdd9e303497e3d")
+            .expect("get staked address"),
+    )
+}
+
```
tomtau: again not sure if this helps much with testing -- left a more detailed comment below

client-core/src/wallet/default_wallet_client.rs
```diff
@@ -132,11 +135,14 @@ where
...
-        let (view_key, _) =
+        let (public_key, private_key) =
             self.hd_key_service
                 .generate_keypair(name, passphrase, AddressType::Transfer)?;
 
-        self.wallet_service.create(name, passphrase, view_key)
+        self.key_service
+            .add_keypair(&private_key, &public_key, passphrase)?;
+
+        self.wallet_service.create(name, passphrase, public_key)
```
tomtau: this appears to solve a different problem than stated in this PR -- I guess it's "Problem: syncing doesn't work with HD wallets", please open a separate PR for it

client-cli/src/command/transaction_command.rs
```diff
@@ -337,3 +360,133 @@ fn ask_transfer_address() -> Result<ExtendedAddr> {
...
+            MockTransactionCipher,
+        );
+
+        assert!(!new_transaction(
+            &wallet_client,
+            &network_ops_client,
+            name,
+            &TransactionType::Unjail,
+        )
+        .is_ok());
```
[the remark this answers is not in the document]
calvinaco: Forget about the format, actually it is normal

client-cli/src/command/transaction_command.rs
leejw51crypto: it's for coverage failure.
i'll find other way

client-core/src/wallet/default_wallet_client.rs
leejw51crypto: i'll make separate PR

client-cli/src/command/transaction_command.rs
```diff
@@ -337,3 +360,133 @@ fn ask_transfer_address() -> Result<ExtendedAddr> {
...
+            match transaction {
+                SignedTransaction::TransferTransaction(_, _) => unreachable!(),
+                SignedTransaction::DepositStakeTransaction(_, _) => unreachable!(),
+                SignedTransaction::WithdrawUnbondedStakeTransaction(_, _, _) => unreachable!(),
+            }
+        }
+    }
+
+    #[test]
+    fn check_unjail_tx() {
```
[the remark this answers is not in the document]
leejw51crypto: ok

client-cli/src/main.rs
leejw51crypto: ok

client-cli/src/main.rs
leejw51crypto: could you make entering value via stream as separate issue?
so we can unit-test for those functions.

leejw51crypto:
i'll fix

client-cli/src/command/transaction_command.rs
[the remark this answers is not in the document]
leejw51crypto: because it's for coverage failure, i'll remove this for now.
will test via stream.

client-cli/src/command/transaction_command.rs
tomtau: coverage is just for reference -- no need to "hack" it in this way; better to do proper tests

PR #562, opened 2019-11-07
Problem:(CRO-540) syncing does not work in HD wallet

leejw51crypto:
Solution: save public, private key after generating keypair

### DIFF
--- a/client-core/src/wallet/default_wallet_client.rs
+++ b/client-core/src/wallet/default_wallet_client.rs
@@ -117,12 +117,15 @@ where
                 self.hd_key_service
                     .add_mnemonic(name, &mnemonic, passphrase)?;
 
-                let (view_key, _) =
+                let (public_key, private_key) =
                     self.hd_key_service
                         .generate_keypair(name, passphrase, AddressType::Staking)?;
 
+                self.key_service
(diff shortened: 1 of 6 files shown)


leejw51crypto:
> can you add a test (e.g. extend the integration test) to demonstrate it works now?
yes, i'll start coding integration test.

client-core/src/wallet/default_wallet_client.rs
```diff
@@ -132,11 +135,14 @@ where
...
-        let (view_key, _) =
+        let (public_key, private_key) =
             self.hd_key_service
                 .generate_keypair(name, passphrase, AddressType::Transfer)?;
 
-        self.wallet_service.create(name, passphrase, view_key)
+        self.key_service
+            .add_keypair(&private_key, &public_key, passphrase)?;
+
+        self.wallet_service.create(name, passphrase, public_key)
```
calvinaco: Since these two parts of code share much in common, it maybe good to consider extracting them to a function.

leejw51crypto: ok
0 of transfer address is used for view-key

leejw51crypto:
i'm writing intergration test for this

leejw51crypto:
i'll implement integration test

PR #563, opened 2019-11-07
Problem: (CRO-405) Querying public information by probablistic filter

calvinlauco:
Solution: Query informatin using event in deliver_tx

--- 

Before this PR staking account is put into bloom filter in `end_block` event. The syncrhonizer will use the filter to probablistically check if there is staking related transactions inside.

But actually the staking account info. is public and stored inside the `deliver_tx` events.
So there is not need to add the staking address to bloom filter anymore; And should use the `deliver_tx` event instead

### DIFF
--- a/chain-tx-filter/src/lib.rs
+++ b/chain-tx-filter/src/lib.rs
@@ -7,10 +7,9 @@
 #[cfg(all(feature = "mesalock_sgx", not(target_env = "sgx")))]
 extern crate sgx_tstd as std;
 mod filter;
-use chain_core::state::account::StakedStateAddress;
+use chain_core::common::TendermintEventKey;
 use filter::Bloom;
 use filter::H2048;
-use parity_scale_codec::Encode;
 use secp256k1::key::PublicKey;
 use std::convert::TryFrom;
 use std::prelude::v1::Vec;
@@ -42,20 +41,9 @@ impl BlockFilter {
         self.bloom.set(&view_key.serialize()[..]);
     }
 
-    /// adds a staked state address to the filter
-    /// FIXME: to be deprecated/removed -- just use events in ABCI and regular Tendermint indexing
-    pub fn add_staked_state_address(&mut self, address: &StakedStateAddress) {
-        self.modified = true;
-        self.bloom.set(&address.encode());
-    }
-
-    /// gets a Key-Value payload for tendermint events (if any view keys were added)
-    pub fn get_tendermint_kv(&self) -> Option<(Vec<u8>, Vec<u8>)> {
-        if self.modified {
-            Some((Vec::from(&b"ethbloom"[..]), self.bloom.data()))
-        } else {
-            None
-        }
+    /// gets a Key-Value payload for tendermint events
+    pub fn get_tendermint_kv(&self) -> (Vec<u8>, Vec<u8>) {
+        (TendermintEventKey::EthBloom.into(), self.bloom.data())
     }
 
     /// tests if a view key is in the filter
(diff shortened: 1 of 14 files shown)


calvinaco:
> you can also remove this field: https://github.com/crypto-com/chain/blob/master/chain-abci/src/app/app_init.rs#L86 from abci state -- no need to maintain that filter inside ABCI app;
> it'd just need to return the filter obtained from the enclave: https://github.com/crypto-com/chain/blob/master/chain-abci/src/app/mod.rs#L375

Oh yes, I will remove it.

calvinlauco:
> you can also remove this field: https://github.com/crypto-com/chain/blob/master/chain-abci/src/app/app_init.rs#L86 from abci state -- no need to maintain that filter inside ABCI app;
> it'd just need to return the filter obtained from the enclave: https://github.com/crypto-com/chain/blob/master/chain-abci/src/app/mod.rs#L375

Ok, I will remove it.

calvinlauco:
> you can also remove this field: https://github.com/crypto-com/chain/blob/master/chain-abci/src/app/app_init.rs#L86 from abci state -- no need to maintain that filter inside ABCI app;
> it'd just need to return the filter obtained from the enclave: https://github.com/crypto-com/chain/blob/master/chain-abci/src/app/mod.rs#L375

I have deleted `filter` from `ChainNodeApp`.

I removed `StakedStateAddress` related functions but left `add_filter` in `BlockFilter`, though it is not used anywhere because it seems to me a basic function of a block filter.

calvinaco:
bors try

chain-tx-filter/src/lib.rs
```diff
@@ -20,42 +19,27 @@ use std::prelude::v1::Vec;
 pub struct BlockFilter {
     // may be replaced with GCS, e.g. https://github.com/dac-gmbh/golomb-set
     bloom: Bloom,
-    modified: bool,
```
tomtau: I think modified could still be useful -- e.g. if there were no transactions processed in the enclave (there were only transactions with public information processed in abci), then it'll make sense not to attach the filter.
Perhaps -- https://github.com/crypto-com/chain/blob/master/enclave-protocol/src/lib.rs#L170 could return directly `Result<Option<(Vec<u8>, Vec<u8>)>, ()>`

Here: https://github.com/crypto-com/chain/blob/master/chain-tx-enclave/tx-validation/enclave/src/validate.rs#L31
instead of `get_raw()`, it'd call `get_tendermint_kv()`

Here: 
https://github.com/crypto-com/chain/pull/563/files#diff-bc2131504081939018fc5af67df2cc02R373

it could then use the response directly, instead of creating the filter and calling  `get_tendermint_kv()`

calvinaco:
bors try

calvinlauco:
Instead of return `(Vec<u8>, Vec<u8>)`, I decided to return an `Option<Box<[u8; 256]>>` instead. Because this reduces the size of transmission and more importantly easier to calculate the size for the response buffer. (Although I can sum the length of key `ethbloom` in base64, but this is less trivial)

calvinaco:
bors try

tomtau:
bors cancel

tomtau:
@calvinaco @calvinlauco looks good, could you squash the commits ?

tomtau:
@calvinaco @calvinlauco also, there are some compilation errors when the tx-validation is compiled with `SGX_TEST=1`, so the drone pipeline fails to execute

calvinlauco:
> @calvinaco @calvinlauco also, there are some compilation errors when the tx-validation is compiled with `SGX_TEST=1`, so the drone pipeline fails to execute

Ok, I will fix the test and squash the commits.

calvinaco:
bors try

tomtau:
bors try

tomtau:
@calvinaco @calvinlauco strange error `mocha not found` in the integration test: https://travis-ci.org/crypto-com/chain/jobs/611260782#L1921

tomtau:
bors retry

calvinaco:
bors retry

calvinaco:
bors retry

tomtau:
bors retry

tomtau:
@calvinaco @calvinlauco i guess it's ready to go -- squash commits

calvinlauco:
> @calvinaco @calvinlauco i guess it's ready to go -- squash commits

Great, I am squashing and rebase from master now.

calvinaco:
bors try

calvinaco:
bors retry

devashishdxt:
Strange. The code is merged, but still the PR is open. How can this happen?

tomtau:
GitHub bug perhaps?

PR #567, opened 2019-11-08
Problem: (CRO-599) kcov installation is not detected in properly in TravisCI

calvinlauco:
Solution: Change cache location such that installation can be detected on restore

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,9 +41,8 @@ rust: &rust
       - $HOME/lib
       - $HOME/include
       - $HOME/bin
-      - /usr/local/bin # kcov
   before_cache:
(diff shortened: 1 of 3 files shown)


calvinaco:
bors try

tomtau:
@calvinaco does it work: https://travis-ci.org/crypto-com/chain/jobs/609136996#L748 ?

calvinlauco:
> @calvinaco does it work: https://travis-ci.org/crypto-com/chain/jobs/609136996#L748 ?

Yes, it is working. Every branch has it own cache and the trying branch is not cached yet. I will retry the task after the current task finishes then it will be ok.

PR #571, opened 2019-11-11
Problem: client code uses a custom tendermint client (CRO-101)

tomtau:
Solution: migrated code to tendermint-rs


client-common/Cargo.toml
```diff
@@ -23,6 +23,7 @@ jsonrpc = { version = "0.11", optional = true }
 serde_json = { version = "1.0", optional = true }
 parity-scale-codec = { features = ["derive"], version = "1.0" }
 websocket = { version = "0.24", default-features = false, features = ["sync"], optional = true }
+tendermint = { git = "https://github.com/yihuang/tendermint-rs.git", rev="95823039071a44dabc24825b777d3c5b95a6665b" }
```
tomtau: @yihuang we can maintain the fork here: https://github.com/crypto-com/tendermint-rs

but we should try to move to the upstream interchain one when possible, so feel free to open PRs on the upstream interchain one with changes in the fork

yihuang: OK, the biggest changes might be the adding of a bunch of ``Default`` instances.

yihuang: Also, there seems no way to use a feature to make the ring dependency conditional, maybe we have to upgrade the secure_connection to newer ring library before we can merge back to them ;(

yihuang: Just created the first PR to upgrade ``ring`` to ``0.16.9``. https://github.com/interchainio/tendermint-rs/pull/60

tomtau:
wait for @devashishdxt to check, but i guess it's ok to go (as an intermediate step before adding the lite client)?

yihuang:
Yeah, I think so.

yihuang:
The ``Client`` trait is basically the same as before, both http and websocket implementation. Only migrated the data types.

tomtau:
> Looks good to me. Just one question: Does this use Json RPC over websockets or HTTP?

@yihuang I think this is still using the current client, it only migrated it to use tendermint-rs types?

tomtau:
bors try

devashishdxt:
> The `Client` trait is basically the same as before, both http and websocket implementation. Only migrated the data types.

Then, I think, we should modify the pull request title.

yihuang:
> > The `Client` trait is basically the same as before, both http and websocket implementation. Only migrated the data types.
> 
> Then, I think, we should modify the pull request title.

I edited the commit message to include this information.

tomtau:
@yihuang seems based on the integration test: https://travis-ci.org/crypto-com/chain/jobs/611272489#L2007
that the client expect `code` field in the response, but that may be optional?

yihuang:
> @yihuang seems based on the integration test: https://travis-ci.org/crypto-com/chain/jobs/611272489#L2007
> that the client expect `code` field in the response, but that may be optional?

``abci_query`` does need to return a code field, is it testing tendermint rpc node?

tomtau:
> > @yihuang seems based on the integration test: https://travis-ci.org/crypto-com/chain/jobs/611272489#L2007
> > that the client expect `code` field in the response, but that may be optional?
> 
> `abci_query` does need to return a code field, is it testing tendermint rpc node?

yeah, it's running Tendermint (0.32)

yihuang:
> > > @yihuang seems based on the integration test: https://travis-ci.org/crypto-com/chain/jobs/611272489#L2007
> > > that the client expect `code` field in the response, but that may be optional?
> > 
> > 
> > `abci_query` does need to return a code field, is it testing tendermint rpc node?
> 
> yeah, it's running Tendermint (0.32)

Except it returns error I guess.

tomtau:
it may be returning an error -- probably better to run locally and see

leejw51crypto:
please resubmit again,
ci builder will check

tomtau:
I can't resolve the conflicts on @yihuang's repo, so closing this PR for now -- feel free to re-open

PR #575, opened 2019-11-12
Problem: (CRO-527) Punishment information is not reflected in staked state

devashishdxt:
Solution: Added punishment information in staked state

### DIFF
--- a/chain-abci/src/slashing.rs
+++ b/chain-abci/src/slashing.rs
@@ -3,6 +3,7 @@ use serde::{Deserialize, Serialize};
 
 use chain_core::common::Timespec;
 use chain_core::init::config::SlashRatio;
+use chain_core::state::account::PunishmentKind;
 
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, Encode, Decode)]
 /// Slashing schedule for a staking account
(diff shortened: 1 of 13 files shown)


chain-abci/src/slashing.rs
```diff
@@ -11,23 +12,31 @@ pub struct SlashingSchedule {
...
     }
 
     /// Updates slash ratio (only if proposed ratio is greater than current)
     #[inline]
-    pub fn update_slash_ratio(&mut self, new_ratio: SlashRatio) {
+    pub fn update_slash_ratio(&mut self, new_ratio: SlashRatio, punishment_kind: PunishmentKind) {
         if new_ratio > self.slash_ratio {
-            self.slash_ratio = new_ratio
+            self.slash_ratio = new_ratio;
+            self.punishment_kind = punishment_kind;
```
tomtau: seems this isn't tested -- perhaps add a test where two infractions happen?

chain-core/src/init/config.rs
```diff
@@ -54,6 +55,9 @@ impl fmt::Display for DistributionError {
             DistributionError::InvalidVotingPower => {
                 write!(f, "Invalid voting power")
             },
+            DistributionError::InvalidPunishmentConfiguration => {
+                write!(f, "Invalid punishment configuration (maybe slash_wait_period < jail_duration)")
```
calvinaco: This message seems to imply that the `slash_wait_period` should not be less than `jail_duration`; I think it should give some hinting words to describe you should make `slash_wait_period < jail_duration`

chain-abci/src/slashing.rs
devashishdxt: Added a test for checking multiple validator faults

chain-core/src/init/config.rs
devashishdxt: Changed

PR #576, opened 2019-11-12
Problem: (CRO-562) tiny-keccak 2.0 fails to compile

devashishdxt:
Solution: Made necessary changes for `tiny-keccak = 2.0`

### DIFF
--- a/chain-core/Cargo.toml
+++ b/chain-core/Cargo.toml
@@ -13,7 +13,7 @@ mesalock_sgx = ["sgx_tstd", "secp256k1zkp/sgx"]
 
 [dependencies]
 digest = { version = "0.8", default-features = false}
-tiny-keccak = { version = "1.5.0", default-features = false, features = ["keccak"] }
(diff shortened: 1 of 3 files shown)


chain-core/Cargo.toml
```diff
@@ -13,7 +13,7 @@ mesalock_sgx = ["sgx_tstd", "secp256k1zkp/sgx"]
 
 [dependencies]
 digest = { version = "0.8", default-features = false}
-tiny-keccak = { version = "1.5.0", default-features = false, features = ["keccak"] }
+tiny-keccak = { version = "2.0", features = ["keccak"] }
```
tomtau: is it `no_std` now? `default-features = false` was there before due to std, if i remember correctly

tomtau:
bors try

chain-core/Cargo.toml
devashishdxt: Yes. It is now `no_std`.

tomtau:
bors cancel

PR #577, opened 2019-11-12
Problem: (CRO-568) client-core depends on unused tokio dependencies

devashishdxt:
Solution: Removed default features and added features that are used in `client-core`

Note: This does not change `Cargo.lock` because other dependencies of `client-core` still depends on features that are removed in this PR.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -50,6 +50,11 @@ dependencies = [
 ]
 
 [[package]]
+name = "anyhow"
+version = "1.0.19"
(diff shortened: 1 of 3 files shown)


devashishdxt:
> perhaps regenerate Cargo.lock ?

Regenerated `Cargo.lock`

tomtau:
which dependency does need all of tokio?

tomtau:
bors retry

PR #578, opened 2019-11-12
Problem:(CRO-574)  integration test not compiled on linux

leejw51crypto:
Solution: add fixes for linux
fix env error


leejw51crypto:
thanks @calvinaco @tomtau

leejw51crypto:
found a glitch,
will post again

PR #579, opened 2019-11-12
Problem:(CRO-574)  integration test cannot run on linux

leejw51crypto:
Solution: add fixes for linux
fix env error

add sudo

### DIFF
--- a/integration-tests/env.sh
+++ b/integration-tests/env.sh
@@ -28,8 +28,11 @@ function check_command_exist() {
 
 # Awlays execute at script located directory
 CWD=$(pwd)
-if [ x"$(basename "${0}")" = "xenv.sh" ]; then
-    cd "$(dirname "${0}")"
+# in non interactive shell
+# ${0} is -bash
+# TODO: replace this line, to make it work on non interactive shell
+if [ x"$(basename -- "${0}")" = "xenv.sh" ]; then
+    cd "$(dirname -- "${0}")"
 fi
 # Travis CI run `. ./env.sh` using `build.sh`. i.e. `${0}`` is `/home/travis/build.sh`
 
--- a/integration-tests/prepare.sh
+++ b/integration-tests/prepare.sh
@@ -5,6 +5,9 @@ IFS=
 # Global function return value
 RET_VALUE=0
 
+# change attribute
+chmod ug+s ./docker-data
+
 # @argument message
 function print_message() {
     echo "[$(date +"%Y-%m-%d|%T")] ${1}"
--- a/chain-tx-enclave/tx-validation/make.sh
+++ b/chain-tx-enclave/tx-validation/make.sh
@@ -3,4 +3,5 @@ set -e
 
 source /root/.docker_bashrc
 cd ./chain-tx-enclave/tx-validation
+make clean
 make


leejw51crypto:
you can activate integration test on linux , mac
like this

```
export SGX_MODE=sw
export NETWORK_ID = ab
export USE_DOCKER_COMPOSE= 1


cd integration-tests
./prepare.sh
. ./env.sh
docker-compose up

cd integration-tests
cd client-rpc
yarn
yarn test
```

leejw51crypto:
thanks @calvinaco  @tomtau

leejw51crypto:
getting scripts' absolute path is necessary,
in non interactive shell,
it's -bash  <- $0

integration-tests/env.sh
```diff
@@ -28,8 +28,13 @@ function check_command_exist() {
 
 # Awlays execute at script located directory
 CWD=$(pwd)
-if [ x"$(basename "${0}")" = "xenv.sh" ]; then
-    cd "$(dirname "${0}")"
+# in non interactive shell
+# ${0} is -bash
+# TODO: replace this line, to make it work on non interactive shell
+if [ -f "${0}" ]; then
```
leejw51: in non-interactive shell,
${0} is -bash
, basename throws an error

leejw51: this fix is not perfect, but it will make it compile

integration-tests/prepare.sh
```diff
@@ -78,7 +78,7 @@ function index_all_tags() {
 function clone_tendermint_config() {
     print_step "Cloning Tendermint config from \"${1}\" to \"${2}\""
 
-    rm -rf "${2}"
+    sudo rm -rf "${2}"
```
leejw51: in travis, it is no problem, because it's root.
so i added sudo for these commands.

integration-tests/env.sh
tomtau: what about replacing the lines 35 and 36 with:
```
    if [ x"$(basename -- "${0}")" = "xenv.sh" ]; then
        cd "$(dirname -- "${0}")"
    fi
```
?

integration-tests/prepare.sh
tomtau: why is sudo required?

integration-tests/env.sh
leejw51crypto: i will check

integration-tests/prepare.sh
leejw51crypto: because in running,
./prepare.sh
it makes root owned file inside docker,
so external shell is normal user,
it cannot continue

integration-tests/env.sh
leejw51crypto: it works fine!!
thanks

integration-tests/prepare.sh
tomtau: i didn't execute as root and it worked fine

tomtau: @calvinaco @calvinlauco is sudo really required?

leejw51crypto:
ok, i will find a way to bypass

integration-tests/prepare.sh
leejw51crypto: which folder did you run the script?
for me, i ran inside 
cd integrations-tests
./prepare.sh

calvinaco: doing sudo also seems overkill; maybe after each generated docker-data, try to change the ownership to the current user?

leejw51crypto: agree, i will try different approach

leejw51crypto:
removed sudo,
chmod ug+s docker-data

leejw51crypto:
ok, i will

leejw51crypto:
done!
good to go

PR #583, opened 2019-11-13
Problem: chain-abci panic due to failed serialization (CRO-575)

yihuang:
Solution:
Skip serialization of invalid json field.

### DIFF
--- a/chain-abci/src/app/app_init.rs
+++ b/chain-abci/src/app/app_init.rs
@@ -46,6 +46,7 @@ pub struct ChainNodeState {
     pub network_params: NetworkParameters,
     /// council nodes metadata
     /// FIXME: delete node metadata if voting power == 0 for longer than unbonding time
+    #[serde(skip)]
     pub council_nodes_by_power: BTreeMap<(TendermintVotePower, StakedStateAddress), CouncilNode>,
     /// stores staking account address corresponding to tendermint validator addresses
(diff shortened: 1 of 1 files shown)


chain-abci/src/app/app_init.rs
```diff
@@ -46,6 +46,7 @@ pub struct ChainNodeState {
     pub network_params: NetworkParameters,
     /// council nodes metadata
     /// FIXME: delete node metadata if voting power == 0 for longer than unbonding time
+    #[serde(skip)]
```
devashishdxt: I'm not sure if it is ok to skip serialization/deserialization of a field in `ChainNodeState`. Can't we change the structure of `council_nodes_by_power` so that the key can be serialized as a `String`?

Maybe break it down into multiple `BTreeMap`s? 

@tomtau Any comments?

tomtau: for chain-abci, it doesn't matter, as it stores/restores in binary where it includes this field.
serde was there only for extra dumping this information in json for `Info` requests (and no code was depending on that) -- skipping this field is the simplest solution to the problem.

tomtau: this field is just an index needed for dynamic validator logic -- clients (or whoever sends `Info` requests / needs in json) can easily reconstruct this information from the other fields + `/validators` endpoint/method

PR #584, opened 2019-11-13
Problem: client can't trust validators field in genesis api (CRO-101)

yihuang:
Solution:
Validate validators in ``RequestInitChain``, make sure the consistency.


yihuang:
Fixed the failed tests.

yihuang:
bors retry

chain-core/src/state/tendermint.rs
```diff
@@ -224,4 +224,9 @@ impl TendermintVotePower {
     pub fn zero() -> Self {
         TendermintVotePower(0)
     }
+
+    /// get number value of voting power
+    pub fn value(self) -> i64 {
+        self.0
+    }
```
tomtau: this one is redundant -- `i64::from(x)` or `x.into()` should work

yihuang: Fixed

calvinlauco:
> seems the chain-abci crashes in integration test due to validators not being consistent -- i think it's because `tendermint init` is called multiple times to create "genesis.json", from the first one, it takes the pubkey value used in `app_data`... and it appends it the same app_data (+ copies the validator key). it's ok if the node was the where `tendermint init` was called, but it'd fail with a different node, as the validators field may be different.
> so i guess updating the prepare.sh of integration tests may be necessary -- @calvinaco @calvinlauco is that correct?

prepare script run `tendermint init` only once. It then clones into two sets (zerofee and withfee) and updates the genesis with different fee and other configs. So the validator keys should be kept consistent all the time.

yihuang:
Moved tests in ``chain-abci/src/app/jail_account.rs`` to ``chain-abci/tests/punishment.rs``, because there seems to be no easy way to share code between tests.

calvinlauco:
I think the inconsistent validator is because of the power difference:
```
[
    pub_key {
      type: "ed25519"
      data: "\362JiJ\245\347\332\253\260k\351\357\263\304:\340`\264\251\323\234E\220E\3550\030e\213\250\367z"
    }
    power: 12500000000
    ,
] [
    pub_key {
      type: "ed25519"
      data: "\362JiJ\245\347\332\253\260k\351\357\263\304:\340`\264\251\323\234E\220E\3550\030e\213\250\367z"
    }
    power: 10
    ,
]
```

yihuang:
Integration testing ``prepare.sh`` need to update ``genesis.conf`` to set validator voting power to correct value.

calvinlauco:
> Integration testing `prepare.sh` need to update `genesis.conf` to set validator voting power to correct value.

@yihuang 
Yes, `10` is the default value from `tendermint init`. I will create a fix now.

calvinlauco:
> > Integration testing `prepare.sh` need to update `genesis.conf` to set validator voting power to correct value.
> 
> @yihuang
> Yes, `10` is the default value from `tendermint init`. I will create a fix now.

Actually, it may be even better to update in `dev-utils`? Because the initial council node information is processed there. What do you think?

yihuang:
> > > Integration testing `prepare.sh` need to update `genesis.conf` to set validator voting power to correct value.
> > 
> > 
> > @yihuang
> > Yes, `10` is the default value from `tendermint init`. I will create a fix now.
> 
> Actually, it may be even better to update in `dev-utils`? Because the initial council node information is processed there. What do you think?

But currently, dev-utils only output ``app_hash`` and ``app_state``, which get appended to ``genesis.json``.

calvinlauco:
> > > > Integration testing `prepare.sh` need to update `genesis.conf` to set validator voting power to correct value.
> > > 
> > > 
> > > @yihuang
> > > Yes, `10` is the default value from `tendermint init`. I will create a fix now.
> > 
> > 
> > Actually, it may be even better to update in `dev-utils`? Because the initial council node information is processed there. What do you think?
> 
> But currently, dev-utils only output `app_hash` and `app_state`, which get appended to `genesis.json`.

Yes, this requires user to update power in `genesis.json` manually now. Perhaps we should update the setup doc first. @lezzokafka  A better solution should make this more automatic.

But the quick fix I am working on will update the `prepare.sh` first.

yihuang:
> > > > > Integration testing `prepare.sh` need to update `genesis.conf` to set validator voting power to correct value.
> > > > 
> > > > 
> > > > @yihuang
> > > > Yes, `10` is the default value from `tendermint init`. I will create a fix now.
> > > 
> > > 
> > > Actually, it may be even better to update in `dev-utils`? Because the initial council node information is processed there. What do you think?
> > 
> > 
> > But currently, dev-utils only output `app_hash` and `app_state`, which get appended to `genesis.json`.
> 
> Yes, this requires user to update power in `genesis.json` manually now. Perhaps we should update the setup doc first @lezzokafka . A better solution should make this more automatic.
> 
> But the quick fix I am working on will update the `prepare.sh` first.

I just wrote a python script ;D

```
$ cat integration-tests/fix_genesis.py
import sys
import base64
import hashlib
import json

GENESIS = json.load(open(sys.argv[1]))


def get_voting_power(state):
    cointype, coin = state
    assert cointype == 'Bonded'
    return str(int(int(coin) / (10 ** 8)))


def validator_addr(pubkey_base64):
    return hashlib.sha256(base64.b64decode(pubkey_base64)).hexdigest().upper()[:40]


GENESIS['validators'] = [
    {
        'address': validator_addr(node[2]['consensus_pubkey_b64']),
        'pub_key': {
            'type': 'tendermint/PubKeyEd25519',
            'value': node[2]['consensus_pubkey_b64'],
        },
        'power': get_voting_power(GENESIS['app_state']['distribution'][addr]),
    }
    for addr, node in GENESIS['app_state']['council_nodes'].items()
]

json.dump(GENESIS, open(sys.argv[1], 'w'), indent=4)
```

calvinlauco:
> > > > > > Integration testing `prepare.sh` need to update `genesis.conf` to set validator voting power to correct value.
> > > > > 
> > > > > 
> > > > > @yihuang
> > > > > Yes, `10` is the default value from `tendermint init`. I will create a fix now.
> > > > 
> > > > 
> > > > Actually, it may be even better to update in `dev-utils`? Because the initial council node information is processed there. What do you think?
> > > 
> > > 
> > > But currently, dev-utils only output `app_hash` and `app_state`, which get appended to `genesis.json`.
> > 
> > 
> > Yes, this requires user to update power in `genesis.json` manually now. Perhaps we should update the setup doc first @lezzokafka . A better solution should make this more automatic.
> > But the quick fix I am working on will update the `prepare.sh` first.
> 
> I just wrote a python script ;D
> 
> ```
> $ cat integration-tests/fix_genesis.py
> import sys
> import base64
> import hashlib
> import json
> 
> GENESIS = json.load(open(sys.argv[1]))
> 
> 
> def get_voting_power(state):
>     cointype, coin = state
>     assert cointype == 'Bonded'
>     return str(int(int(coin) / (10 ** 8)))
> 
> 
> def validator_addr(pubkey_base64):
>     return hashlib.sha256(base64.b64decode(pubkey_base64)).hexdigest().upper()[:40]
> 
> 
> GENESIS['validators'] = [
>     {
>         'address': validator_addr(node[2]['consensus_pubkey_b64']),
>         'pub_key': {
>             'type': 'tendermint/PubKeyEd25519',
>             'value': node[2]['consensus_pubkey_b64'],
>         },
>         'power': get_voting_power(GENESIS['app_state']['distribution'][addr]),
>     }
>     for addr, node in GENESIS['app_state']['council_nodes'].items()
> ]
> 
> json.dump(GENESIS, open(sys.argv[1], 'w'), indent=4)
> ```

Awesome! Let's use this script first 😜

calvinaco:
bors try

tomtau:
the failed build seems related to this: https://github.com/crypto-com/chain/pull/595

tomtau:
bors retry

PR #588, opened 2019-11-14
Problem: (CRO-392) Outdated dependencies in client's storage encryption

devashishdxt:
Solution: Use `aes-gcm-siv` for encryption/decryption

Note: Turns out that the root cause of timeouts was password hashing using `rust-argon2` which, as far as I know, is only useful when we want to store passwords in a database. Adding cache for wallet details is still a good idea and can be done in a different task.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -8,7 +8,7 @@ steps:
   pull: never
   commands:
   - export RUST_BACKTRACE=1
-  - export RUSTFLAGS=-Ctarget-feature=+aes,+ssse3
+  - export RUSTFLAGS=-Ctarget-feature=+aes,+sse2,+sse4.1,+ssse3
   - export LD_LIBRARY_PATH=$HOME/lib
   - export LIBRARY_PATH=$HOME/lib
   - export PATH=$HOME/.cargo/bin:$HOME/.local/bin:$PATH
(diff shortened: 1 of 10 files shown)


devashishdxt:
bors try

tomtau:
> Note: Turns out that the root cause of timeouts was password hashing using `rust-argon2` which, as far as I know, is only useful when we want to store passwords in a database. Adding cache for wallet details is still a good idea and can be done in a different task.

the client stores structured data (not only random seeds / private keys), so with potentially weak passphrases, it could be useful to slow down brute-forcing.
but that may be a different problem to solve

devashishdxt:
> > Note: Turns out that the root cause of timeouts was password hashing using `rust-argon2` which, as far as I know, is only useful when we want to store passwords in a database. Adding cache for wallet details is still a good idea and can be done in a different task.
> 
> the client stores structured data (not only random seeds / private keys), so with potentially weak passphrases, it could be useful to slow down brute-forcing.
> but that may be a different problem to solve

For this, we can just add a `salt` before hashing. No need to change hashing algorithm.

.drone.yml
```diff
@@ -8,7 +8,7 @@ steps:
   pull: never
   commands:
   - export RUST_BACKTRACE=1
-  - export RUSTFLAGS=-Ctarget-feature=+aes,+ssse3
+  - export RUSTFLAGS=-Ctarget-feature=+aes,+sse2,+sse4.1,+ssse3
```
tomtau: update the pipeline signature

devashishdxt: Updated the signature. Hopefully it'll work now.

PR #593, opened 2019-11-16
Problem: (CRO-582) Create transfer address is not re-usable

calvinlauco:
Solution: Extracted MultiSigAddress from root_hash service to serve single responsibility

---
Remarks:
- This PR is a pre-requisite of the JS library by making components re-usable by JS
- Removed n(total_signers) field because it is implied from len of public keys

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -449,6 +449,7 @@ dependencies = [
  "chain-tx-filter 0.1.0",
  "chrono 0.4.9 (registry+https://github.com/rust-lang/crates.io-index)",
  "hex 0.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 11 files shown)


calvinaco:
bors try

tomtau:
bors retry

tomtau:
@calvinlauco can you squash the commits?

calvinaco:
> looks ok -- note that the multi-sig addresses can have more general conditions, e.g. there are signers A,B,C and it could be signed by either A or B&C... and this client API doesn't allow/expose that, but it could perhaps be added later when required?

Yes, I agree. Flexible signing pairs can have a separate task for it. This task focus more on extracting and refactoring existing logic to be re-usable component.

> @calvinlauco can you squash the commits?

Ok, will squash it now.

calvinaco:
bors retry

PR #599, opened 2019-11-19
Problem:(CRO-545) not possible to unjailTx in client-rpc for test automation

leejw51crypto:
Solution: add unjail api


tomtau:
bors retry

tomtau:
bors retry

PR #601, opened 2019-11-19
Problem(CRO-101): Client doesn't verify responses returned from the full node

yihuang:
Solution:
- Migrate the tendermint data types to `tendermint-rs` library
- Integrate tendermint-rs's lite client implementation to verify "commit/validator set/header".
- Current only implement in manual synchronizer.


client-common/src/tendermint/mock.rs
```diff
@@ -0,0 +1,301 @@
...
+    account, chain, channel, net, node, validator, Moniker, PrivateKey, PublicKey, Version,
+};
+
+const DEFAULT_VALIDATOR_KEY: &str = "{
+  \"type\": \"tendermint/PrivKeyEd25519\",
+  \"value\": \"gJWgIetdLxRc/C2t/XjV65NCqZLTqHS9pU69kBzRmyOKCHDqT/6bKPmKajdBp+KCbYtu9ttTX7+MXrEQOw8Kqg==\"
+}";
+const BLOCK_VERSION: u64 = 10;
+const APP_VERSION: u64 = 0;
+const DEFAULT_GENESIS_JSON: &str = r#"{
```
tomtau: Not sure if it'll make this a bit nicer / more readable -- these long strings for testing could be put into separate files (e.g. `mock/genesis.json`) and added here using a macro (e.g. `include_str!("mock/genesis.json");` )

client-core/src/service/global_state_service.rs
```diff
@@ -98,6 +102,41 @@ where
                     .map(|global_state_optional| global_state_optional.unwrap_or_default())
             })
     }
+
+    /// load trust state
+    pub fn load_trust_state(&self) -> Result<Option<lite::TrustedState>> {
+        // get current trusted state
+        let opt = self.storage.get(KEYSPACE, TRUST_STATE_KEY)?;
+        if let Some(bytes) = opt {
+            let state = serde_json::from_slice(bytes.as_slice()).chain(|| {
```
tomtau: AFAIK the rest of the client stores data as binary using SCALE codec, but that's quite chain-specific and it'd be tiresome to retrofit all the Tendermint data types (as we plan to depend on the mainstream rather than a custom fork), so I think it's ok to store this data in json -- @devashishdxt @leejw51 @calvinlauco any thoughts?

Cargo.lock
```diff
@@ -1844,6 +1924,29 @@ dependencies = [
...
+name = "prost-amino-derive"
+version = "0.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "digest 0.7.6 (registry+https://github.com/rust-lang/crates.io-index)",
+ "failure 0.1.6 (registry+https://github.com/rust-lang/crates.io-index)",
+ "itertools 0.7.11 (registry+https://github.com/rust-lang/crates.io-index)",
+ "proc-macro2 0.4.30 (registry+https://github.com/rust-lang/crates.io-index)",
+ "quote 0.6.13 (registry+https://github.com/rust-lang/crates.io-index)",
+ "sha2 0.7.1 (registry+https://github.com/rust-lang/crates.io-index)",
```
tomtau: does tendermint-rs need `prost-amino-derive` or were these more for tmkms? It's just that this brings a few duplicate crates of older versions (e.g. sha2 is already in the cargo.lock under version 0.8)

client-core/src/service/global_state_service.rs
devashishdxt: We can also use `bincode` which'll be more performant. `json` is also fine.

leejw51crypto: json is fine

calvinaco: Since we are going to depend on the mainstream, I also think using JSON seems better in the long term; When later we have more stress test result, we can have more data to determine whether we need to change this approach.

Cargo.lock
yihuang: Yeah, amino is needed to compute vote's signing bytes, also several merkle hashes is computed based on amino encoded bytes.

client-common/src/tendermint/client.rs
```diff
@@ -25,9 +28,67 @@ pub trait Client: Send + Sync {
         heights: T,
     ) -> Result<Vec<BlockResults>>;
 
+    /// Make `validators` call to tendermint
+    fn validators(&self, _height: u64) -> Result<ValidatorsResponse> {
```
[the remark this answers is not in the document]
yihuang: There are several `MockClient` implementation in the test code, I find it tedious to replicate the panic instances there.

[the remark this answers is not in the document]
devashishdxt: Yes. Maybe you can just return `Ok(ValidatorsResponse::new())` instead of `panic!`?

tomtau:
bors try

client-common/src/tendermint/websocket_rpc_client.rs
```diff
@@ -279,15 +281,16 @@ impl Client for WebsocketRpcClient {
...
     }
 
     #[inline]
     fn block_batch<'a, T: Iterator<Item = &'a u64>>(&self, heights: T) -> Result<Vec<Block>> {
         let params = heights
             .map(|height| ("block", vec![json!(height.to_string())]))
             .collect::<Vec<(&str, Vec<Value>)>>();
-        self.call_batch::<Block>(params)
+        let mut rsps = self.call_batch::<BlockResponse>(params)?;
+        Ok(rsps.drain(..).map(|rsp| rsp.block).collect())
```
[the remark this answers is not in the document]
yihuang: Yeah, `into_iter` is better here.

yihuang:
> looks ok if it works with the integration test -- left a few comments; one main thing would be if the tendermint-rs dependencies could be upgraded. I see in https://github.com/interchainio/tendermint-rs/pulls there are a few dependabot PRs, so I assume they'll be merged eventually

I've sent all the changes as separated PRs there, but the merge process could take some time.

client-common/src/tendermint/mock.rs
yihuang: I was trying to use the `Default` trait here, which would save some code, but arbitrary `Default` implementation doesn't always make sense, and the PR of this idea gets rejected upstream.

Also, I marked ignore to one of the tests of manual synchronizer, because of the problem of test data. Now with fully verified synchronizer, we need to mock full verifiable block data, which is not done so far. Maybe we should create a dedicated crate for this, which only included in dev-dependencies.

client-common/src/error.rs
```diff
@@ -54,7 +54,15 @@ impl Error {
 impl fmt::Display for Error {
     #[inline]
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(f, "{}: {}", self.kind, self.message)
```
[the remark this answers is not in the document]
yihuang: I've changed it to `"{}: {}, origin: {:?}", self.kind, self.message, self.origin`.

client-common/src/tendermint/client.rs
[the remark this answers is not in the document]
yihuang: > Yes. Maybe you can just return `Ok(ValidatorsResponse::new())` instead of `panic!`? I'd ideally prefer not having a default implementation.

Actually, there are several methods are not used so far, I removed them, only keep `block_batch_verified` in the trait, now it's more feasible to remove the default implementation, which is done is new commit.

client-common/src/tendermint/websocket_rpc_client.rs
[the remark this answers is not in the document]
yihuang: Fixed in new commit.

client-common/src/error.rs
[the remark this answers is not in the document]
calvinaco: Is there a particular reason you want to include `self.origin` here? I think it makes sense for error message not to include the origin error unless you are debugging it.

yihuang:
@calvinaco Yeah, it's for debugging.

yihuang:
I think it's already solved, that message is from an older commit.

tomtau:
bors retry

client-common/src/error.rs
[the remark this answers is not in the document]
devashishdxt: > Is there a particular reason you want to include `self.origin` here? I think it makes sense for error message not to include the origin error unless you are debugging it.

This was the original intention of not including `self.origin` in display implementation.

client-common/src/tendermint/client.rs
```diff
@@ -25,9 +25,16 @@ pub trait Client: Send + Sync {
         heights: T,
     ) -> Result<Vec<BlockResults>>;
 
+    /// Fetch continuous blocks and verify them.
+    fn block_batch_verified<'a, T: Clone + Iterator<Item = &'a u64>>(
+        &self,
+        _state: lite::TrustedState,
```
devashishdxt:
```suggestion
        state: lite::TrustedState,
```

client-common/src/tendermint/client.rs
```diff
@@ -25,9 +25,16 @@ pub trait Client: Send + Sync {
...
+        _heights: T,
```
devashishdxt:
```suggestion
        heights: T,
```

client-common/src/tendermint/http_rpc_client.rs
```diff
@@ -115,19 +147,21 @@ impl Client for RpcClient {
...
 
-        response
-            .into_iter()
-            .map(|block| block.chain(|| (ErrorKind::InvalidInput, "Block information not found")))
-            .collect::<Result<Vec<Block>>>()
+        let rsps = self.call_batch::<BlockResponse>(&params)?;
+        rsps.into_iter()
+            .map(|rsp| {
+                rsp.chain(|| (ErrorKind::InvalidInput, "Block information not found"))
+                    .map(|rsp_| rsp_.block)
```
devashishdxt:
```suggestion
                    .map(|rsp| rsp.block)
```

yihuang:
The test failed on travis, because `Time` serialization lose precision, changed the format to use `RFC3339::Nanos`, please retry.

client-core/src/synchronizer/manual_synchronizer.rs
```diff
@@ -457,16 +476,26 @@ mod tests {
...
 
-        fn query(&self, _path: &str, _data: &[u8]) -> Result<QueryResult> {
+        fn query(&self, _path: &str, _data: &[u8]) -> Result<AbciQuery> {
             unreachable!()
         }
     }
 
     #[test]
+    #[ignore]
+    // ignore now, mock block verification failed.
```
tomtau: instead of ignoring the test, could there be a flag (default true) to enable block verification -- in this test, it could be false?
Also, if I run a local full node myself (instead of connecting to someone else's), I may want to disable the block verification

devashishdxt: Add a `TODO` here to fix this test case.

client-common/src/tendermint/client.rs
yihuang: Adding a underscore is to remove the unused variable warning.

client-core/src/synchronizer/manual_synchronizer.rs
yihuang: 👌

client-common/src/tendermint/client.rs
devashishdxt: This is a trait. I don't think you'll get an unused variable warning in traits.

client-common/src/tendermint/http_rpc_client.rs
yihuang: It's for avoiding the same name confusion with outer closure.

devashishdxt: Ok.

client-core/src/synchronizer/manual_synchronizer.rs
yihuang: Done.

PR #622, opened 2019-11-26
Problem:(CRO-586) No way to add a node to validator set from client-rpc

linfeng-crypto:
solution: Add a `staking_validatorNodeJoin` method in rpc.

### DIFF
--- a/client-rpc/src/rpc/staking_rpc.rs
+++ b/client-rpc/src/rpc/staking_rpc.rs
@@ -4,12 +4,15 @@ use jsonrpc_core::Result;
 use jsonrpc_derive::rpc;
 
 use chain_core::init::coin::Coin;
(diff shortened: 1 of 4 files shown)


tomtau:
bors retry

tomtau:
bors retry

client-rpc/src/rpc/staking_rpc.rs
```diff
@@ -47,6 +50,15 @@ pub trait StakingRpc: Send + Sync {
 
     #[rpc(name = "staking_unjail")]
     fn unjail(&self, request: WalletRequest, unjail_address: String) -> Result<String>;
+
+    #[rpc(name = "staking_validatorNodeJoin")]
+    fn node_join(
```
leejw51crypto: how about
"join"

PR #630, opened 2019-11-27
Problem (CRO-614): clippy error of mnemonic.rs

yihuang:
Solution:
- Ignore the `new-without-default` one.
- Fix the unneeded collect call.

### DIFF
--- a/client-core/src/mnemonic.rs
+++ b/client-core/src/mnemonic.rs
@@ -36,6 +36,7 @@ impl<'de> Deserialize<'de> for Mnemonic {
 
 impl Mnemonic {
     /// Generate and returns mnemonic words
+    #[allow(clippy::new_without_default)]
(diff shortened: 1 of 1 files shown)


client-core/src/mnemonic.rs
```diff
@@ -36,6 +36,7 @@ impl<'de> Deserialize<'de> for Mnemonic {
 
 impl Mnemonic {
     /// Generate and returns mnemonic words
+    #[allow(clippy::new_without_default)]
```
devashishdxt: Instead of this, can you create a `Default` implementation which calls `new()`

tomtau: I think the reason to ignore this (perhaps a comment could be added) is that Default should be deterministic, but `new()` generates a random mnemonic

PR #643, opened 2019-11-29
Problem: (CRO-608) Transaction builder and signer are coupled with wallet

calvinlauco:
Solution: Refactor transaction builder and signer to be generic

---
- Keep transaction builder and signer; But rename with wallet prefix
- Extract transaction building logic into `RawTransferTransactionBuilder`
- Refactor `Signer` trait to be generic

### DIFF
--- a/client-core/src/signer.rs
+++ b/client-core/src/signer.rs
@@ -1,27 +1,48 @@
 //! Transaction signing
-mod default_signer;
 mod dummy_signer;
+mod key_pair_signer;
 mod unauthorized_signer;
+mod wallet_signer;
 
-pub use default_signer::DefaultSigner;
 pub use dummy_signer::DummySigner;
+pub use key_pair_signer::KeyPairSigner;
 pub use unauthorized_signer::UnauthorizedSigner;
+pub use wallet_signer::{WalletSigner, WalletSignerManager};
 
-use secstr::SecUtf8;
-
-use chain_core::tx::witness::TxWitness;
+use chain_core::tx::data::address::ExtendedAddr;
+use chain_core::tx::witness::{TxInWitness, TxWitness};
 use client_common::Result;
 
 use crate::SelectedUnspentTransactions;
 
-/// Interface for signing transactions
+/// Interface for signing message and transactions
 pub trait Signer: Send + Sync {
-    /// Signs given transaction with private keys corresponding to selected unspent transactions
-    fn sign<T: AsRef<[u8]>>(
+    /// Signs given transaction with private keys corresponding to selected
+    /// unspent transactions
+    fn schnorr_sign_transaction<T: AsRef<[u8]>>(
         &self,
-        name: &str,
-        passphrase: &SecUtf8,
         message: T,
         selected_unspent_transactions: &SelectedUnspentTransactions<'_>,
     ) -> Result<TxWitness>;
+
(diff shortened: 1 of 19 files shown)


client-core/src/signer/key_pair_signer.rs
```diff
@@ -0,0 +1,81 @@
+//! A signer that sign message using the provided key pair
```
tomtau: could this made `1-of-N` easily (or can leave a todo comment), or rename it to reflect that it's 1-of-1?

client-core/src/signer.rs
```diff
@@ -1,27 +1,39 @@
...
         &self,
-        name: &str,
-        passphrase: &SecUtf8,
         message: T,
         selected_unspent_transactions: &SelectedUnspentTransactions<'_>,
     ) -> Result<TxWitness>;
+
+    /// Returns true if the signer is able to schnorr sign message as the
+    /// provided address, false otherwise
+    fn can_schnorr_sign<'a>(&self, signing_addr: &'a ExtendedAddr) -> Result<bool>;
```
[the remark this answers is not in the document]
calvinaco: Good, I think this enum approach is better.

Yes, the method is to verify if the signer contains necessary keys to sign and combination inside the `signing_addr`. It powers the `RawTransferTransactionBuilder` to sign all signable inputs.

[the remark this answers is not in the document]
calvinaco: However, I think our current wallet storage only store root hash, we will need some mechanism to know what other parties are needed on multi-sig scenario. I will put it as a TODO first.

client-core/src/signer/dummy_signer.rs
```diff
@@ -47,14 +47,13 @@ impl DummySigner {
...
         total_pubkeys_len: usize,
-        selected_unspent_transactions: &SelectedUnspentTransactions<'_>,
+        inputs_len: usize,
     ) -> Result<TxWitness> {
-        selected_unspent_transactions
-            .iter()
+        (0..inputs_len)
             .map(|_| self.sign_tx(total_pubkeys_len))
             .collect::<Result<Vec<TxInWitness>>>()
             .map(Into::into)
```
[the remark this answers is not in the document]
calvinaco: Cool, thanks for the suggestion. I will change it.

client-core/src/signer/key_pair_signer.rs
calvinaco: Yes, I think so. I can let the user register 1-of-n addresses, then we can utilize the information to sign inputs. This will also make `MultiSignUnlock` in `SignCondition` enum possible.

I will now leave it as TODO first and add it when l I add support of MultiSig support to JS lib.

client-core/src/signer.rs
```diff
@@ -1,27 +1,49 @@
...
+    fn schnorr_sign_transaction<T: AsRef<[u8]>>(
         &self,
-        name: &str,
-        passphrase: &SecUtf8,
         message: T,
         selected_unspent_transactions: &SelectedUnspentTransactions<'_>,
     ) -> Result<TxWitness>;
+
+    /// Returns true if the signer is able to schnorr sign message as the
+    /// provided address, false otherwise
```
tomtau: I guess the comment is outdated now

calvinaco: Done

PR #644, opened 2019-11-29
Problem: (CRO-619) Missing script to do final checkings before push

calvinlauco:
Solution: Add script to do final checkings including build, test, clippy and audit

---
- By default, it runs at quiet mode with a minimum logs output.
- Toggle verbose mode by `./before_push.sh --verbose`


tomtau:
bors retry

PR #651, opened 2019-12-04
Problem (CRO-594): block results and "fast sync" are not verified in client

yihuang:
Solution:
- Store history hashes to `COL_APP_HASHS`, history state to `COL_APP_STATES`
- Add `query_state_batch` to rpc client, and verify block results with it.
- Make fast-forward syncing optional
- Improve `BlockGenerator` to pass manual synchronizer test

`BlockGenerator` not support mocking transaction data yet, maybe save it for another PR.

### DIFF
--- a/chain-abci/src/app/app_init.rs
+++ b/chain-abci/src/app/app_init.rs
@@ -3,6 +3,7 @@ use std::convert::TryInto;
 
 use abci::*;
 use enclave_protocol::{EnclaveRequest, EnclaveResponse};
+use integer_encoding::VarInt;
 use kvdb::DBTransaction;
 use log::{info, warn};
 use parity_scale_codec::{Decode, Encode};
@@ -28,7 +29,7 @@ use chain_core::init::config::NetworkParameters;
 use chain_core::state::account::StakedStateDestination;
 use chain_core::state::account::{CouncilNode, StakedState, StakedStateAddress};
 use chain_core::state::tendermint::{BlockHeight, TendermintValidatorAddress, TendermintVotePower};
-use chain_core::state::RewardsPoolState;
+use chain_core::state::{ChainState, RewardsPoolState};
 use chain_core::tx::TxAux;
 
 /// Validator state tracking
@@ -87,27 +88,24 @@ pub struct ChainNodeState {
     pub last_apphash: H256,
     /// time in previous block's header or genesis time
     pub block_time: Timespec,
-    /// root hash of the sparse merkle patricia trie of staking account states
-    pub last_account_root_hash: StarlingFixedKey,
-    /// last rewards pool state
-    pub rewards_pool: RewardsPoolState,
     /// Record how many block each validator proposed, used for rewards distribution,
     /// cleared after rewards distributed
     pub proposer_stats: BTreeMap<StakedStateAddress, u64>,
-    /// network parameters (fee policy, staking configuration etc.)
-    pub network_params: NetworkParameters,
     /// state of validators (keys, voting power, punishments, rewards...)
     #[serde(skip)]
     pub validators: ValidatorState,
     /// genesis time
     pub genesis_time: Timespec,
+
+    /// The parts of states which involved in computing app_hash
+    pub top_level: ChainState,
 }
 
 impl ChainNodeState {
     pub fn genesis(
         genesis_apphash: H256,
         genesis_time: Timespec,
-        last_account_root_hash: StarlingFixedKey,
+        account_root: StarlingFixedKey,
         rewards_pool: RewardsPoolState,
         network_params: NetworkParameters,
         validators: ValidatorState,
@@ -116,12 +114,14 @@ impl ChainNodeState {
             last_block_height: 0,
             last_apphash: genesis_apphash,
             block_time: genesis_time,
-            last_account_root_hash,
-            rewards_pool,
             proposer_stats: BTreeMap::new(),
-            network_params,
             validators,
             genesis_time,
+            top_level: ChainState {
+                account_root,
+                rewards_pool,
+                network_params,
+            },
         }
     }
 }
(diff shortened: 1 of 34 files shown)


tomtau:
bors try

yihuang:
Give up on debugging of the compile error of `chain-core`. so I moved back `ChainNodeState` and related types to chain-abci, and let client-core depends on chain-abci directly.

tomtau:
bors retry

chain-abci/src/app/app_init.rs
```diff
@@ -248,13 +249,24 @@ fn get_voting_power(
...
+fn store_valid_genesis_state(
+    genesis_state: &ChainNodeState,
+    inittx: &mut DBTransaction,
+    write_history_states: bool,
+) {
     let encoded = genesis_state.encode();
     inittx.put(COL_NODE_INFO, LAST_STATE_KEY, &encoded);
-    inittx.put(COL_EXTRA, b"init_chain_state", &encoded);
+    let encoded_height = i64::encode_var_vec(0);
+    inittx.put(COL_APP_HASHS, &encoded_height, &genesis_state.last_apphash);
```
tomtau: I guess this was there before, but not sure if historical app hashes (i..e mapping height->app hash) are needed for validation or only for client querying (as abci query can specify a block height to query against)

leejw51crypto:
if abci and client needs same data, consider separate to another lower level.
abci --->   another lower level 
client -->
so that client not dependent abci

chain-abci/src/app/app_init.rs
```diff
@@ -248,13 +249,24 @@ fn get_voting_power(
...
+    if write_history_states {
+        inittx.put(COL_APP_STATES, &encoded_height, &encoded);
```
leejw51crypto: histroy states implies that vector
how about previos state?

chain-abci/src/app/commit.rs
```diff
@@ -168,12 +168,13 @@ impl<T: EnclaveProxy> ChainNodeApp<T> {
...
-        inittx.put(
-            COL_APP_STATES,
-            &i64::encode_var_vec(new_state.last_block_height),
-            &new_state.last_apphash,
-        );
-        inittx.put(COL_NODE_INFO, LAST_STATE_KEY, &new_state.encode());
+        let encoded_state = new_state.encode();
+        inittx.put(COL_NODE_INFO, LAST_STATE_KEY, &encoded_state);
+        let encoded_height = i64::encode_var_vec(new_state.last_block_height);
+        inittx.put(COL_APP_HASHS, &encoded_height, &new_state.last_apphash);
```
leejw51crypto: COL_APP_HASH?
because it's single

chain-abci/src/app/app_init.rs
yihuang: This is only to keep current behavior, it's only used in the query. Because `COL_APP_STATES` is optional, we can't simply remove the `COL_APP_HASHS`.

chain-abci/src/app/commit.rs
yihuang: It's actually a map of `height -> app_hash`

chain-abci/src/app/app_init.rs
yihuang: It's stored as key value map: `height -> state`

yihuang:
> looks good -- client libraries depending on abci is a bit of an overkill / extra coupling.
> It seems for the lite verification, the client doesn't need the whole node state, but only the top level "app state" / the inputs to computing app hash (staked state merkle trie root hash, rewards pool state, network parameters).
> 
> Perhaps extracting this needed information to a struct, e.g. something like:
> 
> ```
> pub struct ChainState {
>   last_account_root_hash: H256,
>   rewards_pool: RewardsPool,
>   network_params: NetworkParameters,
> }
> ```
> 
> that would be in chain-core (it'll just need feature guards for serde),
> and have this around:
> 
> * ChainNodeState could have a field say `last_top_level_state: ChainState` instead of individual components
> * compute_app_hash could take `&ChainState` as argument instead of those components
> * it could be stored for that historical querying and returned to client
> 
> any thoughts?

I'll try this.

yihuang:
> if abci and client needs same data, consider separate to another lower level.
> abci ---> another lower level
> client -->
> so that client not dependent abci

Yeah, chain-core is a good place for this, it's just chain-core are more strict because of the no_std thing,.
I'll try to bring it back to chain-core again.

yihuang:
- Moved ChainState into chain-core.
- Fixed integration testing.
- Fixed enclave build.

integration-tests/docker-compose.yml
```diff
@@ -48,7 +48,7 @@ services:
       - ${TENDERMINT_ZEROFEE_RPC_PORT:-16657}:26657
   chain-abci-zerofee:
     image: "${CHAIN_DOCKER_IMAGE:-integration-tests-chain}"
-    command: "/usr/bin/chain-abci --host=0.0.0.0 --port=26658 --chain_id=${CHAIN_ID} --genesis_app_hash=${ZEROFEE_APP_HASH} --enclave_server=tcp://chain-tx-enclave-zerofee:25933 --data=/.storage"
+    command: "/usr/bin/chain-abci --host=0.0.0.0 --port=26658 --chain_id=${CHAIN_ID} --genesis_app_hash=${ZEROFEE_APP_HASH} --enclave_server=tcp://chain-tx-enclave-zerofee:25933 --tx_query=tcp://chain-tx-enclave-zerofee:25933 --data=/.storage"
```
tomtau: tx-query address != enclave server

just a note to myself, as this will then need to be fixed in https://github.com/crypto-com/chain/pull/637

tomtau:
bors retry

yihuang:
Rebased.

PR #656, opened 2019-12-05
Problem: (CRO-622) Raw call function of WebsocketRpcClient cannot be accessed from other crates

devashishdxt:
Solution: Made `call` and `call_batch` functions `pub`


devashishdxt:
> lgtm -- @devashishdxt I guess this is needed for the faucet / explorer?

Yes.

PR #657, opened 2019-12-05
Problem (CRO-627) council node pubkey format different from tendermint

yihuang:
Solution:
- serialize TendermintValidatorPubKey the same as tendermint.
- Remove ValidatorPubkey in favor of TendermintValidatorPubKey.

### DIFF
--- a/chain-core/src/state/tendermint.rs
+++ b/chain-core/src/state/tendermint.rs
@@ -1,30 +1,63 @@
 use parity_scale_codec::{Decode, Encode};
 #[cfg(feature = "serde")]
 use serde::{
-    de::{self, Visitor},
+    de::{self, Error as _, Visitor},
     Deserialize, Deserializer, Serialize, Serializer,
 };
 use sha2::{Digest, Sha256};
 use std::convert::TryFrom;
 use std::fmt;
 use std::prelude::v1::{String, ToString, Vec};
+#[cfg(feature = "base64")]
+use thiserror::Error;
 
 /// Tendermint block height
 /// TODO: u64?
 pub type BlockHeight = i64;
+
+/// ed25519 public key size
+pub const PUBLIC_KEY_SIZE: usize = 32;
 
 /// The protobuf structure currently has "String" to denote the type / length
 /// and variable length byte array. In this internal representation,
 /// it's desirable to keep it restricted and compact. (TM should be encoding using the compressed form.)
 #[derive(Debug, PartialEq, Eq, Hash, PartialOrd, Ord, Clone, Encode, Decode)]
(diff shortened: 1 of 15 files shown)


yihuang:
Oh, this one needs to wait for https://github.com/crypto-com/chain/pull/651

chain-core/src/state/tendermint.rs
```diff
@@ -13,20 +13,53 @@ use std::prelude::v1::{String, ToString, Vec};
...
+
+#[cfg(feature = "base64")]
+fn deserialize_ed25519_base64<'de, D>(deserializer: D) -> Result<[u8; PUBLIC_KEY_SIZE], D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let bytes = base64::decode(String::deserialize(deserializer)?.as_bytes())
+        .map_err(|e| D::Error::custom(format!("{}", e)))?;
+    let mut result = [0u8; PUBLIC_KEY_SIZE];
+    result.copy_from_slice(&bytes);
```
tomtau: `copy_from_slice` will panic if the lengths don't match, so ideally before this, there should be a check on `bytes.len() == PUBLIC_KEY_SIZE`

chain-core/src/state/tendermint.rs
```diff
@@ -37,15 +70,26 @@ impl fmt::Display for TendermintValidatorPubKey {
 }
 
 impl TendermintValidatorPubKey {
+    #[cfg(feature = "base64")]
+    pub fn from_base64(bytes: &[u8]) -> Result<TendermintValidatorPubKey, base64::DecodeError> {
+        let mut result = [0u8; PUBLIC_KEY_SIZE];
+        result.copy_from_slice(&base64::decode(bytes)?);
```
tomtau: copy_from_slice will panic if the lengths don't match, so ideally before this, there should be a check on bytes.len() == PUBLIC_KEY_SIZE

chain-core/src/state/tendermint.rs
tomtau: also perhaps just deserialise the string and then call `TendermintValidatorPubKey::from_base64` ?

yihuang: Fixed, also brings in `thiserror` crate, looks like a decent way to handle error type hierarchies.

tomtau:
bors try

yihuang:
Fixed the integration test script error.

tomtau:
bors retry

tomtau:
bors retry

tomtau:
bors retry

tomtau:
bors retry

PR #658, opened 2019-12-05
Problem: monetary expansion calculation could panic for misconfigration

yihuang:
Solution:
- Checked the arithemetic error, when happends print a warning
  and don't mint new coins.

### DIFF
--- a/chain-abci/src/app/rewards.rs
+++ b/chain-abci/src/app/rewards.rs
@@ -1,34 +1,42 @@
-use std::cmp::min;
+use std::cmp::{max, min};
 
-use chain_core::common::fixed::{exp, FixedNumber};
+use chain_core::common::fixed::{exp, FixedNumber, EXP_LOWER_BOUND};
 use chain_core::init::{coin::Coin, params::NetworkParameters};
 use chain_core::state::account::StakedStateAddress;
 use chain_core::state::tendermint::{TendermintValidatorAddress, TendermintVotePower};
-use chain_core::tx::fee::Milli;
 
 use crate::app::{update_account, ChainNodeApp};
 use crate::enclave_bridge::EnclaveProxy;
 use crate::storage::tx::get_account;
 
-fn milli_to_fixed(n: Milli) -> FixedNumber {
-    FixedNumber::from_num(n.as_millis()) / 1000
-}
-
 pub fn monetary_expansion(
-    tau: Milli,
+    tau: u64,
     total_staking: Coin,
     minted: Coin,
     params: &NetworkParameters,
 ) -> Coin {
     let cap = params.get_rewards_monetary_expansion_cap();
-    let r0 = milli_to_fixed(params.get_rewards_monetary_expansion_r0());
-    let tau = milli_to_fixed(tau);
-    let total_staking = FixedNumber::from_num(u64::from(total_staking)) / 1_0000_0000;
-    let amount = total_staking * r0 * exp(-total_staking / tau) * 1_0000_0000;
+    let r0 = FixedNumber::from_num(params.get_rewards_monetary_expansion_r0().as_millis()) / 1000;
+    let tau = FixedNumber::from_num(tau);
+    let total_staking = FixedNumber::from_num(u64::from(total_staking));
+    let amount = total_staking
+        * r0
+        * exp(max(
+            FixedNumber::from_num(EXP_LOWER_BOUND),
+            -total_staking / tau,
+        ));
     min(
         (cap - minted).unwrap_or_default(),
         Coin::new(amount.to_num()).unwrap(),
     )
+}
+
+// rate < 1_000_000, no overflow.
+fn mul_micro(n: u64, rate: u64) -> u64 {
+    assert!(rate <= 1_000_000);
+    let div = n / 1_000_000;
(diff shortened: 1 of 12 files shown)


chain-abci/src/app/rewards.rs
```diff
@@ -10,25 +11,34 @@ use crate::app::{update_account, ChainNodeApp};
...
-    let amount = total_staking * r0 * exp(-total_staking / tau) * 1_0000_0000;
-    min(
+    let r0 = milli_to_fixed(params.get_rewards_monetary_expansion_r0())?;
+    let tau = milli_to_fixed(tau)?;
+    let total_staking =
+        FixedNumber::checked_from_num(u64::from(total_staking))?.checked_div(*FIXED_100M)?;
+    let amount = total_staking
+        .checked_mul(r0)?
+        .checked_mul(exp(-total_staking.checked_div(tau)?)?)?
+        .checked_mul(*FIXED_100M)?;
```
tomtau: is there a reason why `total_staking` needs to be scaled down by decimal places / FIXED_100M (and then scaled up again)?
r0 should be < 1
exp(-....) should be < 1
so that shouldn't overflow.

yihuang: The reason to scale `total_staking` is because of how the formula works, if not scale, the formula or parameters needs to adjust somehow.
The error happens when `tau` is too small, and overflow or division by zero could happen when calculating `exp(-S / tau)`, I'm still trying to figure out the details.

tomtau: @lezzokafka ?

tomtau: The formula should work on base units / carsons / Coin fine ?
the only "strange" result may be if one gets say 0.5 Coin or something like that, in which case just round down?

yihuang: In the fold step of `continued_fraction`: `y / (x + acc)`, when the end result is very small and precision is limited, `(x + acc)` could become zero and cause division by zero error, or the division result becomes very large and cause overflow.
Maybe we only need to do checked calculation in `continued_fraction`, and treat error result as zero.

tomtau: can we put static bounds / compile-time assertions on `x`? And `pub fn exp(x: FixedNumber) -> FixedNumber` instead of `FixedNumber` would take some `BoundFixedNumber` which is a fixed number in some safe range?

tomtau: there could then be bounds on `tau`, so that the most extreme possible values are safe:

* `tau` can't be 0
* `total_staking` can be between `0` (non-sensical value, but static...) and `max_supply`

tomtau: what if minimal valid tau is 1_0000_0000 ?

lezzokafka: The `continued_fraction :y / (x + acc)` is to caculate $e^{x/y}$ type,
Just wondering  would it be easier to do sort of substitution and say calculate $z=x/y$ first then do continued fraction for $ e^z$?🤔🤔  @yihuang

yihuang: > The `continued_fraction :y / (x + acc)` is to calculate $e^{x/y}$ type,
> Just wondering would it be easier to do sort of substitution and say calculate $z=x/y$ first then do continued fraction for $ e^z$?🤔🤔 @yihuang

That's what we are doing now, doing `-S/tau` first, then do the `exp` on the result.

yihuang: > what if minimal valid tau is 1_0000_0000 ?

Assuming total_staking is `50_000_000_000`(half of max_coin), the critical value of tau causing panic is `836253711`, but although `836253712` don't crash, the result of exp is very far from correct answer.

Also, the current formula seems to vary a lot for different `total_staking` values, which seems like a problem @lezzokafka :
```
>>> math.exp(-10000000000.0 / 1450000000) * 10000000000.0
10112665.497004168
>>> math.exp(-50000000000.0 / 1450000000) * 50000000000.0
5.2880824093933016e-05
```

tomtau: > the current formula seems to vary a lot for different total_staking values,

 I think that's expected, as with the huge total stake, the value reward diminishes? @lezzokafka

yihuang: After discussing with @lezzokafka, we can set a lower bound on `-S/tau`. The formula becomes: `S * R0 * exp(max(-20, -S/tau))`

yihuang: > The formula should work on base units / carsons / Coin fine ?
> the only "strange" result may be if one gets say 0.5 Coin or something like that, in which case just round down?

To avoid scaling up and down on `total_staking`, we can scale up `tau` in configuration, `tau` is currently configed as a `Milli`, we can change it into a `u64` or `Coin`.

tomtau: is there still a need for scaling by `1_0000_0000`?

tomtau: and to prevent misconfigurations, InitConfig validation could check the parameters

yihuang: It's done.

tomtau:
bors retry

PR #661, opened 2019-12-09
Problem (CRO-630): fixed 0.5.0 fails to compile

yihuang:
Solution:
Feature guard the fixed module.

### DIFF
--- a/chain-core/Cargo.toml
+++ b/chain-core/Cargo.toml
@@ -7,9 +7,8 @@ readme = "../README.md"
 edition = "2018"
 
 [features]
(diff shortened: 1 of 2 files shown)


tomtau:
oh, the basic drone pipeline compiles the whole thing

tomtau:
It may be better just to wait for SGX SDK 1.10.0 -- the beta version has an updated toolchain: https://github.com/apache/incubator-mesatee-sgx/blob/v1.1.0-beta/dockerfile/Dockerfile.1804.nightly#L27

PR #662, opened 2019-12-09
Problem: (CRO-631) Panic while generating genesis.json from example-dev-conf.json

devashishdxt:
Solution: Fixed error in `example-dev-conf.json`


devashishdxt:
bors retry

devashishdxt:
bors retry

PR #664, opened 2019-12-10
Problem: (CRO-527) Missing validator information in block explorer

devashishdxt:
Solution: Created an `abci_query` endpoint to return list of validators

### DIFF
--- a/chain-abci/src/app/query.rs
+++ b/chain-abci/src/app/query.rs
@@ -4,7 +4,7 @@ use crate::storage::tx::get_account;
 use crate::storage::*;
 use abci::*;
 use chain_core::common::{MerkleTree, Proof as MerkleProof, H256, HASH_SIZE_256};
-use chain_core::state::account::StakedStateAddress;
+use chain_core::state::account::{CouncilNodeMetadata, StakedStateAddress};
 use chain_core::state::ChainState;
 use chain_core::tx::data::{txid_hash, TXID_HASH_ID};
 use integer_encoding::VarInt;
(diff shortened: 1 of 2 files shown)


chain-abci/src/app/query.rs
```diff
@@ -180,6 +179,32 @@ impl<T: EnclaveProxy> ChainNodeApp<T> {
                     }
                 }
             }
+            "validators" => {
```
yihuang: In crypto-com terminology, it's called "council nodes"?

tomtau: The convention is "council node" info =  tendermint validator + custom Chain staking stuff
"validator" info = TM validator info only.
so in this case, I guess it's the former

devashishdxt: Changed.

chain-core/src/state/account.rs
```diff
@@ -203,6 +203,22 @@ impl CouncilNode {
...
+    /// Name of validator
+    pub name: ValidatorName,
+    /// Current voting power of validator
+    pub voting_power: TendermintVotePower,
+    /// Address of staking account of validator
+    pub staking_address: StakedStateAddress,
+    /// Optional security email address of validator
+    pub security_contact: ValidatorSecurityContact,
+    /// Tendermint consensus validator-associated public key
+    pub tendermint_pubkey: TendermintValidatorPubKey,
```
tomtau: the compilation fails because of encoding/decoding in no_std -- could this be simplified to:
```
CouncilNodeMetadata {
pub node: CouncilNode,
    /// Current voting power of validator
      pub voting_power: TendermintVotePower,
      /// Address of staking account of validator
      pub staking_address: StakedStateAddress,
```

devashishdxt: We don't need `Encode`/`Decode` implementations for this. Removed.

devashishdxt:
bors try

tomtau:
bors retry

devashishdxt:
bors retry

PR #669, opened 2019-12-11
Problem: (CRO-583) dev-utils generate incorrect power

calvinlauco:
Solution: Change dev-utils to generate whole genesis with correct power

---

```
cargo run --bin dev-utils
    -- genesis generate
    --genesis_dev_config_path ./dev-utils/example-dev-conf.json
    --tendermint_genesis_path ~/.tendermint/config/genesis.json
    [--in_place]
```

### Options

- in_place: Replace the Tendermint `genesis.json` file inline and backup current file as `genesis.bak.json`


calvinaco:
bors try

chain-core/src/state/tendermint.rs
```diff
@@ -250,6 +250,12 @@ impl From<TendermintVotePower> for u64 {
     }
 }
 
+impl fmt::Display for TendermintVotePower {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", self.0.to_string())
```
devashishdxt: I don't think you have to manually call `.to_string()` here.

```suggestion
        write!(f, "{}", self.0)
```

dev-utils/src/commands/genesis_command.rs
```diff
@@ -23,79 +29,240 @@ pub enum GenesisCommand {
         about = "Generate new genesis.json parameters (app_hash + app_state)"
     )]
     Generate {
+        #[structopt(
+            name = "tendermint_genesis_path",
+            short,
+            long,
+            help = "Path to the Tendermint genesis.json file (e.g. ~/.tendermint/config/genesis.json)"
+        )]
+        tendermint_genesis_path: PathBuf,
```
devashishdxt: Can we make `tendermint_genesis_path` optional and default to `$TMHOME/config/genesis.json` or `$HOME/config/genesis.json` if `$TMHOME` is not defined?

yihuang:
I guess we can remove the `fix-genesis.py` now.

calvinaco:
bors try

> I guess we can remove the `fix-genesis.py` now.

Removed, thanks!

dev-utils/src/commands/genesis_command.rs
```diff
@@ -23,79 +29,240 @@ pub enum GenesisCommand {
...
+            )
+            .map(|_| ()),
         }
     }
+}
 
-    pub fn do_generate(genesis_dev: &GenesisDevConfig) -> Result<(String, InitConfig)> {
-        let mut dist: BTreeMap<RedeemAddress, (StakedStateDestination, Coin)> = BTreeMap::new();
+#[derive(Debug, Clone, Serialize)]
+pub struct Validator {
```
devashishdxt: Instead of creating this `struct` manually. You can use the one in `tendermint-rs`. Currently, its not re-exported from `client-common` but you can do so by adding following line in `client-common/src/tendermint/types.rs`:

```
pub use tendermint::validator::Info as ValidatorInfo;
```

After that, you can do `use client_common::tendermint::types::ValidatorInfo` in `dev-utils`. You may also need to add few more functions via extension traits.

dev-utils/src/commands/genesis_command.rs
```diff
@@ -18,84 +25,268 @@ use crate::commands::genesis_dev_config::GenesisDevConfig;
...
         )]
         genesis_dev_config_path: PathBuf,
+
+        #[structopt(
+            name = "inline",
+            short,
+            long,
+            help = "Replace Tendermint genesis.json file in place, saving backups with .bak.json extension"
+        )]
+        inline: Option<bool>,
```
devashishdxt: No need to have `Option<bool>`. Only `bool` should work.

See this: https://github.com/crypto-com/chain/blob/1e3485c5bf3f9b83565878af314d8858213f88c7/client-cli/src/command.rs#L105

dev-utils/src/commands/genesis_command.rs
```diff
@@ -18,84 +25,268 @@ use crate::commands::genesis_dev_config::GenesisDevConfig;
...
+    let tendermint_genesis_path = match tendermint_genesis_path {
+        Some(path) => path.clone(),
+        None => find_default_tendermint_path().chain(|| {
+            (
+                ErrorKind::InvalidInput,
+                "Unable to find Tendermint folder in $TMHOME or $HOME",
+            )
+        })?,
+    };
+    let inline = match inline {
```
devashishdxt: We won’t have to do this once `inline` is `bool`.

dev-utils/src/commands/genesis_command.rs
```diff
@@ -18,84 +25,268 @@ use crate::commands::genesis_dev_config::GenesisDevConfig;
...
+}
+
+fn find_default_tendermint_path() -> Option<PathBuf> {
+    find_tendermint_path_from_env("TMHOME")
+        .or_else(|| find_tendermint_path_from_env("HOME"))
+}
+
+fn find_tendermint_path_from_env(env_name: &str) -> Option<PathBuf> {
+    if let Ok(home) = env::var(env_name) {
+        let path_buf = PathBuf::from(format!("{}/.tendermint/config/genesis.json", home));
```
devashishdxt: Default `TMHOME` is `$HOME/.tendermint`. You don’t have to add `.tendermint` to path when `TMHOME` is present in environment.

dev-utils/src/commands/genesis_command.rs
calvinlauco: Thanks! I notice `bool` does not need Optional before but I check the `structopt` they mention optional field is denoted by adding `Option`. That's why I add it. I will remove the unnecessary `Option`

dev-utils/src/commands/genesis_command.rs
calvinlauco: Good catch, I will change it.

dev-utils/src/commands/genesis_command.rs
calvinlauco: Great, I am reexporting those structs now and add necessary traits.

calvinaco:
bors try

dev-utils/src/commands/genesis_command.rs
calvinaco: After I tried, I decided to add `TendermintValidator` into `chain-core/src/state/tendermint.rs` instead of re-exporting types from `tendermint-rs`.

The reason is that the command now uses structs implemented in `chain-core/src/state/tendermint.rs`. If I introduce the `tendermint-rs` types, it will be using a hybrid of custom-defined Tendermint types and from `tendermint-rs`.

I will left this as a TODO before we migrate everything to use `tendermint-rs` types.

dev-utils/src/commands/genesis_command.rs
calvinaco: Done

dev-utils/src/commands/genesis_command.rs
calvinaco: Done

tomtau:
the integration test still fails
```
error: Found argument '--tendermint_genesis_path' which wasn't expected, or isn't valid in this context
6048	
6049	USAGE:
6050	    dev-utils genesis generate --genesis_dev_config_path <genesis_dev_config_path>
6051	
6052	For more information try --help
```

calvinaco:
> the integration test still fails
> 
> ```
> error: Found argument '--tendermint_genesis_path' which wasn't expected, or isn't valid in this context
> 6048	
> 6049	USAGE:
> 6050	    dev-utils genesis generate --genesis_dev_config_path <genesis_dev_config_path>
> 6051	
> 6052	For more information try --help
> ```

Yes, I am looking into it.

calvinaco:
bors try

calvinaco:
bors try

calvinaco:
bors try-

calvinaco:
bors try

calvinaco:
bors try-

calvinaco:
bors try

calvinaco:
bors try-

calvinaco:
bors try

calvinaco:
bors try-

calvinaco:
bors try

calvinaco:
bors try

calvinaco:
bors try

calvinaco:
bors try-

calvinaco:
bors try

PR #673, opened 2019-12-12
Problem: (CRO-599) No way to get public key for creating new MultiSig address

devashishdxt:
Solution: Exposed `wallet_newPublicKey` method in `client-rpc`

### DIFF
--- a/client-rpc/src/rpc/wallet_rpc.rs
+++ b/client-rpc/src/rpc/wallet_rpc.rs
@@ -12,8 +12,8 @@ use chain_core::tx::data::output::TxOut;
 use chain_core::tx::TxObfuscated;
 use chain_core::tx::{TxAux, TxEnclaveAux};
 use client_common::{PublicKey, Result as CommonResult};
-use client_core::types::TransactionChange;
 use client_core::types::WalletKind;
+use client_core::types::{AddressType, TransactionChange};
 use client_core::{Mnemonic, MultiSigWalletClient, WalletClient};
 
 use crate::server::{rpc_error_from_string, to_rpc_error, WalletRequest};
@@ -28,6 +28,9 @@ pub trait WalletRpc: Send + Sync {
 
     #[rpc(name = "wallet_restore")]
     fn restore(&self, request: WalletRequest, mnemonics: Mnemonic) -> Result<String>;
+
+    #[rpc(name = "wallet_newMultiSigAddressPublicKey")]
+    fn new_multi_sig_address_public_key(&self, request: WalletRequest) -> Result<String>;
 
     #[rpc(name = "wallet_createStakingAddress")]
     fn create_staking_address(&self, request: WalletRequest) -> Result<String>;
@@ -127,6 +130,17 @@ where
             .new_transfer_address(&request.name, &request.passphrase)
             .map_err(to_rpc_error)?;
         Ok(request.name)
+    }
+
+    fn new_multi_sig_address_public_key(&self, request: WalletRequest) -> Result<String> {
+        self.client
+            .new_public_key(
+                &request.name,
+                &request.passphrase,
+                Some(AddressType::Transfer),
+            )
+            .map(|public_key| public_key.to_string())
+            .map_err(to_rpc_error)
     }
 
     fn create_staking_address(&self, request: WalletRequest) -> Result<String> {


client-core/src/hd_seed.rs
```diff
@@ -39,13 +39,14 @@ impl HDSeed {
...
+        address_type: Option<AddressType>,
         index: u32,
     ) -> Result<(PublicKey, PrivateKey)> {
         let coin_type = get_bip44_coin_type_from_network(network);
         let account = match address_type {
-            AddressType::Transfer => 0,
-            AddressType::Staking => 1,
+            Some(AddressType::Transfer) => 0,
+            Some(AddressType::Staking) => 1,
+            None => 2,
```
tomtau: I don't have enough context for this, but "multisig" address is a transfer address, so why this separation? Just create a transfer address and get an associated public key?

devashishdxt: Transfer addresses in HD wallet are 1-of-n transfer addresses which are created using `wallet_client.new_transfer_address()`.

tomtau:
The issue / problem was "No way to get public key of transfer address for MultiSig address creation",
not "No way to get public key for creating new MultiSig address"...

I assume it could be solved, say, by API that lists public keys or given an address, it'll return the associated public key or something like that -- @calvinaco ?

devashishdxt:
> The issue / problem was "No way to get public key of transfer address for MultiSig address creation",
> not "No way to get public key for creating new MultiSig address"...
> 
> I assume it could be solved, say, by API that lists public keys or given an address, it'll return the associated public key or something like that -- @calvinaco ?

@tomtau Some context about this in comments of this: https://mcoproduct.atlassian.net/browse/CRO-601

calvinlauco:
> The issue / problem was "No way to get public key of transfer address for MultiSig address creation",
> not "No way to get public key for creating new MultiSig address"...
> 
> I assume it could be solved, say, by API that lists public keys or given an address, it'll return the associated public key or something like that -- @calvinaco ?

That was my original intention. But the core of the question is actually to re-enable the ability to create a MultiSig address for a wallet, which is broken before this fix.

I have discussed with @devashishdxt  and I think this is more trivial way to re-enable this feature.

devashishdxt:
> 1. as multisig addresses are transfer addresses, the extra "account" separation is just an accidental complexity - or what does it help with to have a separate index?
> 2. calling the api "new public key" is a bit confusing (it's in the internals, but that could be renamed with it), because all addresses are derived from public keys, so when I come to as a first time, its purpose is unclear -- perhaps something like `new_transfer_address_component`, `new_transfer_address_multi_part`... (as `new_transfer_address` currently assumes 1-of-1) may make it more explicit

1. Removed an extra index.
2. Renamed RPC function to `new_multi_sig_address_public_key`

calvinlauco:
> > The issue / problem was "No way to get public key of transfer address for MultiSig address creation",
> > not "No way to get public key for creating new MultiSig address"...
> > I assume it could be solved, say, by API that lists public keys or given an address, it'll return the associated public key or something like that -- @calvinaco ?
> 
> That was my original intention. But the core of the question is actually to re-enable the ability to create a MultiSig address for a wallet, which is broken/not available before this fix.
> 
> I have discussed with @devashishdxt and I think this is a trivial way to re-enable this feature.

My intention is trying to simply this create MultiSig address process: With a `createPublicKey` call we need only two steps:
1. create public key
2. use it to create MultiSig address
Done

But then I study again how Bitcoin does it and I have the following thoughts:
https://en.bitcoin.it/wiki/Multisignature#Creating_a_Multisignature_Address_with_Bitcoin-Qt
- Creating a transfer address and query public key from it is acceptable
- The transfer address can be kind of act as a key to query back the public key used?

On the other hand, I agree with @tomtau that MultiSig address is a kind of transfer address. So I don't prefer to separate it from the transfer address index in HD wallet.

@devashishdxt any thought?

devashishdxt:
> * Creating a transfer address and query public key from it is acceptable

I don't think we should re-use a public key from an already created transfer address. This may be acceptable but should not be recommended.

> * The transfer address can be kind of act as a key to query back the public key used?

You can easily get your public key from existing multi-sig address (1-of-n/m-of-n). See this:
https://github.com/crypto-com/chain/blob/8af48ceac8c580aa91c89bedc906f06fbba225d7/client-core/src/service/root_hash_service.rs#L67

PR #675, opened 2019-12-12
Problem: (CRO-643) client-rpc does not retry connections to tendermint

devashishdxt:
Solution: Added retry logic in case of I/O failure

Sample log message:

```
[2019-12-12T09:00:20Z ERROR client_rpc::server] Initialization error: Unable to connect to websocket RPC at: ws://localhost:26657/websocket
     => WebSocketError: I/O failure
```

### DIFF
--- a/client-rpc/src/server.rs
+++ b/client-rpc/src/server.rs
@@ -5,6 +5,8 @@ use crate::rpc::sync_rpc::{SyncRpc, SyncRpcImpl};
 use crate::rpc::transaction_rpc::{TransactionRpc, TransactionRpcImpl};
 use crate::rpc::wallet_rpc::{WalletRpc, WalletRpcImpl};
 use std::net::SocketAddr;
+use std::thread;
+use std::time::Duration;
 
 use chain_core::init::network::{get_network, get_network_id, init_chain_id};
 use chain_core::tx::fee::LinearFee;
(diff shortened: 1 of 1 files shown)


yihuang:
Do we embed retry behavior like this to all services? 
Another solution if using some sort of process manager to re-start the whole thing whatever reason it crashes.
Sometimes, it might be a bit of confusion when process just hanging there when it can't start.

tomtau:
> Do we embed retry behavior like this to all services?
> Another solution is using some sort of process manager to re-start the whole thing whatever reason it crashes.
> Sometimes, it might be a bit of confusion when process just hanging there when it can't start.

@yihuang yeah, I agree that infinite looping is a bit confusing -- if someone runs this and doesn't set up an alert for error logs, they may come back to a seemingly running process, but they'll just see a log spammed with errors that it couldn't connect.
Ideally, there should be some exponential backoff or upper bound on retries, after which it gives up -- could be improved in a different PR

yihuang:
> > Do we embed retry behavior like this to all services?
> > Another solution is using some sort of process manager to re-start the whole thing whatever reason it crashes.
> > Sometimes, it might be a bit of confusion when process just hanging there when it can't start.
> 
> @yihuang yeah, I agree that infinite looping is a bit confusing -- if someone runs this and doesn't set up an alert for error logs, they may come back to a seemingly running process, but they'll just see a log spammed with errors that it couldn't connect.
> Ideally, there should be some exponential backoff or upper bound on retries, after which it gives up -- could be improved in a different PR

Or add an option to make it optional, I guess this feature is mainly for the convenience of development?

tomtau:
I suppose the original motivation was development / integration tests, so making it optional is an option as well :)

PR #677, opened 2019-12-12
Problem (CRO-578 WIP): client keeps loading and decrypting the same information in loops

yihuang:
Solution:
- Cache wallet state during synchronization.
- Refactored synchronizer into `WalletSyncer`, and moved to `wallet/syncer.rs`.
- Moved block/transaction handler to `wallet/syncer_logic.rs`, and only work with in-memory data structure.
- Provided simplified disk services api, load/save objects in one pass.
- Renamed `GlobalState` to `SyncState`.
- Removed `MannualSynconizer` in favor of `WalletSyncer`, and moved `PollingSynconizer` to `synchronizer.rs`. 

Other changes to the codebase:
- Add an `err_kind` api to `ResultExt`, the reason is more compact code format then `chain`, e.g.:
```
            .chain(|| {
                (
                    ErrorKind::DeserializationError,
                    format!("Unable to deserialize to_address ({})", to_address),
                )
            })
```
VS
```
                .err_kind(ErrorKind::InvalidInput, || {
                    format!("wallet private key not found: {}", self.name)
                })?,
```
Please feel free to comment on the necessity of this or the naming choice.

- Make `Clone` a requirement of `Storage`,`Client`,`TransactionObfuscation`, the reason is many usage places of these traits do need `Clone`, so no need to populate `Clone` trait in those usage locations. 

Data Contention
=============

Since we cached data in memory now, there could be data contention problems from concurrent tasks. currently we have these tasks which uses the wallet states:

Wallet Syncing
----------------

* Long-lived
* Readonly `Wallet`, cached during the whole syncing process.
* Read/Write SyncState, WalletState

Create address
-----------------

* Short-lived
* Modify `Wallet` directly to disk.

Other short-lived queries
---------------------------

* Load and decrypt data from disk.

---
This change is [Reviewable](https://reviewable.io/reviews/crypto-com/chain/677)


client-core/src/wallet/syncer.rs
```diff
@@ -0,0 +1,441 @@
...
+    F: Fn(&[TxId], &PublicKey) -> Result<Vec<Transaction>> + Clone + Send + Sync,
+{
+    fn decrypt_tx(&self, txids: &[TxId], viewkey: &PublicKey) -> Result<Vec<Transaction>> {
+        self(txids, viewkey)
+    }
+}
+
+/// Load view key and decrypt transaction with `TransactionObfuscation` and `KeyService`
+#[derive(Clone)]
+pub struct TxObfuscationDecryptor<O: TransactionObfuscation, S: SecureStorage> {
```
tomtau: I guess it's all right to separate this out, containing all "external state things" needed as inputs for  `TransactionObfuscation`.
I'd need to look at it in more detail what its life-span is and whether it makes sense to hold all those things

tomtau:
bors try

yihuang:
> ad data contentions -- shouldn't this be prevented by `WalletStateMemento`, or perhaps there may be contentions on data outside walletstate?

Current syning process is like this:
```
start sync
  load Wallet/WalletState/SyncState in memory
  handle_blocks
    mutate SyncState/WalletState
    save SyncState/WalletState
```
So if there are other tasks need to make modifications to `WalletState` or `SyncState` to the same wallet concurrently, they need to be synchronized. But AFAIK, it seems currently no other task needs to modify `WalletState` and `SyncState`, there are only tasks who mutates `Wallet`, which is read-only for the syncing process, so I guess it's fine.

client-common/src/storage.rs
```diff
@@ -60,10 +61,29 @@ pub trait Storage: Send + Sync {
 
     /// Returns all the keyspaces currently available.
     fn keyspaces(&self) -> Result<Vec<Vec<u8>>>;
+
+    /// load and deserialize object
+    fn load<T: Decode>(&self, keyspace: &str, key: &str) -> Result<Option<T>> {
```
devashishdxt: `Storage` and `SecureStorage` traits are supposed to be agnostic of de/serialisation. If you need a function to automatically de/serialise using `parity_scale_codec`, you can create that function outside of this trait which internally uses `get()` and `set()`.

This applies to all the new functions in these traits which are specific to `parity_scale_codec`.

client-core/src/service.rs
```diff
@@ -1,20 +1,86 @@
...
+pub fn delete_sync_state<S: Storage>(storage: &S, name: &str) -> Result<()> {
+    storage.delete(SYNC_STATE_KEYSPACE, name)?;
+    Ok(())
+}
+
+/// Delete wallet state from storage
+pub fn delete_wallet_state<S: Storage>(storage: &S, name: &str) -> Result<()> {
+    storage.delete(WALLET_STATE_KEYSPACE, name)?;
+    Ok(())
+}
```
devashishdxt: The idea of having these functions inside their own service was separation-of-concerns. Having these functions outside of their service goes against the principle of SoC. I'd suggest removing these functions to keep the design more modular.

client-core/src/service/wallet_service.rs
```diff
@@ -9,13 +9,19 @@ use chain_core::state::account::StakedStateAddress;
 use chain_core::tx::data::address::ExtendedAddr;
 use client_common::{Error, ErrorKind, PublicKey, Result, ResultExt, SecureStorage, Storage};
 
-const KEYSPACE: &str = "core_wallet";
+/// Key space of wallet
+pub const KEYSPACE: &str = "core_wallet";
```
devashishdxt: This should not be made `pub`.

client-core/src/service/wallet_state_service.rs
```diff
@@ -192,20 +187,19 @@ impl Default for WalletState {
...
 
     /// Applies a memento operation to wallet state
-    fn apply_memento_operation(&mut self, memento_operation: &MementoOperation) -> Result<()> {
+    fn apply_memento_operation(&mut self, memento_operation: &MementoOperation) {
         match memento_operation {
             MementoOperation::AddTransactionChange(ref transaction_id, ref transaction_change) => {
                 if !self.transaction_history.contains_key(transaction_id) {
-                    self.balance = (self.balance + transaction_change.balance_change)?;
+                    self.balance = (self.balance + transaction_change.balance_change)
+                        .expect("balance overflow");
```
devashishdxt: Handle error instead of `.expect()`

client-core/src/service/wallet_state_service.rs
```diff
@@ -192,20 +187,19 @@ impl Default for WalletState {
...
-            self.apply_memento_operation(operation)?
+            self.apply_memento_operation(operation);
         }
-
-        Ok(())
     }
 
     /// Applies a memento operation to wallet state
-    fn apply_memento_operation(&mut self, memento_operation: &MementoOperation) -> Result<()> {
+    fn apply_memento_operation(&mut self, memento_operation: &MementoOperation) {
```
devashishdxt: This function should not be infallible.

client-common/src/storage.rs
yihuang: Ok, I can move them to service mod.

client-core/src/service/wallet_state_service.rs
tomtau: I think `expect` here may be fine -- this would only fail if `self.balance + change` went over maximum total supply... is `transaction_change` validated or could it be random user input?

client-core/src/service.rs
yihuang: I just thought it might be too many changes a time if I modified too much in the specific service modules, which are depended by others.
I believe in the end, we only need three methods(`load`/`save`/`modify`) for each of `Wallet`/`WalletState`/`SyncState`.

client-core/src/service/wallet_state_service.rs
tomtau: if it's validated, expect is fine. If it's a random user input (e.g. balance_change could be `Coin::max()`), then Result make sense

devashishdxt: Right now, `AddTransactionChange` will only come from wallet synchroniser. But, I can see this functionality being used when user has the transaction data and manually enters (lite client) it instead of full-sync.

yihuang: `Coin` overflow is a phenomenon of a deep bug in the whole system, which I think we should probably just let it crash. Even with `-doverflow-check` on crate level in the future.

client-core/src/service/wallet_state_service.rs
yihuang: Other then the `Coin` addition overflow, there are no other error condition in this function.

client-core/src/service.rs
yihuang: I think it's good to move them into different service modules for now, I'll do it.

devashishdxt: We can make those changes later (or in this PR? Your choice.) if needed. But, this seems to be an unnecessary hack.

client-core/src/service/wallet_state_service.rs
yihuang: > Right now, `AddTransactionChange` will only come from wallet synchroniser. But, I can see this functionality being used when user has the transaction data and manually enters (lite client) it instead of full-sync.

We should check user-input at the interface level.

devashishdxt: I don't think we should think of `Coin` overflow in this way. If this is the case, why not just panic at the source, i.e., in `Add` impl.

yihuang: > I don't think we should think of `Coin` overflow in this way. If this is the case, why not just panic at the source, i.e., in `Add` impl.

That is indeed a good option, `Add` panic, if you want `Option`, use `checked_add`.

client-core/src/wallet/syncer.rs
yihuang: > I guess it's all right to separate this out, containing all "external state things" needed as inputs for `TransactionObfuscation`.
> I'd need to look at it in more detail what its life-span is and whether it makes sense to hold all those things

`TxDecryptor` is mainly useful when doing the mocking test, you can simply write a closure to provide the functionality.
Another thing is, in the future, the passphrase stored in `TxObfuscationDecryptor` might be different from the passphrase used to store `WalletState`.

client-core/src/service/wallet_service.rs
yihuang: Done

client-core/src/service.rs
yihuang: Done

client-common/src/storage.rs
yihuang: May I argue for this change a bit more ;D, because it's really convenient this way, and I think we can have a "default" serialization choice for `Storage`, so `parity_scale_codec` is not some arbitrary format, but the default format we use for `Storage`.

client-core/src/service/wallet_state_service.rs
tomtau: panicing at `Add impl.` of `Coin` would cause a network shutdown if there was a way to sneak in / gossip invalid transaction that causes an overflow, as all chain-abci would panic at the same time.

crashing a single wallet / client due to a bug (e.g. not validating its local user input) is a different story.

yihuang: I take back this `most of coin operations .. unwrap` statement, it probably just my wrong impression.
Talk about this case, if we are not sure if it's only used for validated block data, we might should check the overflow.
I guess we need some kind of guideline on this situation, or this discussion may happen again another time.

devashishdxt:
> > ad data contentions -- shouldn't this be prevented by `WalletStateMemento`, or perhaps there may be contentions on data outside walletstate?
> 
> Current syning process is like this:
> 
> ```
> start sync
>   load Wallet/WalletState/SyncState in memory
>   handle_blocks
>     mutate SyncState/WalletState
>     save SyncState/WalletState
> ```
> 
> So if there are other tasks make modifications to `WalletState` or `SyncState` to the disk directly, we have problem. But AFAIK, it seems currently no other task needs to modify `WalletState` and `SyncState`, there are tasks modify `Wallet`, but `Wallet` is read-only to the syncing process, so I guess it's fine.

I don't think we should depend on this assumption. Just because there is no other code component updating `WalletState` and `SyncState` does not mean that this won't happen in future.

client-common/src/storage.rs
devashishdxt: Ok.

tomtau:
> > > ad data contentions -- shouldn't this be prevented by `WalletStateMemento`, or perhaps there may be contentions on data outside walletstate?
> > 
> > 
> > Current syning process is like this:
> > ```
> > start sync
> >   load Wallet/WalletState/SyncState in memory
> >   handle_blocks
> >     mutate SyncState/WalletState
> >     save SyncState/WalletState
> > ```
> > 
> > 
> > So if there are other tasks make modifications to `WalletState` or `SyncState` to the disk directly, we have problem. But AFAIK, it seems currently no other task needs to modify `WalletState` and `SyncState`, there are tasks modify `Wallet`, but `Wallet` is read-only to the syncing process, so I guess it's fine.
> 
> I don't think we should depend on this assumption. Just because there is no other code component updating `WalletState` and `SyncState` does not mean that this won't happen in future.

I agree, any quick solutions for that? Probably not in this PR, but perhaps switching to `sqlite` may make this easier (if one ensures that each `sqlite` connection is held at one thread at a time, then it'll guarantee high-level transaction transaction serializability.)

yihuang:
> > > ad data contentions -- shouldn't this be prevented by `WalletStateMemento`, or perhaps there may be contentions on data outside walletstate?
> > 
> > 
> > Current syning process is like this:
> > ```
> > start sync
> >   load Wallet/WalletState/SyncState in memory
> >   handle_blocks
> >     mutate SyncState/WalletState
> >     save SyncState/WalletState
> > ```
> > 
> > 
> > So if there are other tasks make modifications to `WalletState` or `SyncState` to the disk directly, we have problem. But AFAIK, it seems currently no other task needs to modify `WalletState` and `SyncState`, there are tasks modify `Wallet`, but `Wallet` is read-only to the syncing process, so I guess it's fine.
> 
> I don't think we should depend on this assumption. Just because there is no other code component updating `WalletState` and `SyncState` does not mean that this won't happen in future.

I get it now, so the modifications needs to be done to `WalletState` is already encapsulated in`TransactionChange`, we can accumulated `TransactionChange`s for a batch of blocks, do one `compare_and_swap` to update `WalletState`.
For `SyncState`, I guess we can assume exclusive access in the syncing process, the current behavior is similar.

client-core/src/wallet/syncer.rs
yihuang: Actually, since there is only one viewkey per wallet, we should remove the `viewkey` paramter from `decrypt_tx`. And `TxObfuscationDecryptor` can hold the private view key in memory, rather than load and decrypt every time.

yihuang:
* `handle_blocks` handle a batch of blocks at a time.
* `WalletState` is modified atomically.
* `tx_query` requests are batched.

client-core/src/wallet/syncer.rs
```diff
@@ -0,0 +1,546 @@
...
+        )?;
+        Ok(())
+    }
+
+    fn handle_batch(&mut self, blocks: NonEmpty<BlockHeader>) -> Result<()> {
+        let enclave_txids = blocks
+            .iter()
+            .flat_map(|block| block.enclave_transaction_ids.iter().copied())
+            .collect::<Vec<_>>();
+        let enclave_txs = self.env.decryptor.decrypt_tx(&enclave_txids)?;
```
tomtau: Have you tested this? What if `enclave_txids` is an empty vector?

yihuang: Yes, it passes `cargo test` and integration tests locally.
The result of `decrypt_tx` depends on the obfuscation implementation, maybe we should check emptiness to avoid the call to `decrypt_tx`.

yihuang: Currently, `BlockGenerator` doesn't support tx mocking yet, after that implemented, we can test all the syncer behavior in unit testing.

yihuang:
> could you also squash commits and rebase it on top of the latest master?

Squashed and rebased.
Add empty input check in `TransactionObsfucation` implementation for effiency.

tomtau:
bors try

yihuang:
Just altered an incorrect code comment.

client-core/src/wallet/syncer.rs
```diff
@@ -0,0 +1,546 @@
...
+                &block,
+                &block_result,
+            )?))
+        } else {
+            Ok(None)
+        }
+    }
+}
+
+fn decrypt_staking_transactions(
```
tomtau: There's no decryption in these, as a staking address is plain

client-core/src/wallet/syncer.rs
```diff
@@ -0,0 +1,546 @@
...
+    let valid_transaction_ids = block_result.transaction_ids()?;
+    let block_filter = block_result.block_filter()?;
+
+    let staking_transactions =
+        decrypt_staking_transactions(&block_result, &wallet.staking_addresses(), block)?;
+
+    let enclave_transaction_ids = if block_filter.check_view_key(&wallet.view_key.clone().into()) {
+        block.enclave_transaction_ids()?
+    } else {
+        vec![]
```
tomtau: The logic depending on an empty vector may be a bit puzzling / indirect: one may expect that information in `BlockHeader` is kind of "raw" / wallet-independent (just processing what's returned from RPC), but this creates an extra hidden assumption around the view key.

client-core/src/service/wallet_state_service.rs
```diff
@@ -192,11 +230,10 @@ impl Default for WalletState {
 
 impl WalletState {
     /// Applies memento to wallet state
-    fn apply_memento(&mut self, memento: &WalletStateMemento) -> Result<()> {
+    pub fn apply_memento(&mut self, memento: &WalletStateMemento) -> Result<()> {
```
tomtau: I guess this is now public as `WalletStateMemento` can be created outside `WalletState`? not sure if intended

yihuang: Yes, before this, all the operations on `WalletState` go through `WalletStateService`, now outside code will access `WalletState` directly.

client-core/src/wallet/syncer.rs
yihuang: But `prepare_block_header` already filter staking transactions for current wallet, that's why I think it's ok to move this logic here.

client-core/src/wallet/syncer.rs
yihuang: Oh, I mistakenly took the `unencrypted` as `decrypted`, I'll change it.

client-core/src/wallet/syncer.rs
yihuang: Maybe we should move `struct BlockHeader` into `syncer`, and make it private, there are no other code use it.

yihuang: And the `Header` part of the name is not accurate, because it actually contains transactions. So maybe change it into `FilteredBlock`?

tomtau: It's just an extra implicit invariant around `BlockHeader` (one could still construct BlockHeader that breaks this invariant). To make it more explicit, as you proposed, you can move to syncer, make it private and rename as `FilteredBlock`.

(The other option would be to move this prepare function into `impl FilteredBlock/BlockHeader ` and make it possible to only construct via this function, e.g. by making some of those fields private... but the first option may be simpler)

yihuang: Done both.

yihuang:
- Moved `BlockHeader` into `wallet/syncer.rs`, renamed to `FilteredBlock`, change `prepare_block_header`to be constructor method of `FilteredBlock`, renamed to `from_block`.

- `spent_outputs` use `NonEmpty`.

- Remove two lines of debug `println!` in test code.

yihuang:
> looks ok, but it's a huge PR (preferable next time, this could be split up), so I may have overlooked something -- if you'd like to look at some particular code section, let me know.
> 
> I left a few comments, but there could addressed in other PRs

I think further improvement needs here is unit testing, ideally `BlockGenerator` can generate legitimate blocks contains transactions, that means mocking the transaction logic, which would take some time I think, I'll open an issue for that.

client-core/src/wallet/syncer.rs
yihuang: Done, changed to `filter_staking_transactions`.

yihuang:
Force pushed to resolve conflicts.

tomtau:
> lgtm
> can we use other name such ash "callback or delegate" for memento?
> also caching for latest state makes sense.

it's called memento, because that's what is usually called: https://en.wikipedia.org/wiki/Memento_pattern

PR #688, opened 2019-12-16
Problem: (CRO-635) Transaction builder is not Node.js lib friendly

calvinlauco:
Solution: Refactor RawTransferTransactionBuilder and related functions

---

- Rename existing `fee` method to `required_fee`; `fee` method now calculates difference between inputs and outputs
- Extract `get_tx_query` method into DefaultTransactionObfuscation
- Extract transaction obfuscation to until needed
  - Creating incomplete transaction now does not require `TxObfuscation` and can be done offline


tomtau:
bors try

PR #690, opened 2019-12-17
Problem:(cro-560) no default --genesis_app_hash option for chain-abci

leejw51crypto:
Solution: when user didn't specify apphash, and tendermint apphash already exists, use it automatically.

### DIFF
--- a/client-cli/src/command.rs
+++ b/client-cli/src/command.rs
@@ -45,7 +45,14 @@ use crate::{ask_passphrase, storage_path, tendermint_url};
 #[derive(Debug, StructOpt)]
 #[structopt(
     name = "client-cli",
-    about = "Basic CLI tool for interacting with Crypto.com Chain"
+    about = r#"Basic CLI tool for interacting with Crypto.com Chain
+
+ENVIRONMENT VARIABLES:
+    CRYPTO_CLIENT_DEBUG             Set to `true` for detailed error messages (Default: `false`)
+    CRYPTO_CHAIN_ID                 Chain ID of Crypto.com Chain
+    CRYPTO_CLIENT_STORAGE           Storage directory (Default: `.storage`)
+    CRYPTO_CLIENT_TENDERMINT        Websocket endpoint for tendermint (Default: `ws://localhost:26657/websocket`)
+"#
 )]
 pub enum Command {
     #[structopt(name = "wallet", about = "Wallet operations")]


chain-abci/src/main.rs
```diff
@@ -84,6 +85,21 @@ fn get_enclave_proxy(opts: &AbciOpt) -> MockClient {
     MockClient::new(get_network_id())
 }
 
+pub fn get_tendermint_filename() -> String {
+    match std::env::var("TENDERMINT_HOME") {
```
tomtau: I think the standard env variable used is `TM_HOME` or `TMHOME`: https://github.com/tendermint/tendermint/search?q=tm_home&unscoped_q=tm_home
```suggestion
    match std::env::var("TM_HOME") {
```

chain-abci/src/main.rs
```diff
@@ -84,6 +85,21 @@ fn get_enclave_proxy(opts: &AbciOpt) -> MockClient {
...
+        Ok(path) => format!("{}/config/genesis.json", path).to_owned(),
+        Err(_) => format!("/Users/leejw51/.tendermint/config/genesis.json").to_owned(),
```
tomtau: just return the error, don't default to `leejw51` -- it's not a default username of everyone

chain-abci/src/main.rs
```diff
@@ -1,17 +1,17 @@
...
 #[cfg(feature = "mock-validation")]
 use chain_abci::enclave_bridge::mock::MockClient;
 #[cfg(not(feature = "mock-validation"))]
 use chain_abci::enclave_bridge::ZmqEnclaveClient;
 use chain_abci::storage::*;
 use chain_core::init::network::{get_network, get_network_id, init_chain_id};
+use log::info;
 #[cfg(feature = "mock-validation")]
 use log::warn;
+use std::fs::{self};
```
devashishdxt:
```suggestion
use std::fs;
```

chain-abci/src/main.rs
```diff
@@ -41,6 +41,7 @@ struct AbciOpt {
     )]
     host: IpAddr,
     #[structopt(
+        default_value = "",
```
devashishdxt: Instead of providing a default blank string. Can we make `genesis_app_hash: Option<String>`?

chain-abci/src/main.rs
devashishdxt: It's `TMHOME`. Changed in suggestion.

chain-abci/src/main.rs
devashishdxt: Maybe instead of error, return `$HOME/.tendermint/config.genesis.json`?

chain-abci/src/main.rs
```diff
@@ -96,12 +112,19 @@ fn main() {
     let proxy = get_enclave_proxy(&opt);
 
     let addr = SocketAddr::new(opt.host, opt.port);
+
+    let mut app_hash = opt.genesis_app_hash.clone();
+    if opt.genesis_app_hash.len() == 0 {
```
devashishdxt: You can use `match` or `if let` here if `genesis_app_hash` is `Option<String>`.

chain-abci/src/main.rs
leejw51crypto: oh, i'll fix it.

chain-abci/src/main.rs
leejw51crypto: i'll try with if let

chain-abci/src/main.rs
leejw51crypto: ok

tomtau:
a small code style issue: https://travis-ci.org/crypto-com/chain/jobs/626084896#L4435

leejw51crypto:
i'll modify now

PR #695, opened 2019-12-17
Problem:(CRO-557) no way to import transactions (without a view key)

linfeng-crypto:
solution:
- add a interface `export_tx` for sender to get the plain transaction
- add a interface `import_tx` for receiver to import the plain transaction and get the output UTXO

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ of Rust SGX SDK (0.1.0 used a beta version of 1.1.0).
 
 ### Breaking changes
 * *client* [703](https://github.com/crypto-com/chain/pull/703): HD wallet generate view key with a different account index.
+* *client* [695](https://github.com/crypto-com/chain/pull/695): export and import transaction -- transactions that do not include receiver's view key can be exported, giving a base64 encoded plain transaction string which can be imported by the receiver.
 
 ### Improvements
 * *dev-utils* [692](https://github.com/crypto-com/chain/pull/692): dev-utils init command logs error when it goes wrong
@@ -25,6 +26,7 @@ of Rust SGX SDK (0.1.0 used a beta version of 1.1.0).
 
 The release is a more complete alpha version meant to be deployed on the second iteration of the public testnet.
 There are no guarantees on future API and binary compatibility at this stage.
+
 
 ## v0.1.0
 
--- a/client-common/src/transaction.rs
+++ b/client-common/src/transaction.rs
@@ -10,6 +10,15 @@ use chain_core::tx::data::{Tx, TxId};
 use chain_core::tx::witness::TxWitness;
 use chain_core::tx::TransactionId;
 
+/// A struct which the sender can download and the receiver can import
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize, Encode, Decode)]
+pub struct TransactionInfo {
+    /// enum Transaction type
+    pub tx: Transaction,
+    /// block height when the tx broadcast
+    pub block_height: u64,
+}
+
 /// Enum containing different types of transactions
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize, Encode, Decode)]
 #[serde(tag = "type")]
--- a/client-core/src/transaction_builder/unauthorized_wallet_transaction_builder.rs
+++ b/client-core/src/transaction_builder/unauthorized_wallet_transaction_builder.rs
@@ -4,9 +4,10 @@ use chain_core::tx::data::address::ExtendedAddr;
 use chain_core::tx::data::attribute::TxAttributes;
 use chain_core::tx::data::output::TxOut;
 use chain_core::tx::TxAux;
-use client_common::{ErrorKind, Result, SignedTransaction};
+use client_common::{ErrorKind, PrivateKey, Result, SignedTransaction, Transaction};
 
 use crate::{UnspentTransactions, WalletTransactionBuilder};
+use chain_core::tx::data::TxId;
 
 /// Implementation of `WalletTransactionBuilder` which always returns
 /// permission denied
@@ -27,6 +28,10 @@ impl WalletTransactionBuilder for UnauthorizedWalletTransactionBuilder {
     }
 
     fn obfuscate(&self, _: SignedTransaction) -> Result<TxAux> {
+        Err(ErrorKind::PermissionDenied.into())
+    }
+
+    fn decrypt_tx(&self, _txid: TxId, _private_key: &PrivateKey) -> Result<Transaction> {
         Err(ErrorKind::PermissionDenied.into())
     }
 }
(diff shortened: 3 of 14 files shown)


client-core/src/wallet.rs
```diff
@@ -185,6 +185,9 @@ pub trait WalletClient: Send + Sync {
 
     /// Broadcasts a transaction to Crypto.com Chain
     fn broadcast_transaction(&self, tx_aux: &TxAux) -> Result<BroadcastTxResponse>;
+
+    /// Get the plain transaction
```
tomtau: this comment just repeats the name, so not that helpful -- it'd be better to include some information that one can't read from the type signature... e.g. under what circumstances it could fail or what kind of `String` it returns (it seems to return base64-encoded ?)

client-common/src/transaction.rs
```diff
@@ -10,6 +10,15 @@ use chain_core::tx::data::{Tx, TxId};
 use chain_core::tx::witness::TxWitness;
 use chain_core::tx::TransactionId;
 
+/// A struct which the sender can download and the receiver can import
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize, Encode, Decode)]
+pub struct TransactionInfo {
+    /// enum Transaction type
+    pub tx: Transaction,
+    /// block height of the tx
```
tomtau: I guess block height when the transaction was committed?

client-cli/src/command.rs
```diff
@@ -111,6 +111,27 @@ pub enum Command {
...
+    DownloadTx {
+        #[structopt(name = "name", short, long, help = "Name of wallet")]
+        name: String,
+        #[structopt(name = "tx_id", short, long, help = "transaction id")]
+        tx_id: String,
+    },
+
+    #[structopt(
+        name = "sync_tx",
+        about = "Synchronize a plain transaction that view key not included"
```
tomtau: in `about`, it may good to list the transaction types this applies to

leejw51crypto: also how about
./client-cli transaction download --name a --type staking
sub-command of transaction

download_tx is too long to type

leejw51crypto:
overall, lgtm,
could you add some unit-test?

client-cli/src/command.rs
linfeng-crypto: yes, I think so.

client-core/src/transaction_builder/unauthorized_wallet_transaction_builder.rs
```diff
@@ -29,4 +30,8 @@ impl WalletTransactionBuilder for UnauthorizedWalletTransactionBuilder {
     fn obfuscate(&self, _: SignedTransaction) -> Result<TxAux> {
         Err(ErrorKind::PermissionDenied.into())
     }
+
+    fn decrypt_tx(&self, _txid: TxId, _private_key: &PrivateKey) -> Result<Transaction> {
```
yihuang: The only two methods of trait `TransactionObfuscation` get re-exported by trait `WalletTransactionBuilder` ;D

client-core/src/wallet/syncer.rs
```diff
@@ -374,6 +381,52 @@ impl<'a, S: SecureStorage, C: Client, D: TxDecryptor> WalletSyncerImpl<'a, S, C,
         Ok(())
     }
 
+    fn sync_tx(&mut self, tx_str: &str) -> Result<Coin> {
```
yihuang: I guess this is perfectly ok to write as a separate function, it's not "sync".

client-core/src/wallet/syncer.rs
```diff
@@ -374,6 +381,52 @@ impl<'a, S: SecureStorage, C: Client, D: TxDecryptor> WalletSyncerImpl<'a, S, C,
...
+            &self.wallet,
+            &self.wallet_state,
+            &mut memento,
+            &tx_info.tx,
+            tx_info.block_height,
+            block.header.time,
+            spent_flags?,
+        )
+        .chain(|| (ErrorKind::InvalidInput, "import error"))?;
+        self.save(&memento)?;
```
yihuang: This is the side-effect of implement this feature in `Syncer`, `save` will also write `SyncState` which is not used by this feature, but if there is concurrently running syncing process, it will override data.

client-core/src/wallet/syncer_logic.rs
```diff
@@ -127,6 +129,44 @@ fn on_transaction_change(
     }
 
     memento.add_transaction_change(transaction_change);
+    Ok(())
+}
+
+/// update WalletStateMemento when import a transaction
+pub(crate) fn import_transaction(
```
yihuang: Maybe we should rename `syncer_logic` to `wallet_logic` or something when it's function goes beyond syncing, to prevent future confusion.

client-rpc/src/rpc/wallet_rpc.rs
```diff
@@ -64,6 +64,9 @@ pub trait WalletRpc: Send + Sync {
 
     #[rpc(name = "wallet_transactions")]
     fn transactions(&self, request: WalletRequest) -> Result<Vec<TransactionChange>>;
+
+    #[rpc(name = "wallet_downloadTransaction")]
+    fn plain_transaction(&self, request: WalletRequest, txid: String) -> Result<String>;
```
yihuang: How about `sync_tx` -> `importTransaction`, `downloadTransaction` -> `exportTransaction`.

client-core/src/wallet/syncer.rs
linfeng-crypto: I'll make an individual function called `update_status` and use it in `save`

client-rpc/src/rpc/wallet_rpc.rs
linfeng-crypto: good

client-core/src/transaction_builder/unauthorized_wallet_transaction_builder.rs
linfeng-crypto: I think if I don't use the `WalletTransactionBuilder`, I have to put a `TransactionObfuscation` implied object into the `DefaultWalletClient` and  I have to deal with the `new_read_only` function where I have to create a new mock transaction obfuscation like `UnauthorizedObfuscation` and make impl the `TransactionObfuscation`, which is much expensive.

CHANGELOG.md
```diff
@@ -5,6 +5,11 @@
 The release is a more complete alpha version meant to be deployed on the second iteration of the public testnet.
 There are no guarantees on future API and binary compatibility at this stage.
 
+## v0.1.1
+
+### Features
+* *client*: export and import transaction -- transaction in which the receiver's view key not included can be exported by a give transaction id (the sender can list transactions and find it), getting a base64 encoded plain transaction string and the receiver can import the string into wallet.
```
tomtau: master is already unreleased 0.2, so perhaps no "0.1.1", just move the features paragraph under 0.2

tomtau:
it seems it fails to compile after merges

PR #698, opened 2019-12-18
Problem (CRO-642): no "watch-only" wallet mode in client

yihuang:
Solution:
- Add these apis to client-rpc:
  - restoreBasic(WalletRequest, String[hex encoded private view key])
  - createStakingWatchAddress(WalletRequest, PublicKey)
  - createTransferWatchAddress(WalletRequest, PublicKey)
  - listUTxO(WalletRequest)
- Add several paramters to transactions api:
  - transactions(WalletRequest, offset: usize, limit: usize, reversed: bool)
- Makes several api changes for testing convinience:
  - Add private option to `view_key` to get private view key.
  - Add `list_pub_key` command to client-cli to list public keys of
transfer addresses.
  - Also record the public key in wallet when creating transfer address.

Implemented integration tests in python.

### DIFF
--- a/client-core/src/service/wallet_service.rs
+++ b/client-core/src/service/wallet_service.rs
@@ -104,7 +104,8 @@ where
         WalletService { storage }
     }
 
-    fn get_wallet(&self, name: &str, passphrase: &SecUtf8) -> Result<Wallet> {
+    /// Load wallet
+    pub fn get_wallet(&self, name: &str, passphrase: &SecUtf8) -> Result<Wallet> {
         load_wallet(&self.storage, name, passphrase)?.err_kind(ErrorKind::InvalidInput, || {
             format!("Wallet with name ({}) not found", name)
         })
--- a/client-rpc/src/rpc/wallet_rpc.rs
+++ b/client-rpc/src/rpc/wallet_rpc.rs
@@ -3,6 +3,7 @@ use std::str::FromStr;
 
 use jsonrpc_core::Result;
 use jsonrpc_derive::rpc;
+use secstr::SecUtf8;
 
 use chain_core::init::coin::Coin;
 use chain_core::tx::data::access::{TxAccess, TxAccessPolicy};
@@ -11,10 +12,10 @@ use chain_core::tx::data::attribute::TxAttributes;
 use chain_core::tx::data::output::TxOut;
 use chain_core::tx::TxObfuscated;
 use chain_core::tx::{TxAux, TxEnclaveAux};
-use client_common::{PublicKey, Result as CommonResult};
+use client_common::{PrivateKey, PublicKey, Result as CommonResult};
 use client_core::types::WalletKind;
 use client_core::types::{AddressType, TransactionChange};
-use client_core::{Mnemonic, MultiSigWalletClient, WalletClient};
+use client_core::{Mnemonic, MultiSigWalletClient, UnspentTransactions, WalletClient};
 
 use crate::server::{rpc_error_from_string, to_rpc_error, WalletRequest};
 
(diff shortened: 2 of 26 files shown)


tomtau:
There's now a strict mode that fails builds on warnings (e.g. unused import)

tomtau:
bors try

tomtau:
One thing to check: are we using the hardened derivation for HD-wallet? otherwise by knowing the view key, one can get the rest

client-rpc/src/rpc/wallet_rpc.rs
```diff
@@ -132,6 +158,16 @@ where
         Ok(request.name)
     }
 
+    fn restore_watch(&self, request: WalletRequest, view_key: String) -> Result<String> {
```
tomtau: `view_key: SecUtf8`?

integration-tests/bot/tests/test_watch_only_wallet.py
```diff
@@ -0,0 +1,29 @@
+import time
+from chainrpc import RPC
+from uuid import uuid1
+
+rpc = RPC()
+
+
+def test_watch_only_wallet(addresses):
```
tomtau: could this test be easily added to one of the integration test pipelines (either or Travis or Drone)? If not, we can do it in another PR

client-rpc/src/rpc/wallet_rpc.rs
yihuang: > One thing to check: are we using the hardened derivation for HD-wallet? otherwise by knowing the view key, one can get the rest

I guess view key should be generated as a leaf, not the root key.

integration-tests/bot/tests/test_watch_only_wallet.py
yihuang: > could this test be easily added to one of the integration test pipelines (either or Travis or Drone)? If not, we can do it in another PR

I'll consult @leejw51

client-rpc/src/rpc/wallet_rpc.rs
yihuang: > `view_key: SecUtf8`?

Done

tomtau: > It's generated the same as a transfer address

It depends on whether it's hardened: https://bitcoin.stackexchange.com/questions/56916/derivation-of-parent-private-key-from-non-hardened-child

yihuang: > > It's generated the same as a transfer address
> 
> It depends on whether it's hardened: https://bitcoin.stackexchange.com/questions/56916/derivation-of-parent-private-key-from-non-hardened-child

Oh, I see, it's not hardened at the lowest level. Use a different account index for view key would ok then.

tomtau: not sure if this breaks anything -- @leejw51 @leejw51crypto @devashishdxt -- but I guess it'll be good to change, or add a warning fixme note

devashishdxt: It may break a few test cases. Also, the generated addresses will not be backwards compatible. But, I think we shouldn't worry about backwards compatibility right now because HD wallets were not a part of our last release.

client-core/src/service/wallet_service.rs
```diff
@@ -23,16 +23,19 @@ pub struct Wallet {
     pub staking_keys: BTreeSet<PublicKey>,
     /// root hashes of multi-sig transfer addresses
     pub root_hashes: BTreeSet<H256>,
+    /// is this wallet watch only
+    pub watch_only: bool,
```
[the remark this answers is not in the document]
yihuang: I think `WalletRole` is a name too broad for this. Actually, I'm thinking of eliminating this boolean variable, because it doesn't really make much difference. A watch-only wallet is actually a fully functioning `BasicWallet` since it already has view key pair.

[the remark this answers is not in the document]
yihuang: Changed `restoreWatch` to `restoreBasic`, removed the `watch_only` field in `Wallet`.

integration-tests/bot/tests/test_watch_only_wallet.py
yihuang: I've added it to the ci script, but can't test locally because of the docker image building failure, maybe try my luck on the CI server.

client-rpc/src/rpc/wallet_rpc.rs
yihuang: I'll implement this in a separate PR then.

tomtau:
bors try

integration-tests/bot/tests/test_watch_only_wallet.py
yihuang: It seems no luck with this, I'll revert this now, maybe try again another time.

PR #702, opened 2019-12-18
Problem: transfer private key can be compromised by leaked private view key

yihuang:
Solution: use different account_index(2) for view key derivation.

### DIFF
--- a/client-core/src/service/hd_key_service.rs
+++ b/client-core/src/service/hd_key_service.rs
@@ -16,7 +16,38 @@ const KEYSPACE: &str = "core_hd_key";
 struct HdKey {
     staking_index: u32,
     transfer_index: u32,
+    // curently only one viewkey per wallet, but it's good to keep it for uniformity.
+    viewkey_index: u32,
     seed: HDSeed,
+}
+
+/// Enum for specifying different types of accounts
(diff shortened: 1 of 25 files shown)


client-core/src/service/hd_key_service.rs
```diff
@@ -16,9 +16,40 @@ const KEYSPACE: &str = "core_hd_key";
 struct HdKey {
     staking_index: u32,
     transfer_index: u32,
+    // curently only one viewkey per wallet, but it's good to keep it for uniformity.
+    viewkey_index: u32,
```
leejw51crypto: because transfer_index & viewkey_index always the same,
how about use just one index?

lezzokafka: ` staking_index `and  `transfer_index` are supposed to be the index for Staking address and Transfer address?

leejw51crypto: yes,
so it's policy issue.
for ease of restoration addresses when you resotre hdwallet.

which do you think easier?

yihuang: Staking addresses and transfer addresses are independent, so `staking_index` and `transfer_index` will have different values.

PR #705, opened 2019-12-19
Problem: (CRO-577) Client allows weak passphrases

devashishdxt:
Solution: Estimate password strength using `zxcvbn` crate. Return `Error` when password score is less than 3.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -185,6 +185,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "bit-set"
+version = "0.5.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "bit-vec 0.5.1 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "bit-vec"
+version = "0.5.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
 name = "bit-vec"
 version = "0.6.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -556,6 +569,7 @@ dependencies = [
  "websocket 0.24.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "yasna 0.3.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "zeroize 1.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "zxcvbn 2.0.0 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


client-core/src/wallet/default_wallet_client.rs
```diff
@@ -620,3 +636,31 @@ where
...
+            }
+
+            for suggestion in feedback.suggestions() {
+                feedback_str.push_str(&format!("Suggestion: {} | ", suggestion));
+            }
+
+            if feedback_str.is_empty() {
+                feedback_str = "No feedback available!".to_string();
+            } else {
+                // Pop last three chars
```
yihuang: I guess we can (join)[https://doc.rust-lang.org/1.30.0/std/slice/trait.SliceConcatExt.html#tymethod.join] here.

devashishdxt: Done

devashishdxt:
bors try

Cargo.lock
```diff
@@ -184,6 +184,19 @@ dependencies = [
...
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "bit-vec 0.5.1 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "bit-vec"
+version = "0.5.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
```
tomtau: this one is a bit annoying / bloaty, as there's already `bit-vec 0.6.1`  (and `bitvec`).

probably open an issue for another dependency review / cleanup

client-core/src/wallet/default_wallet_client.rs
```diff
@@ -102,6 +103,21 @@ where
...
+        let password_entropy = estimate_password_strength(passphrase.unsecure(), &[name])
+            .chain(|| (ErrorKind::IllegalInput, "Blank passphrase"))?;
+
+        if password_entropy.score() < 3 {
+            return Err(Error::new(
+                ErrorKind::IllegalInput,
+                format!(
+                    "Weak passphrase: {}",
+                    parse_feedback(password_entropy.feedback().as_ref())
+                ),
```
tomtau: I guess this is probably fine, but not sure how annoying it'd be in quick testing -- we'll see, perhaps later there could be a "unsafe_development_only" flag to bypass that

devashishdxt:
bors try

client-core/src/wallet/default_wallet_client.rs
devashishdxt: We can add following code in future to only build this in release builds:
```rust
if !cfg!(debug_assertions) {
    // Check password score only in non-debug builds
}
```

devashishdxt:
bors try

devashishdxt:
bors try

PR #719, opened 2019-12-20
fix cargo clippy error in rust v1.40

linfeng-crypto:


tomtau:
I sneaked the patch here: https://github.com/crypto-com/chain/pull/711/commits/f3eed40780285a29b2bf94e1b2d582e3b71e08ca

PR #723, opened 2019-12-20
Problem: (CRO-577) Client uses weak hashing algorithm for passwords

devashishdxt:
Solution: Started using `rust-argon2`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -508,7 +508,6 @@ version = "0.2.0"
 dependencies = [
  "aes-gcm-siv 0.3.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "base64 0.11.0 (registry+https://github.com/rust-lang/crates.io-index)",
- "blake2 0.8.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "chain-core 0.2.0",
  "chain-tx-filter 0.2.0",
  "chrono 0.4.10 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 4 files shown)


devashishdxt:
bors try

devashishdxt:
> running the integration test now takes 2x time (before it was around 20 minutes from scratch, now around 40).
> passing around the hmac-ed keys from this strengthened secret (instead of re-computing in every request) as discussed on Slack should be preferable -- but I guess that would be require more changes in the client + tests. Perhaps in a different PR?
> 
> @leejw51crypto approved this but I guess if this gets merged, be prepared that on Monday, he'd be like "strang-eee, why is client so slow?" 😂

Passing around HMAC-ed keys will take substantial amount of work on client side. Both in backend and frontend. We can do that in different PRs. One way to reduce integration test times is to keep using `blake2` for debug builds and only use `argon2` for release builds. Any comments?

tomtau:
> > running the integration test now takes 2x time (before it was around 20 minutes from scratch, now around 40).
> > passing around the hmac-ed keys from this strengthened secret (instead of re-computing in every request) as discussed on Slack should be preferable -- but I guess that would be require more changes in the client + tests. Perhaps in a different PR?
> > @leejw51crypto approved this but I guess if this gets merged, be prepared that on Monday, he'd be like "strang-eee, why is client so slow?" 😂
> 
> Passing around HMAC-ed keys will take substantial amount of work on client side. Both in backend and frontend. We can do that in different PRs. One way to reduce integration test times is to keep using `blake2` for debug builds and only use `argon2` for release builds. Any comments?

 it's fine -- slow tests will annoy and motivate to prioritise that client work.

bors r+

leejw51crypto:
how about conditional compilation for integration test?
but password generation takes really tons of time, we may consider some lightway

devashishdxt:
> @leejw51crypto approved this but I guess if this gets merged, be prepared that on Monday, he'd be like "strang-eee, why is client so slow?" 😂

Turns out to be true 😂

leejw51crypto:
because it's too heavy operation, login attempt itself can be used as an attack.
server performance will be impacted much.
i propose like this
1. blacke2b
2. make argon2 hash in client side , <- wondering is it possible in javascript?
3. HMAC token method <- yes, it's viable option

PR #731, opened 2019-12-23
Problem: (CRO-647) Integration tests occasionally fails due to errors in tx-query enclave

devashishdxt:
Solution: Made change suggested here: https://software.intel.com/en-us/node/802449


devashishdxt:
bors try

devashishdxt:
bors try

PR #735, opened 2019-12-24
Problem: websocket rpc method id may overlap

yihuang:
Solution: Use sequencial atomic variable to generate unique method ids.

### DIFF
--- a/client-common/src/tendermint/websocket_rpc_client.rs
+++ b/client-common/src/tendermint/websocket_rpc_client.rs
@@ -6,15 +6,12 @@ pub use types::ConnectionState;
 
 use itertools::izip;
 use std::collections::HashMap;
-use std::iter;
 use std::sync::mpsc::{sync_channel, Receiver, SyncSender};
-use std::sync::{Arc, Mutex};
+use std::sync::{atomic::AtomicUsize, atomic::Ordering, Arc, Mutex};
 use std::thread;
 use std::time::Duration;
 
 use base64;
-use rand::distributions::Alphanumeric;
-use rand::{thread_rng, Rng};
 use serde::Deserialize;
 use serde_json::{json, Value};
 use tendermint::{lite::verifier, validator};
@@ -39,6 +36,7 @@ pub struct WebsocketRpcClient {
     connection_state: Arc<Mutex<ConnectionState>>,
     websocket_writer: Arc<Mutex<Writer<TcpStream>>>,
     channel_map: Arc<Mutex<HashMap<String, SyncSender<JsonRpcResponse>>>>,
+    unique_id: Arc<AtomicUsize>,
 }
 
 impl WebsocketRpcClient {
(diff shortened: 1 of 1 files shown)


yihuang:
> I'll suggest to get rid of `HashMap` entirely and use [slab](https://crates.io/crates/slab) instead. `Slab` is the default choice for many networking crates. Any comments?

Yeah, slab makes sense here.

yihuang:
> > I'll suggest to get rid of `HashMap` entirely and use [slab](https://crates.io/crates/slab) instead. `Slab` is the default choice for many networking crates. Any comments?
> 
> Yeah, slab makes sense here.

`slab` reusing method ids is a problem, when receiver timeout, it removed the method id from slab, the same id gets reused by another request, then two responses with the same id may arrive.

client-common/src/tendermint/websocket_rpc_client.rs
```diff
@@ -126,7 +125,8 @@ impl WebsocketRpcClient {
         method: &str,
         params: &[Value],
     ) -> Result<(String, Receiver<JsonRpcResponse>)> {
-        let (message, id) = prepare_message(method, params)?;
+        let id = self.unique_id.fetch_add(1, Ordering::Relaxed).to_string();
```
devashishdxt:
```suggestion
        let id = self.unique_id.fetch_add(1, Ordering:: SeqCst).to_string();
```

yihuang: As I understands, `Relaxed` is enough to keep it unique.

devashishdxt: `Relaxed` performs almost no synchronisation. In our case, this value can be accessed from multiple threads, so, it'd be good to use `Ordering::SeqCst`. You can read more about this [here](https://en.cppreference.com/w/cpp/atomic/memory_order).

yihuang: > `Relaxed` performs almost no synchronisation. In our case, this value can be accessed from multiple threads, so, it'd be good to use `Ordering::SeqCst`. You can read more about this [here](https://en.cppreference.com/w/cpp/atomic/memory_order).

We only need the atomicity of the increment operation to get the unique id, no needs to synchronize with other operations, so I think it's fine with `Relaxed` here.

yihuang: I find some explanations [here](https://vorner.github.io/2018/03/25/Atomics.html#relaxed-order--sanity-just-for-me).

devashishdxt: If there are two separate requests from different threads, can you explain how will this work?

yihuang: > If there are two separate requests from different threads, can you explain how will this work?

No matter what the `Ordering` is, the operation is atomic, what's what `Atomic` always guarantee.
So two concurrent request happens, either one gets 1 and the other gets 2, or one gets 2 and the other gets 1.

devashishdxt: Ok

PR #737, opened 2019-12-24
Problem: block id field in vote could be empty

yihuang:
Solution:
- Fix the problem in tendermint-rs
- Upgrade dependency

tendermint-rs PR for this: https://github.com/interchainio/tendermint-rs/pull/103


leejw51:
thanks!
no more sync bug

PR #738, opened 2019-12-24
Problem (CRO-663): Not easy to build enclave within chain-abci.

yihuang:
Solution:
- Wrap and mock sgx at low level, make enclave buildable on non-sgx environment(without sgx security guarantee, good for development and unit testing, etc.).
- When mocking, enclave wrapper app calls enclave code directly, `cargo build` build mock version by default, use feature `mesalock_sgx` to switch to the real thing.
- `chain-abci` starts the tx validation app inside its own process, currently, validation app keeps its own storage, the logic is to be integrated deeper in the future. `chain-abci`'s enclave bridge calls tx validation app directly (not going through zmq anymore), `tx-query` still access tx validation through zmq.
- Almost build everything with cargo now, except linking and signing of enclave library which are still handled by makefile.
- Build everything in `docker/Dockerfile` now, both executables and enclaves, resulting in a small image with only executables, enclaves, and sgx-sdk runtime dependencies.

### Build mock
```
cargo build
```

### Build sgx
Setup sgx-sdk and env vars(`SGX_MODE`, `SGX_SDK`) first.
```
cargo build --no-default-features --features mesalock_sgx --manifest-path chain-abci/Cargo.toml
cargo build --no-default-features --features mesalock_sgx --manifest-path chain-tx-enclave/tx-query/app/Cargo.toml
cargo build --no-default-features --features mesalock_sgx --manifest-path chain-tx-enclave/tx-validation/app/Cargo.toml
make -C ./chain-tx-enclave/tx-query SGX_DEBUG=1
make -C ./chain-tx-enclave/tx-validation SGX_DEBUG=1
cargo build -p client-cli -p client-rpc -p dev-utils
```
All the executables and signed enclaves resides in `target/debug/`.

The behavior of feature `mesalock_sgx`:
- For enclave code:
  - `mesalock_sgx`: build with sgx std library, access real sgx functionality.
  - `default`: build with normal std, mocked sgx functionality.
- For enclave wrapper code:
  - `mesalock_sgx`: normal std, needs sgx library to load and call enclave.
  - `default`: normal std, calls enclave code(built with mock sgx) directly.


yihuang:
There's no rush to merge this PR I think, put it up for some review and discussions first, for decision making on `ADR-001`.

chain-tx-enclave/tx-validation/enclave/src/lib.rs
```diff
@@ -1,42 +1,39 @@
-#![crate_name = "txvalidationenclave"]
-#![crate_type = "staticlib"]
```
tomtau: @yihuang have you actually tried running the outcome of building NON-mock version in SGX environment ?

yihuang: > @yihuang have you actually tried running the outcome of building NON-mock version in SGX environment ?

Yeah, it's buildable with image `baiduxlab/sgx-rust:1804-1.1.0`, or you can try `docker/Dockerfile`.

tomtau: buildable != correct ... I spent a countless of hours debugging situations where the enclave code was successfully "built", but mysteriously crashed or returned error codes... because of different issues that went wrong in the compilation process: e.g. did a syscall or called an instruction not allowed inside SGX env

yihuang: Yeah, I'll continue to test it after the issue 786.

tomtau:
closing this down for now, as the diff is huge and the change set is quite complex (and has 0% coverage -- as a sidenote, it's possible to measure coverage of code in SGX env: https://github.com/apache/incubator-teaclave-sgx-sdk/tree/master/samplecode/sgx-cov ) and does some radical re-design.

For a careful review, it'll be better to split it up in smaller and more iterative sub-PRs: there are some directly related to the decision in https://github.com/crypto-com/chain/pull/789 , and there are some related to other concerns (e.g. testability of mock environment)

yihuang:
> closing this down for now, as the diff is huge and the change set is quite complex (and has 0% coverage -- as a sidenote, it's possible to measure coverage of code in SGX env: https://github.com/apache/incubator-teaclave-sgx-sdk/tree/master/samplecode/sgx-cov ) and does some radical re-design.
> 
> For a careful review, it'll be better to split it up in smaller and more iterative sub-PRs: there are some directly related to the decision in #789 , and there are some related to other concerns (e.g. testability of mock environment)

I believe the drop of test coverage mainly due the enclave code under the radar now which is not before? There's not much new code in this PR.

PR #740, opened 2019-12-25
Problem:(CRO-661) not enough logging in debug enclave execution

linfeng-crypto:
solution: add some logs.

### DIFF
--- a/chain-tx-enclave/tx-query/enclave/src/attest.rs
+++ b/chain-tx-enclave/tx-query/enclave/src/attest.rs
@@ -169,7 +169,7 @@ fn parse_response_attn_report(resp: &[u8]) -> Result<AttnReport, RAError> {
                     attn_report = str::from_utf8(resp_body)
                         .map_err(|_| RAError::ParseError)?
                         .to_string();
-                    // println!("Attestation report: {}", attn_report);
+                    println!("Attestation report: {}", attn_report);
                 }
             }
             "X-IASReport-Signature" => {
@@ -259,8 +259,8 @@ fn get_sigrl_from_intel(ias_key: &str, fd: c_int, gid: u32) -> Result<Vec<u8>, R
 
     match tls.read_to_end(&mut plaintext) {
         Ok(_) => (),
-        Err(_) => {
-            // println!("get_sigrl_from_intel tls.read_to_end: {:?}", e);
+        Err(e) => {
+            println!("get_sigrl_from_intel tls.read_to_end: {:?}", e);
             return Err(RAError::CommunicationError);
         }
     }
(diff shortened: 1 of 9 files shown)


chain-tx-enclave/tx-validation/enclave/src/validate.rs
```diff
@@ -68,12 +68,16 @@ fn construct_sealed_response(
 ) -> Result<IntraEnclaveResponse, sgx_status_t> {
     let to_seal = to_seal_tx.encode();
     match result {
-        Err(e) => Ok(Err(e)),
+        Err(e) => {
+            log::error!("encode tx witoutputs failed: {:?}", e);
+            Ok(Err(e))
+        },
```
tomtau: this is not a failure of encoding -- this will be any time invalid TX is submitted, and the reason for it being rejected is in `e`... no need to log here, as this is returned directly

enclave-protocol/src/lib.rs
```diff
@@ -68,24 +68,24 @@ pub fn is_basic_valid_tx_request(
...
             Some(ref i) if !i.is_empty() => Ok(()),
-            _ => Err(()),
+            _ => Err("sealed log is empty".into()),
         },
         TxEnclaveAux::WithdrawUnbondedStakeTx { .. } => {
             if request.account.is_some() {
                 Ok(())
             } else {
-                Err(())
+                Err("request account is empty".into())
```
tomtau: better to return `enum` listing all possible errors rather than a string

tomtau:
bors try

tomtau:
bors retry

tomtau:
bors try-

chain-tx-enclave/tx-query/enclave/src/attest.rs
```diff
@@ -657,7 +660,7 @@ pub(crate) fn get_tls_config() -> Arc<rustls::ServerConfig> {
             }
             Err(_) => {
                 // Poisoned
-                // println!("SVRCONFIGCACHE invalidate cache failed {}!", x);
+                println!("SVRCONFIGCACHE invalidate cache failed {}!", x);
```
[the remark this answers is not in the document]
tomtau: @linfeng-crypto there's a compilation error:
```
/chain/chain-tx-enclave/tx-query/enclave)
3389	error[E0425]: cannot find value `x` in this scope
3390	   --> src/attest.rs:663:72
3391	    |
3392	663 |                 println!("SVRCONFIGCACHE invalidate cache failed {}!", x);
3393	    |                                                                        ^ help: a local variable with a similar name exists: `s`
3394	
```

enclave-protocol/src/lib.rs
linfeng-crypto: I'll add a error.rs file and use enum errors

chain-tx-enclave/tx-validation/enclave/src/validate.rs
linfeng-crypto: fixed

chain-tx-enclave/tx-query/enclave/src/attest.rs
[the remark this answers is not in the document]
linfeng-crypto: fixed

PR #741, opened 2019-12-26
Problem (CRO-663): no clear decision on ADR-001

yihuang:
Solution: Update design docs for further discussions.

### DIFF
--- a/architecture-docs/adr-001.md
+++ b/architecture-docs/adr-001.md
@@ -3,6 +3,7 @@
 ## Changelog
 * 10-12-2019: Initial Draft
 * 11-12-2019: Extra comments
+* 26-12-2019: Add enclave development solution, add design candidates and proposal
 
 ## Context
 Currently (not counting Tendermint's internal storage or wallets), two processes maintain their internal storage:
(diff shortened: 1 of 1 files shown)


architecture-docs/adr-001.md
```diff
@@ -47,44 +48,151 @@ for these things.
...
+
+  - `tx-query` replicate `chain-abci`storage asynchronously, then access local storage directly.
+    - Positives: Best performance
+    - Negatives: Needs to implement storage replication, this should be a robust and generic storage replication solution, so there is no data consistency problem we have with seperate storage processes now, either implement with help of low level storage, or implement at application level with custom designed replication log structure.
+  - Client fetch sealed transactions through `abci_query`, then pass them to `tx-query` together with view key
+    - Positives: Simpler `tx-query` implementation, no `ocall`s
+    - Negatives: More client logic, more communication overhead
+  - `tx-query` fetch them from `chain-abci` through `abci_query` or `zmq`
+
+- Non privacy preserving
```
tomtau: not an option, shouldn't be there

yihuang:
> "non-privacy preserving" isn't an option, as it defeats the whole purpose.
> 
> instead, there could be a full node operator mode where one can submit a view-key signed request that the enclave would non-probabilistically index its transaction by public view key

By "non-privacy preserving", I mean to access a trusted full node, maybe set up locally. In this setup, I think it's ok to pass unsealed requests directly to the `tx-validation` enclave through `abci_query`?

tomtau:
I've just read through the changes in this PR and I'm a bit puzzled about it -- the purpose of ADR is to record a decision rather than a discussion:

- overall, this set of changes seems to solve a different problem than what was originally specified: if that's necessary, the original ADR should status should be marked as deprecated: https://github.com/crypto-com/chain/blob/master/architecture-docs/adr-template.md#status and a new one superseding it with a relevant context to that problem should be created
- I haven't read through https://github.com/crypto-com/chain/pull/738/files yet -- I guess that may give me more context

In any case, let's move the discussion to Slack and re-open this PR when it's clearer

PR #747, opened 2019-12-28
Problem: (CRO-679) cargo build failed on nightly of error description

calvinlauco:
Solution: Change to use to_string


devashishdxt:
bors retry

PR #779, opened 2019-12-30
Problem: (CRO-680) Latest sled fails to compile

devashishdxt:
Solution: Changed `Db::open` to `sled::open`. `Db::open` is now deprecated.


chain-tx-enclave/tx-validation/app/src/main.rs
```diff
@@ -24,7 +24,7 @@ fn main() {
         error!("Please provide the ZMQ connection string (e.g. \"tcp://127.0.0.1:25933\") as the first argument");
         return;
     }
-    let db = Db::open(storage_path()).expect("failed to open a storage path");
+    let db = sled::open(storage_path()).expect("failed to open a storage path");
```
calvinlauco: I think we can skip the `use sled::Db;` line in this file.

devashishdxt: Done.

Issue #840, opened 2020-01-07
Problem: incorrect generated withdraw transaction with an extra view key


tomtau:
It seems the estimated fee for a withdraw transaction is not correct if an extra view key is included, as it'll fail with input-output coin sum mismatch error

tomtau:
Or another detail was that I didn't run `sync` when creating the transaction

tomtau:
this may be fixed https://github.com/crypto-com/chain/pull/841 but re-open if it was not

PR #841, opened 2020-01-07
Problem: User may enter same view key multiple times in client-cli

devashishdxt:
Solution: Used `BTreeSet` instead of `Vec`. Fixes https://github.com/crypto-com/chain/issues/840


tomtau:
@devashishdxt does that really fix https://github.com/crypto-com/chain/issues/840 ? did you manage to reproduce it?

devashishdxt:
> @devashishdxt does that really fix #840 ? did you manage to reproduce it?

No. I was not able to reproduce https://github.com/crypto-com/chain/issues/840 locally. But, the changes in this PR are still relevant.

Issue #831, opened 2020-01-06
Problem: client sync fails if TM is catching up with the network


tomtau:
Some RPC endpoints (e.g. /validators) may not return the latest or historical information if TM node is not yet fully sync (didn't catch up with the network's latest block). as observed with @lezzokafka 

It'll be good if client could check `/status` endpoint before running sync, as otherwise it may fail with a confusing error message

yihuang:
Strange, wallet syncing always start with a `status` call, and only sync to the latest_height returned. Is it easy to reproduce?

lezzokafka:
@yihuang  
Yaa
For example, when joining a network with current block heigh of 100k, 
start the tendermint node and so on to sync blocks. While it is syncing, the sync will fail because it is still trying to catch up with the network. 
  
Similar issue with the tendermint rpc as well, say you have synced from block #1 to block #100, and it is still catching up with  the network, when you do something like `http://localhost:26657/block?height=50` to check block#50, the will be an error message with something like 
"Height must be less than or equal to the current blockchain height."

leejw51crypto:
is there any possible to ask blocks bigger than current height
which is received from status?

yihuang:
I reproduced, we are not requesting data with a height bigger the `latest_height` reported by `status`, I guess it's just not stable when tendermint catching up, maybe we should hold wallet syncing when tendermint catching up.

PR #842, opened 2020-01-07
Problem #831: client sync fails if TM is catching up with the network

yihuang:
Solution:
- refuse to sync when tendermint catching up.
- retry several times for temparary catching up status.

### DIFF
--- a/client-core/src/wallet/syncer.rs
+++ b/client-core/src/wallet/syncer.rs
@@ -303,6 +303,12 @@ impl<'a, S: SecureStorage, C: Client, D: TxDecryptor> WalletSyncerImpl<'a, S, C,
 
     fn sync(&mut self) -> Result<()> {
         let status = self.env.client.status()?;
+        if status.sync_info.catching_up {
+            return Err(Error::new(
+                ErrorKind::TendermintRpcError,
+                "Tendermint node is catching up with full node (retry after some time)",
+            ));
+        }
         let current_block_height = status.sync_info.latest_block_height.value();
         self.init_progress(current_block_height);
 


client-core/src/wallet/syncer.rs
```diff
@@ -302,7 +302,21 @@ impl<'a, S: SecureStorage, C: Client, D: TxDecryptor> WalletSyncerImpl<'a, S, C,
     }
 
     fn sync(&mut self) -> Result<()> {
-        let status = self.env.client.status()?;
+        let mut counter = 0;
```
devashishdxt: I don't think we need automatic retries here.

- `PollingSynchronizer` should retry in case of failure.
- When manually calling sync (from `client-cli` or `client-rpc`), user can call it again after sometime.

client-core/src/wallet/syncer.rs
```diff
@@ -302,7 +302,21 @@ impl<'a, S: SecureStorage, C: Client, D: TxDecryptor> WalletSyncerImpl<'a, S, C,
...
+        let status = loop {
+            let status = self.env.client.status()?;
+            if !status.sync_info.catching_up {
+                break Ok(status);
+            }
+            counter += 1;
+            if counter > 10 {
+                break Err(Error::new(
+                    ErrorKind::TendermintRpcError,
+                    "Tendermint node is still catching up",
```
devashishdxt:
```suggestion
                    "Tendermint node is catching up with full node (retry after some time)",
```

client-core/src/wallet/syncer.rs
yihuang: Ok, done.

client-core/src/wallet/syncer.rs
yihuang: Done

Issue #838, opened 2020-01-07
Problem: ambiguous error message for empty staking address with client-cli 


lezzokafka:
`Error: Tendermint RPC error: account lookup failed: account not found`

When checking a staking address with no transaction history

PR #845, opened 2020-01-07
Problem: Ambiguous error message for empty staking address

devashishdxt:
Solution: Added a more detailed error message. Fixes https://github.com/crypto-com/chain/issues/838

### DIFF
--- a/chain-tx-validation/src/lib.rs
+++ b/chain-tx-validation/src/lib.rs
@@ -126,7 +126,7 @@ impl fmt::Display for Error {
             // FIXME: IoError(ref err) => write!(f, "IO error: {}", err),
             IoError => write!(f, "database lookup error"),
             EnclaveRejected => write!(f, "enclave error or invalid TX"),
-            AccountNotFound => write!(f, "account not found"),
+            AccountNotFound => write!(f, "account does not exist for given staking address on blockchain"),
             AccountNotUnbonded => write!(f, "account not unbonded for withdrawal"),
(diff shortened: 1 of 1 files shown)


chain-tx-validation/src/lib.rs
```diff
@@ -126,7 +126,7 @@ impl fmt::Display for Error {
             // FIXME: IoError(ref err) => write!(f, "IO error: {}", err),
             IoError => write!(f, "database lookup error"),
             EnclaveRejected => write!(f, "enclave error or invalid TX"),
-            AccountNotFound => write!(f, "account not found"),
+            AccountNotFound => write!(f, "account does not exist for given staking address on blockchain"),
```
yihuang: I guess the main confusion in the issue is, the staking address do exists in the wallet, it just has empty transaction history on the blockchain, so a better solution maybe show a empty value to user. A default `StakingState` when query staking state, a empty list when query transaction history.

devashishdxt: We do not any command to show transaction history for a staking account. Only fetching current state is supported. When an account does not exist, we have 2 options:

1. Show a more detailed error message that the account for given staking address does not exist on blockchain.
1. Show a default state.

After discussing with @tomtau and @lezzokafka, we reached on conclusion to use option 1. But, we can show a default state if that seems to be more correct option.

yihuang: 👌

Issue #853, opened 2020-01-08
Problem: multi-node integration tests not executed on CI


tomtau:
Most of the multi-node integration tests *don't* need spinning up any enclaves (can use the mock compilation or use the SW emulation mode), as all transactions are public (unjail, nodejoin...) and entirely processed by chain-abci.

The current pipeline was disabled mainly because it wasn't stable and didn't do any cleaning: https://github.com/crypto-com/chain/blob/master/.drone.yml#L126

Instead of having a separate pipeline, the multi-node integration tests would ideally be added to the existing integration test pipeline -- either on Travis CI or on Drone.

@leejw51crypto Jason can help with this

leejw51crypto:
let's build once for all integration test.
and remove all compiling tmp files after test

PR #855, opened 2020-01-08
Problem:(CRO-587) no integration test for new validator join tx

leejw51:
add integration test
check compile result


integration-tests/bot/chainbot.py
```diff
@@ -340,7 +341,16 @@ def coin_to_voting_power(coin):
...
 
-async def gen_genesis(cfg):
+async def gen_genesis(cfg_orig):
+    cfg = copy.deepcopy(cfg_orig)
+    newnodes= []
+    for n in cfg["nodes"]:
+        print(n)
+        if n["bonded_coin"]> 0:
+            newnodes.append(n)
+    cfg["nodes"]= newnodes
```
yihuang: It's simpler to filter the nodes out with `bonded_coin == 0` in `app_state_cfg` when constructing `council_nodes`.

integration-tests/bot/chainrpc.py
```diff
@@ -93,8 +93,8 @@ def create_watch(self, public_key, name=DEFAULT_WALLET, type='staking', enckey=N
 
 
 class Wallet:
-    def enckey(self, name=DEFAULT_WALLET):
-        '''Get encryption key of wallet
+    def balance(self, name=DEFAULT_WALLET):
+        '''Get balance of wallet
```
yihuang: Is this a typo?

integration-tests/bot/chainrpc.py
```diff
@@ -183,6 +183,9 @@ def withdraw_all_unbonded(self, from_address, to_address, view_keys=None, name=D
     def unjail(self, address, name=DEFAULT_WALLET, enckey=None):
         return call('staking_unjail', [name, enckey or get_enckey()], fix_address(address))
 
+    def join(self, node_name, node_pubkey, node_staking_address, name=DEFAULT_WALLET):
+        return call('staking_validatorNodeJoin', [name, get_passphrase()], node_name, node_pubkey,  fix_address(node_staking_address))
```
yihuang:
```suggestion
        return call('staking_validatorNodeJoin', [name, enckey or get_enckey()], node_name, node_pubkey,  fix_address(node_staking_address))
```

integration-tests/bot/chainrpc.py
```diff
@@ -183,6 +183,9 @@ def withdraw_all_unbonded(self, from_address, to_address, view_keys=None, name=D
     def unjail(self, address, name=DEFAULT_WALLET, enckey=None):
         return call('staking_unjail', [name, enckey or get_enckey()], fix_address(address))
 
+    def join(self, node_name, node_pubkey, node_staking_address, name=DEFAULT_WALLET):
```
yihuang:
```suggestion
    def join(self, node_name, node_pubkey, node_staking_address, name=DEFAULT_WALLET, enckey=None):
```

integration-tests/bot/join_test.py
```diff
@@ -0,0 +1,331 @@
...
+    # tendermint rpc
+
+
+
+    def get_staking_state(self,name, passphrase, addr):
+        return self.rpc.staking.state(addr, name)
+       
+
+    def create_staking_address(self,name, passphrase):
+        return self.rpc.address.create(name,'staking')
```
yihuang: It needs to pass around encryption key(or called authentication token) now, or it probably won't run correctly.

integration-tests/bot/chainrpc.py
leejw51crypto: oh, merge is broken, i'll restore.

integration-tests/bot/chainbot.py
leejw51crypto: it's intended operation, i wan to add node which is not validator, but has big amount of bonds.

.drone.yml
```diff
@@ -126,36 +126,36 @@ trigger:
...
+     - success
+     - failure
+
+trigger:
+  branch:
+    - master
+    - staging
+    - trying
+  event:
+    - push
```
tomtau: shouldn't be re-enabled before FIXME notes are addressed -- https://github.com/crypto-com/chain/issues/853

integration-tests/bot/chainbot.py
yihuang: Isn't this part of code try to remove the node with `bonded_coin == 0`?

integration-tests/bot/join_test.py
```diff
@@ -0,0 +1,331 @@
...
+        self.node1_address1= ""
+        self.node1_transfer_address = ""
+        self.node1_mnemonics=""
+
+        # wallet b
+        self.node2_address = ""
+        self.node2_mnemonics=""
+
+
+
```
yihuang: Why so many blank lines ;D, you can run a python linter tools like flake8.

integration-tests/bot/make.py
```diff
@@ -4,16 +4,20 @@
...
     {'op': 'replace', 'path': '/nodes/1/bonded_coin', 'value':   1250000000000000000},
     {'op': 'replace', 'path': '/nodes/1/unbonded_coin', 'value': 3700000000000000000},
+    {'op': 'replace', 'path': '/nodes/2/bonded_coin', 'value':                     0},
+    {'op': 'replace', 'path': '/nodes/2/unbonded_coin', 'value':                   0},
     {'op': 'replace', 'path': '/nodes/0/base_port', 'value':26650},
     {'op': 'replace', 'path': '/nodes/1/base_port', 'value':26650},
+    {'op': 'replace', 'path': '/nodes/2/base_port', 'value':26650},
     {'op': 'replace', 'path': '/nodes/0/hostname', 'value':'chain0'},
     {'op': 'replace', 'path': '/nodes/1/hostname', 'value':'chain1'},
+    {'op': 'replace', 'path': '/nodes/2/hostname', 'value':'chain2'},
```
yihuang: You can set `hostname` parameter to `chain{index}`

integration-tests/bot/make.py
```diff
@@ -4,16 +4,20 @@
...
 a = CLI()
 print("make config")
-src=a._gen(count=2, chain_id='test-ab', expansion_cap=50000000000000000,  root_path='./', hostname='localhost')
+src=a._gen(count=3, chain_id='test-ab', expansion_cap=50000000000000000,  root_path='./', hostname='localhost')
 patch = jsonpatch.JsonPatch([
     {'op': 'replace', 'path': '/nodes/0/bonded_coin', 'value':   3750000000000000000},
     {'op': 'replace', 'path': '/nodes/0/unbonded_coin', 'value': 1250000000000000000},
     {'op': 'replace', 'path': '/nodes/1/bonded_coin', 'value':   1250000000000000000},
     {'op': 'replace', 'path': '/nodes/1/unbonded_coin', 'value': 3700000000000000000},
+    {'op': 'replace', 'path': '/nodes/2/bonded_coin', 'value':                     0},
```
yihuang: You can simply `src['nodes'][2].update(bonded_coin=0, unbonded_coin=0)`.

.drone.yml
leejw51crypto: ok

integration-tests/bot/join_test.py
leejw51crypto: ok, i'll remove it

.drone.yml
leejw51crypto: restored back

leejw51crypto:
fixed enckey, work more for above feedbacks

integration-tests/bot/join_test.py
leejw51crypto: yes, i modified using enckey

integration-tests/jail/Dockerfile
```diff
@@ -9,7 +9,7 @@ RUN sed -i 's/security.ubuntu.com/ftp.daum.net/g' /etc/apt/sources.list
 RUN sed -i 's/archive.ubuntu.com/ftp.daum.net/g' /etc/apt/sources.list
 RUN apt-get update
 RUN apt install -y --no-install-recommends whois wget curl python3 libssl-dev libcurl4-openssl-dev libelf-dev libdw-dev  gcc binutils-dev libc6-dev pkg-config build-essential openssh-server cmake libgflags-dev libzmq3-dev pkg-config libssl-dev libzmq3-dev  unzip  tmux 
-ENV SGX_MODE=SW
+ENV SGX_MODE=HW
 ENV NETWORK_ID=AB
 ENV RUSTFLAGS=-Ctarget-feature=+aes,+ssse3
 ENV PATH=$HOME/bin:$PATH
```
tomtau: are these docker files needed?

integration-tests/jail/disk/go_test.sh
```diff
@@ -20,6 +20,8 @@ source /opt/sgxsdk/environment
 source /root/.cargo/env
 echo "sgx mode=" $SGX_MODE
 echo "network id=" $NETWORK_ID
+echo "spid = " $SPID
+echo "ias_api_key = " $IAS_API_KEY
```
tomtau: these values are typically set as secrets on CI, so do not to print them

leejw51: ok

integration-tests/join/disk/go_test.sh
```diff
@@ -0,0 +1,34 @@
...
+echo "clear disk"
+/root/bin/tendermint unsafe_reset_all
+sleep 2 
+source /root/disk/prepare.sh
+source /opt/sgxsdk/environment
+source /root/.cargo/env
+echo "sgx mode=" $SGX_MODE
+echo "network id=" $NETWORK_ID
+echo "spid = " $SPID
+echo "ias_api_key = " $IAS_API_KEY
```
tomtau: again, as secrets, they shouldn't be printed out

leejw51crypto: removed

integration-tests/jail/Dockerfile
leejw51crypto: this is multiple nodes test, docker-compose is necessary.

tomtau: but one can just use one image (e.g. `cryptocom/chain:latest`) instead of building all these custom images and set this up in docker-compose with `command` and/or `entrypoint`:
https://docs.docker.com/compose/compose-file/#command
https://docs.docker.com/compose/compose-file/#entrypoint

integration-tests/jail/docker-compose.yml
```diff
@@ -20,6 +23,9 @@ services:
     volumes:
       - ./disk:/root/disk
       - ./disk/config1:/root/config
+      - /nix:/nix
+    devices:
+      - /dev/isgx
```
tomtau: the device part isn't portable, as e.g. on azure it's `/dev/sgx`

integration-tests/join/jail.nix
```diff
@@ -0,0 +1,104 @@
+with import <nixpkgs> {};
```
tomtau: isn't this file identical to https://github.com/crypto-com/chain/blob/master/integration-tests/jail/jail.nix ? why not to reuse that / import that?

integration-tests/jail/docker-compose.yml
leejw51: i will modify it

integration-tests/join/jail.nix
leejw51: i'll change it

integration-tests/jail/docker-compose.yml
leejw51: on ubuntu  std installation, it is /dev/isgx,  i think azure is old style, but i will change to azure paradigm.

leejw51: also for docker, it's coded before using docker-compose, i will move it to docker-compose

leejw51: for docker, i'll change it.

leejw51crypto:
rebased

integration-tests/jail/docker-compose.yml
tomtau: you can just use an environment variable (and default it to /dev/isgx)... just like in the existing integration tests @calvinaco wrote -- have you managed to read them?

tomtau: https://github.com/crypto-com/chain/blob/master/integration-tests/docker-compose.yml#L25

tomtau: https://github.com/crypto-com/chain/blob/master/integration-tests/const-env.sh#L41

tomtau: but I think there may be a problem in the existing integration tests that they do `/dev/something:/dev/else` -- for example, if you expose `/dev/isgx:/dev/sgx` (or `/dev/sgx:/dev/isgx`), it may not work

integration-tests/jail/rust.nix
```diff
@@ -0,0 +1,25 @@
+with import <nixpkgs> {};
```
tomtau: again a duplicate file

leejw51crypto: yes, i'll modify

leejw51crypto: also changed  isgx to sgx

PR #875, opened 2020-01-10
Problem (Fix #864): repeating complex set of arguments to chain-abci

### LINKED ISSUE #864: Problem: repeating complex set of arguments to chain-abci
As there's a quite of a few arguments one needs to provide to chain-abci, it may be a bit error-prone when restarting it (as experienced with the tx-query argument).
It'd be desirable to have an option to start chain-abci without providing any arguments or only one argument, so that the rest is read from a file.

Perhaps chain-abci could read from a JSON or TOML file put in `DATA_DIRECTORY/config/...` (where DATA_DIRECTORY is .cro-storage or what's passed in as `-d` argument) which contains a subset of AbciOpt?

linfeng-crypto:
Solution: Add a config file for chain-abci, the default path is  `DATA_PATH/config.yaml`, and you also can set by using `--config to/your/config/path`. The command arguments is the priority.


tomtau:
bors try

linfeng-crypto:
I think in mock mode, the integration test does not set the tx-query address, this is not allowed in this PR, I'll fix it.

tomtau:
> I think in mock mode, the integration test does not set the tx-query address, this is not allowed in this PR, I'll fix it.

it shouldn't be necessary in the current mock mode

tomtau:
mock-validation + moc-enc-dec

linfeng-crypto:
> I think in mock mode, the integration test does not set the tx-query address, this is not allowed in this PR, I'll fix it.

It's not the problem, the client-rpc gives an IO error: 
```
client-rpc-zerofee_1              | 2020-01-14T14:47:40.582294890Z [2020-01-14T14:47:40Z DEBUG sled::pagecache::iterator] failed to read log message at lid 4248 with expected lsn 4248 during iteration: IO error: failed to fill whole buffer
```

I run the integration test on  local host, it works fine, maybe there are some other reasons.

Issue #870, opened 2020-01-09
Problem: enclave code not covered by cargo tools


yihuang:
`cargo fmt`, `cargo clippy`, test coverage, etc. Because they can't be built by cargo directly.

tomtau:
@yihuang this will be the case with the current SDK.
For the moment, the approach could be:
1) for SDK-independent logic, isolate that into SGX-friendly crates (like `enclave-protocol`, `chain-tx-validation`, `chain-core`)  -- that then gets covered automatically
2) for SDK-dependent logic, one can't use the standard cargo tooling like `cargo test` -- instead, it'll need the custom `sgx-test` / `cov` feature-guard, see https://github.com/apache/incubator-teaclave-sgx-sdk/tree/master/samplecode/sgx-cov

yihuang:
> @yihuang this will be the case with the current SDK.
> For the moment, the approach is:
> 
> 1. for SDK-independent logic, isolate that into SGX-friendly crates (like `enclave-protocol`, `chain-tx-validation`, `chain-core`)  -- that then gets covered automatically
> 2. for SDK-dependent logic, one can't use the standard cargo tooling like `cargo test` -- instead, it'll need the custom `sgx-test` / `cov` feature-guard, see https://github.com/apache/incubator-teaclave-sgx-sdk/tree/master/samplecode/sgx-cov

Yeah, `cargo test` is still problematic, but with current approach, `cargo fmt` and `cargo clippy` is runnable on all the codes.

tomtau:
yes, `fmt` and `clippy` could have extra steps in CI pipelines to check the SDK-dependent enclave code

PR #883, opened 2020-01-14
Problem (#870): There are unfixed clippy suggestions on enclave code

yihuang:
Solution:
- Fix existing clippy suggestions.
- Can't add clippy on enclave into travis ci script yet due to lack of sgx sdk

The `cast_ptr_alignment` warning looks suspicious, the correctness dependents on the alignment of `Vec<u8>` pointer, I think normal allocator will produce aligned pointer, so it should be fine normally.

Run on linux:
```
$ cargo clippy -p tx-validation-app -p tx-query-app
$ cargo clippy -p tx-validation-enclave -p tx-query-enclave
```

### DIFF
--- a/chain-tx-enclave/enclave-t-common/src/lib.rs
+++ b/chain-tx-enclave/enclave-t-common/src/lib.rs
@@ -60,6 +60,8 @@ where
             return None;
         }
         let opt = unsafe {
+            // TODO check alignment correctness
+            #[allow(clippy::cast_ptr_alignment)]
             SgxSealedData::<[u8]>::from_raw_sealed_data_t(
                 sealed_log.as_mut_ptr() as *mut sgx_sealed_data_t,
                 sealed_log.len() as u32,
--- a/chain-tx-enclave/tx-query/app/src/enclave_u/mod.rs
+++ b/chain-tx-enclave/tx-query/app/src/enclave_u/mod.rs
@@ -50,7 +50,7 @@ pub extern "C" fn ocall_get_txs(
 ) -> sgx_status_t {
     let req = unsafe { std::slice::from_raw_parts(txids, txids_len as usize) };
 
-    let r = ZMQ_SOCKET.with(|socket| {
+    ZMQ_SOCKET.with(|socket| {
         let send_r = socket.send(req, FLAGS);
         if send_r.is_err() {
             error!("failed to send a request for obtaining sealed data");
@@ -59,19 +59,18 @@ pub extern "C" fn ocall_get_txs(
         if let Ok(msg) = socket.recv_bytes(FLAGS) {
             if msg.len() > (txs_len as usize) {
                 error!("Not enough allocated space to return the sealed tx data");
-                return sgx_status_t::SGX_ERROR_UNEXPECTED;
+                sgx_status_t::SGX_ERROR_UNEXPECTED
             } else {
                 unsafe {
                     std::ptr::copy(msg.as_ptr(), txs, msg.len());
                 }
-                return sgx_status_t::SGX_SUCCESS;
+                sgx_status_t::SGX_SUCCESS
             }
         } else {
             error!("failed to receive a response for obtaining sealed data");
-            return sgx_status_t::SGX_ERROR_UNEXPECTED;
+            sgx_status_t::SGX_ERROR_UNEXPECTED
         }
-    });
-    r
+    })
 }
 
 /// Untrusted function called from the enclave -- sends a ZMQ message to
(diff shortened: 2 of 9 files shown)


chain-tx-enclave/tx-query/app/src/enclave_u/mod.rs
```diff
@@ -197,13 +196,13 @@ fn get_spid() -> sgx_spid_t {
 
 fn decode_hex(hex: &str) -> Vec<u8> {
     let mut r: Vec<u8> = Vec::new();
-    let mut chars = hex.chars().enumerate();
+    let mut chars = hex.as_bytes().iter().copied().enumerate();
```
devashishdxt: What warning does `clippy` show with this?

chain-tx-enclave/tx-validation/enclave/src/lib.rs
```diff
@@ -30,14 +30,17 @@ pub extern "C" fn ecall_initchain(chain_hex_id: u8) -> sgx_status_t {
     }
 }
 
+/// # Safety
+///
+/// This function should not be called with null pointer.
 #[no_mangle]
-pub extern "C" fn ecall_check_tx(
+pub unsafe extern "C" fn ecall_check_tx(
```
tomtau: does clippy suggest moving the unsafe?

chain-tx-enclave/tx-query/app/src/enclave_u/mod.rs
```diff
@@ -197,13 +196,13 @@ fn get_spid() -> sgx_spid_t {
...
     loop {
         let (pos, first) = match chars.next() {
             None => break,
             Some(elt) => elt,
         };
-        if first == ' ' {
+        if first == b' ' {
```
leejw51crypto: b' ' == first

chain-tx-enclave/tx-validation/enclave/src/lib.rs
yihuang: Yeah, clippy suggest `ecall_check_tx` itself should be unsafe.

chain-tx-enclave/tx-query/app/src/enclave_u/mod.rs
yihuang: It shows warning in `decode_hex_digit` for the conversion of char to u8, and since `char` in rust is Unicode, so I just changed them all to u8, which I think would be more correct.

chain-tx-enclave/enclave-t-common/src/lib.rs
```diff
@@ -60,6 +60,7 @@ where
             return None;
         }
         let opt = unsafe {
+            #[allow(clippy::cast_ptr_alignment)]
```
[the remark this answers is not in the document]
yihuang: Done

chain-tx-enclave/tx-query/app/src/enclave_u/mod.rs
yihuang: Done

chain-tx-enclave/tx-validation/enclave/src/lib.rs
tomtau: it'd be preferable to limit the unsafe scope... I'm thinking if `unsafe { slice::from_raw_parts(tx_request, tx_request_len).to_vec() }` would help?

yihuang: > it'd be preferable to limit the unsafe scope... I'm thinking if `unsafe { slice::from_raw_parts(tx_request, tx_request_len).to_vec() }` would help?

clippy says the pointer argument get dereferenced directly, maybe add a null pointer check is helpful.

yihuang:
```
error: this public function dereferences a raw pointer but is not marked `unsafe`
  --> chain-tx-enclave/tx-validation/enclave/src/lib.rs:43:63
   |
43 |     let mut tx_request_slice = unsafe { slice::from_raw_parts(tx_request, tx_request_len) };
   |                                                               ^^^^^^^^^^
   |
   = note: `#[deny(clippy::not_unsafe_ptr_arg_deref)]` on by default
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#not_unsafe_ptr_arg_deref
```

yihuang: I guess it's indeed unsafe to dereference a arbitrary raw pointer.

PR #890, opened 2020-01-15
Problem(Fix #858): complex Bash setup for integration test environment

### LINKED ISSUE #858: Problem: complex Bash setup for integration test environment
to prevent accidental errors, it may be easier if the integration test setup is done in a higher-level language, e.g. utilise the chain-bot setup scripts

yihuang:
- greatly simplfy integration test scripts.
- use chain-bot to run integration tests locally.
- use drone docker pipeline to run integration tests.

Breaking changes:
- change drone secret keys to `SPID` and `IAS_API_KEY` to keep consistency.

Prepare drone node:
```
$ mkdir /tmp/drone-cache
$ # this step is not needed if we push the image to docker hub.
$ docker build -f docker/Dockerfile.new --target TEST -t cryptocom/chain-test:latest .
```

Read `integration-tests/README.md` for instructions to run integration tests locally (only support Linux, mac issue can be solved separately).

> `Dockerfile.new` can be consolidated after @linfeng-crypto 's work.

The relationships of images in `Dockerfile.new`:

```
digraph G {
"ubuntu:18.04" -> RUNTIME_BASE
"baiduxlab/sgx-rust:1804-1.1.0" -> BUILDER_BASE
BUILDER_BASE -> TEST
BUILDER_BASE -> BUILDER
BUILDER -> Default [label="copy --from"]
RUNTIME_BASE -> Default
}
```


tomtau:
bors try

yihuang:
> bors try

Needs to execute these two commands first on drone worker:
```
$ mkdir /tmp/drone-cache
$ # this step is not needed if we push the image to docker hub.
$ docker build -f docker/Dockerfile.new --target TEST -t cryptocom/chain-test:latest .
```

tomtau:
> > bors try
> 
> Needs to execute these two commands first on drone worker:
> 
> ```
> $ mkdir /tmp/drone-cache
> $ # this step is not needed if we push the image to docker hub.
> $ docker build -f docker/Dockerfile.new --target TEST -t cryptocom/chain-test:latest .
> ```

I built it locally and pushing it to Dockerhub now

tomtau:
bors retry

yihuang:
> looks pretty good -- just one worry is about the cache:
> 
> 1. size bloat
> 2. cache-related bugs (e.g. running integration tests against some other version binaries)
> 
> does the cache-rebuild deal with that? should we also include `ttl` (say with one day)?

I guess it wouldn't grow too fast, most of it is just the cargo build target directory. 
Regarding switching between different versions, I think as long as cargo can handle that, it's ok.
It needs to be verified in the real world though.

tomtau:
probably disable the tests on Travis CI -- https://travis-ci.org/crypto-com/chain/jobs/637317416#L455 ?

yihuang:
> probably disable the integration tests on Travis CI -- https://travis-ci.org/crypto-com/chain/jobs/637317416#L455 ?

Is there any needs to keep Travis CI? Can we remove it altogether?

yihuang:
I seems to removed the free-port finding logic, I'll add it back.

tomtau:
> > probably disable the integration tests on Travis CI -- https://travis-ci.org/crypto-com/chain/jobs/637317416#L455 ?
> 
> Is there any needs to keep Travis CI? Can we remove it altogether?

Travis CI can be there for the different rustc version builds, static checks + coverage.

The integration test flow on Travis CI can be disabled or removed

tomtau:
> I missed the free port finding logic, I'll add it back.
I guess that's only needed for the local mode? In Drone mode, it should be isolated?

yihuang:
> > > probably disable the integration tests on Travis CI -- https://travis-ci.org/crypto-com/chain/jobs/637317416#L455 ?
> > 
> > 
> > Is there any needs to keep Travis CI? Can we remove it altogether?
> 
> Travis CI can be there for the different rustc version builds, static checks + coverage.
> 
> The integration test flow on Travis CI can be disabled or removed

Ok, I'll remove the integration test part.
And if there are cache-related bugs, we can set up a remove cache step or even separate pipeline for CI drone, keep the cached pipeline for local running.

tomtau:
I'm now trying the local Drone build -- one other thing I had to change was the device binding (from `/dev/sgx/` to `/dev/isgx`) -- not sure if there's a good solution for that in Drone, but that's a small thing

yihuang:
> I'm now trying the local Drone build -- one other thing I had to change was the device binding (from `/dev/sgx/` to `/dev/isgx`) -- not sure if there's a good solution for that in Drone, but that's a small thing

We fixed this by symlink the device (https://unix.stackexchange.com/questions/445450/how-to-create-a-permanent-symlink-to-a-device).
```
$ cat /etc/udev/rules.d/99_sgx.rules
KERNEL=="isgx", SYMLINK+="sgx"
```

yihuang:
> > I missed the free port finding logic, I'll add it back.
> > I guess that's only needed for the local mode? In Drone mode, it should be isolated?

Oh, yeah, it's isolated now, so no problem. For local mode, the user can set the `BASE_PORT` environment variable.

tomtau:
> > I'm now trying the local Drone build -- one other thing I had to change was the device binding (from `/dev/sgx/` to `/dev/isgx`) -- not sure if there's a good solution for that in Drone, but that's a small thing
> 
> We fixed this by symlink the device (https://unix.stackexchange.com/questions/445450/how-to-create-a-permanent-symlink-to-a-device).
> 
> ```
> $ cat /etc/udev/rules.d/99_sgx.rules
> KERNEL=="isgx", SYMLINK+="sgx"
> ```

yeah, that works as well -- perhaps add it to the README

.travis.yml
```diff
@@ -91,24 +91,6 @@ jobs:
...
-        - cd integration-tests
-        - ./prepare.sh || travis_terminate 1;
-        - . ./env.sh
-        - docker-compose up -d || travis_terminate 1;
-        - ./wait-for-setup.sh || (docker-compose ps; docker-compose logs -t --tail="all"; travis_terminate 1);
-        - ./run-test.sh || (docker-compose ps; docker-compose logs -t --tail="all"; travis_terminate 1);
-        - |
-          cd client-rpc
-          yarn
-          yarn test || (docker-compose ps; docker-compose logs -t --tail="all"; travis_terminate 1);
```
calvinaco: I suggest to keep running integration tests in mock/SW mode on TravisCI to ensure mock/SW mode is working fine.

integration-tests/deps.sh
```diff
@@ -0,0 +1,20 @@
...
+    source $PYTHON_VENV_DIR/bin/activate
+    curl https://bootstrap.pypa.io/get-pip.py | python
+else
+    source $PYTHON_VENV_DIR/bin/activate
+fi
+pip3 install -e .
+pip3 install supervisor
+popd
+
+pushd client-rpc; npm install; popd
```
calvinaco: Out of curiosity, `pushd` and `popd` is to ensure even `npm install` change the directory, it can still go back to the original one?

docker/Dockerfile.new
```diff
@@ -0,0 +1,100 @@
...
+      libgflags-dev \
+      libzmq3-dev \
+      pkg-config \
+      clang; \
+    rm -rf /var/lib/apt/lists/*
+
+COPY --from=tendermint/tendermint:v0.32.9 /usr/bin/tendermint /usr/bin/tendermint
+
+FROM BUILDER_BASE AS TEST
+LABEL maintainer="blockchain@crypto.com"
```
calvinaco: How will this TEST image be used?

calvinaco:
> > looks pretty good -- just one worry is about the cache:
> > 
> > 1. size bloat
> > 2. cache-related bugs (e.g. running integration tests against some other version binaries)
> > 
> > does the cache-rebuild deal with that? should we also include `ttl` (say with one day)?
> 
> I guess it wouldn't grow too fast, most of it is just the cargo build target directory.
> Regarding switching between different versions, I think as long as cargo can handle that, it's ok.
> It needs to be verified in the real world though.
> I don't think there is a `ttl` option here, but we can always delete it manually.

Their docs seem to say there is a `ttl`
https://github.com/Drillster/drone-volume-cache/blob/master/DOCS.md

integration-tests/deps.sh
yihuang: > Out of curiosity, `pushd` and `popd` is to ensure even `npm install` change the directory, it can still go back to the original one?

Yeah, I think so, like a stack of absolute paths.

docker/Dockerfile.new
yihuang: It's used as a base image for integration test in drone.

yihuang:
> > > looks pretty good -- just one worry is about the cache:
> > > 
> > > 1. size bloat
> > > 2. cache-related bugs (e.g. running integration tests against some other version binaries)
> > > 
> > > does the cache-rebuild deal with that? should we also include `ttl` (say with one day)?
> > 
> > 
> > I guess it wouldn't grow too fast, most of it is just the cargo build target directory.
> > Regarding switching between different versions, I think as long as cargo can handle that, it's ok.
> > It needs to be verified in the real world though.
> > I don't think there is a `ttl` option here, but we can always delete it manually.
> 
> Their docs seem to say there is a `ttl`?
> https://github.com/Drillster/drone-volume-cache/blob/master/DOCS.md

Nice, it's not reflected on their website(http://plugins.drone.io/drillster/drone-volume-cache/). Still, I'm not sure what benefits can it bring us. Maybe we can try out to see how fast the cache grows.

.travis.yml
yihuang: We can support the mock mode in another PR, maybe another pipeline in drone?
The advantage of travis is able to run in free cloud?

yihuang:
> Amazing, I found this PR full of things to learn :)
> 
> Back to the PR, it removes the possibility of running integration tests on mock environment, I prefer still letting the TravisCI to run it.
> 
> Another question I have is: in drone, if there are multiple PRs/merges running, how does the integration tests differentiate the ports?

They will run in a different docker container.

.travis.yml
calvinaco: Yes, we can do it in another PR. It is not necessarily TravisCI, can be another pipeline in drone as well.

docker/Dockerfile.new
calvinaco: Ok, but when we build the test image we have to modify this Dockerfile?

calvinaco:
> > Amazing, I found this PR full of things to learn :)
> > Back to the PR, it removes the possibility of running integration tests on mock environment, I prefer still letting the TravisCI to run it.
> > Another question I have is: in drone, if there are multiple PRs/merges running, how does the integration tests differentiate the ports?
> 
> They will run in a different docker container.

Oh yes, I got it.

docker/Dockerfile.new
yihuang: > Ok, but when we build the test image we have to modify this Dockerfile?

No need to modify, just `docker build --target TEST -t ... .`

.travis.yml
yihuang: > Yes, we can do it in another PR. It is not necessarily TravisCI, can be another pipeline in drone as well.

I think since it doesn't need the sgx stuff in mock mode, it might quite easy to set up in Travis too, need to get familiar with the syntax though.

PR #891, opened 2020-01-15
Problem:(cro-819) no C bindings for client libraries

leejw51crypto:
Solution: add c-bindings

### DIFF
--- a/cro-clib/chain.h
+++ b/cro-clib/chain.h
@@ -1,0 +1,44 @@
+/*
+chain c bindings
+*/
+
+#define MAX_LENGTH 1024
+#define MAX_STRING_LENGTH 512
+/* general byte buffer */
+typedef struct {
+    unsigned char buf[MAX_STRING_LENGTH];
+    int length;
+} Buf;
+
+/* general api result */
+typedef struct {
+    int error;
+    Buf value;
+} ApiResult;
+
+/* hdwallet info */
+typedef struct {
+    int error;
+    Buf name;
+    Buf value;
+    Buf mnemonics;
+    Buf viewkey;
+    Buf seed;
+    Buf enckey;
+} HDWallet;
+
+/* initialize state */
+typedef struct {
+    int error;
+    Buf chain_id;
+    Buf server_url;
+    Buf storage_folder;
+} ApiContext;
+
+extern ApiContext initialize(const char* chain_id, const char* server, const char* storage);
+extern ApiResult get_network_id();
+extern HDWallet create_hdwallet(const char* name, const char* passphrase); /* hd wallet */
+extern HDWallet restore_hdwallet(const char* passpharase); 
+extern ApiResult make_hdwallet_staking_address(HDWallet* wallet);
+extern ApiResult make_hdwallet_transfer_address(HDWallet* wallet);
+extern void print_buf(const char* name,Buf* buf); /* for debugging */
--- a/cro-clib/src/lib.rs
+++ b/cro-clib/src/lib.rs
@@ -1,0 +1,245 @@
+pub mod bindings;
+use crate::bindings::{ApiContext, ApiResult, Buf, HDWallet, MAX_STRING_LENGTH};
+use chain_core::init::network::init_chain_id;
+use client_common::error::ErrorKind;
+use client_common::SecKey;
+use client_core::types::WalletKind;
+use client_core::wallet::{DefaultWalletClient, WalletClient};
+use client_core::HDSeed;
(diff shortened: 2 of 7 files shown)


cro-clib/src/bindings.rs
```diff
@@ -0,0 +1,260 @@
+/* automatically generated by rust-bindgen */
```
yihuang: Maybe don't commit this file since it's generated on the fly?

cro-clib/src/lib.rs
```diff
@@ -0,0 +1,245 @@
...
+use client_common::storage::MemoryStorage;
+#[derive(Default)]
+pub struct ApiInfo {
+    pub chain_id: String,
+}
+static mut G_DB: Option<MemoryStorage> = None;
+
+/// # Safety
+/// this function should not be called inside rust
+pub unsafe fn get_string(src: *const ::std::os::raw::c_char) -> String {
```
yihuang: Do we need to make this `pub`?

cro-clib/chain.h
```diff
@@ -0,0 +1,44 @@
...
+
+/* initialize state */
+typedef struct {
+    int error;
+    Buf chain_id;
+    Buf server_url;
+    Buf storage_folder;
+} ApiContext;
+
+extern ApiContext initialize(const char* chain_id, const char* server, const char* storage);
```
yihuang:
```suggestion
extern "C" ApiContext initialize(const char* chain_id, const char* server, const char* storage);
```

cro-clib/chain.h
```diff
@@ -0,0 +1,44 @@
...
+extern ApiResult get_network_id();
+extern HDWallet create_hdwallet(const char* name, const char* passphrase); /* hd wallet */
```
yihuang:
```suggestion
extern HDWallet* create_hdwallet(const char* name, const char* passphrase); /* hd wallet */
```

leejw51crypto: it's intended, not to use heap

cro-clib/chain.h
leejw51crypto: this is auto-matically generated by bindgen, but i'll check

cro-clib/src/lib.rs
leejw51crypto: for now, no, but later , if we use separate files, it's necessary.

cro-clib/src/bindings.rs
leejw51crypto: that's correct, i'll remove it.

cro-clib/chain.h
```diff
@@ -0,0 +1,44 @@
...
+#define MAX_STRING_LENGTH 512
+/* general byte buffer */
+typedef struct {
+    unsigned char buf[MAX_STRING_LENGTH];
+    int length;
+} Buf;
+
+/* general api result */
+typedef struct {
+    int error;
```
yihuang: I suggest following Cardano's style (https://github.com/input-output-hk/rust-cardano/blob/master/cardano-c/cardano.h#L183) to handle the error.

leejw51crypto: ok, i'll check, 
but using heap is difficult to use.

cro-clib/chain.h
yihuang: > it's intended, not to use heap

A `Buf` is 516 bytes, a single `HDWallet` is already 3096 bytes, it's more costly than using the heap.

cro-clib/src/lib.rs
```diff
@@ -0,0 +1,245 @@
...
+        error: 0,
+        value: get_buf("wallet"),
+    }
+}
+
+/// # Safety
+/// this function should not be called inside rust
+pub unsafe fn do_create_hdwallet(
+    name: *const ::std::os::raw::c_char,
+    passphrase: *const ::std::os::raw::c_char,
```
tomtau: `use std::os::raw::c_char`?

cro-clib/chain.h
tomtau: I suggested that yesterday

cro-clib/src/lib.rs
leejw51crypto: c-bindgen doesn't process struct,
output is the same.
i'll change to use heap.

devashishdxt:
> * `bindgen` is for interfacing to an existing C library (C->Rust)  -- for generating C bindings (Rust->C, which is what we want), there's `cbindgen`

I Agree. We should generate C bindings from Rust code.

PR #895, opened 2020-01-15
Problem:(cro-819) no C bindings for client libraries

leejw51crypto:
Solution: add c-bindings
remove auto-generated files

working

tidy up

fix clippy issues

remove chain.h

### DIFF
--- a/chain-core/src/init/network.rs
+++ b/chain-core/src/init/network.rs
@@ -2,6 +2,7 @@ use std::sync::Once;
 static INIT_NETWORK: Once = Once::new();
 static INIT_NETWORK_ID: Once = Once::new();
 
+#[repr(C)]
 #[derive(Debug, Copy, Clone, Eq, PartialEq)]
 pub enum Network {
     Mainnet,
@@ -54,8 +55,9 @@ pub fn get_network_id() -> u8 {
     unsafe { chosen_network::NETWORK_ID }
 }
 
+#[no_mangle]
 /// Returns the chosen network type
-pub fn get_network() -> Network {
+pub extern "C" fn get_network() -> Network {
     unsafe { chosen_network::NETWORK }
 }
 
--- a/cro-clib/.gitignore
+++ b/cro-clib/.gitignore
@@ -1,0 +1,1 @@
+./examples/cro
--- a/cro-clib/Cargo.toml
+++ b/cro-clib/Cargo.toml
@@ -1,0 +1,22 @@
+[package]
+name = "cro-clib"
+version = "0.1.0"
+authors = ["jongwhan lee <jonghwan@crypto.com>"]
+edition = "2018"
+
+[lib]
+name = "cro_clib"
+crate-type =["cdylib"]
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+hex="0.4.0"
+chain-core = { path = "../chain-core" }
+client-common = { path = "../client-common" }
+client-core = { path = "../client-core" }
+secstr = { version = "0.3.2", features = ["serde"] }
+
+[build-dependencies]
+cbindgen = "0.12.2"
+
--- a/cro-clib/chain.h
+++ b/cro-clib/chain.h
@@ -1,0 +1,97 @@
+#include <stdarg.h>
+#include <stdbool.h>
+#include <stdint.h>
+#include <stdlib.h>
+
+#define FAIL -1
+
+#define SUCCESS 0
+
+typedef struct CroAddress CroAddress;
+
+typedef struct CroHDWallet CroHDWallet;
+
+typedef struct CroResult {
+  int result;
+} CroResult;
+
+typedef CroHDWallet *CroHDWalletPtr;
+
+typedef CroAddress *CroAddressPtr;
+
+/**
+ * create hd wallet
+ * minimum  300 byte-length is necessary
+ * # Safety
+ */
+CroResult cro_create_hdwallet(CroHDWalletPtr *wallet_out,
+                              uint8_t *mnemonics,
+                              uint32_t mnemonics_length);
+
+/**
+ * create staking address from bip44 hdwallet
+ * # Safety
+ */
+CroResult cro_create_staking_address(CroHDWalletPtr wallet_ptr,
+                                     Network network,
+                                     CroAddressPtr *address_out,
+                                     uint32_t index);
+
+/**
+ * create utxo address from bip44 wallet, which is for withdrawal, transfer amount
+ * # Safety
+ */
+CroResult cro_create_transfer_address(CroHDWalletPtr wallet_ptr,
+                                      Network network,
+                                      CroAddressPtr *address_out,
+                                      uint32_t index);
+
+/**
+ * create viewkey, which is for encrypted tx
+ * # Safety
+ */
+CroResult cro_create_viewkey(CroHDWalletPtr wallet_ptr,
+                             Network network,
+                             CroAddressPtr *address_out,
+                             uint32_t index);
+
+/**
+ * destroy address
+ * # Safety
+ */
+CroResult cro_destroy_address(CroAddressPtr addr);
+
+/**
+ * destroy bip44 hdwallet
+ * # Safety
+ */
+CroResult cro_destroy_hdwallet(CroHDWalletPtr hdwallet);
+
+/**
+ * print address information
+ * minimum byte length 100 is necessary
+ * # Safety
+ */
+CroResult cro_get_printed_address(CroAddressPtr address_ptr,
+                                  uint8_t *address_output,
+                                  uint32_t address_output_length);
+
+/**
+ * print address information
+ * minimum 32 length is necessary
+ * # Safety
+ */
+CroResult cro_get_raw_address(CroAddressPtr address_ptr,
+                              uint8_t *address_output,
+                              uint32_t *address_output_length);
+
+/**
+ * print address information
+ * # Safety
+ */
+CroResult cro_print_address(CroAddressPtr address_ptr);
+
+/**
+ * # Safety
+ */
+CroResult cro_restore_hdwallet(const char *mnemonics_string, CroHDWalletPtr *wallet_out);
--- a/cro-clib/src/lib.rs
+++ b/cro-clib/src/lib.rs
@@ -1,0 +1,3 @@
+pub mod types;
+pub mod wallet;
+pub use chain_core::init::network::Network;
--- a/cro-clib/src/types.rs
+++ b/cro-clib/src/types.rs
@@ -1,0 +1,52 @@
+use client_common::{PrivateKey, PublicKey};
+use client_core::HDSeed;
+use std::ffi::CStr;
+use std::os::raw::c_char;
+use std::os::raw::c_int;
+pub type CroHDWalletPtr = *mut CroHDWallet;
+pub type CroAddressPtr = *mut CroAddress;
+
+pub const SUCCESS: i32 = 0;
+pub const FAIL: i32 = -1;
+
+/// account types
+#[repr(C)]
+pub enum CroAccount {
+    /// Account for transfer address
+    Transfer = 0,
+    /// Account for staking address
+    Staking = 1,
+    /// Account for viewkey
+    Viewkey = 2,
+}
+
+#[derive(Clone)]
+pub struct CroHDWallet {
+    pub seed: HDSeed,
+}
+
+#[derive(Clone)]
+pub struct CroAddress {
+    pub privatekey: PrivateKey,
+    pub publickey: PublicKey,
+    pub raw: Vec<u8>,
+    pub address: String,
+}
+
+#[repr(C)]
+pub struct CroResult {
+    result: c_int,
+}
+impl CroResult {
+    pub fn success() -> CroResult {
+        CroResult { result: SUCCESS }
+    }
+    pub fn fail() -> CroResult {
+        CroResult { result: FAIL }
+    }
+}
+
+/// # Safety
+pub unsafe fn get_string(src: *const c_char) -> String {
+    CStr::from_ptr(src).to_string_lossy().into_owned()
+}
--- a/cro-clib/src/wallet.rs
+++ b/cro-clib/src/wallet.rs
@@ -1,0 +1,238 @@
+use crate::types::get_string;
+use crate::types::{CroAccount, CroAddress, CroAddressPtr, CroHDWallet, CroHDWalletPtr, CroResult};
+use chain_core::init::address::RedeemAddress;
+use chain_core::init::network::Network;
+use chain_core::state::account::StakedStateAddress;
+use chain_core::tx::data::address::ExtendedAddr;
+use client_common::MultiSigAddress;
+use client_core::{HDSeed, Mnemonic};
+use secstr::SecUtf8;
+use std::ffi::CString;
+use std::os::raw::c_char;
+use std::ptr;
+
+/// create hd wallet
+/// minimum  300 byte-length is necessary
+#[no_mangle]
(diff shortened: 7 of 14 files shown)


leejw51crypto:
chain.h is automatically generated from rust
this is draft for api design.
i will modify for feedbacks.

"so" file is "no mangled" , just like c library

cro-clib/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+
+[dependencies]
+hex="0.4.0"
+chain-core = { path = "../chain-core" }
+client-common = { path = "../client-common" }
+client-core = { path = "../client-core" }
+secstr = { version = "0.3.2", features = ["serde"] }
+
+[build-dependencies]
+bindgen = "0.52.0"
```
tomtau: is there a need for `bindgen`? Only `cbindgen` is used

cro-clib/main.c
```diff
@@ -0,0 +1,25 @@
+#include <stdio.h>
```
tomtau: probably move this to `example` directory inside this crate

cro-clib/src/lib.rs
```diff
@@ -0,0 +1,263 @@
...
+#[derive(Copy, Clone)]
+pub struct ApiResult {
+    pub error: ::std::os::raw::c_int,
+    pub value: Buf,
+}
+
+#[repr(C)]
+#[derive(Copy, Clone)]
+pub struct HDWallet {
+    pub error: ::std::os::raw::c_int,
```
tomtau: as the `c_int` etc. is used in a couple of places, just do `use std::os::raw::c_int`?

cro-clib/Makefile
```diff
@@ -0,0 +1,4 @@
+all:
```
tomtau: move to `example`?

cro-clib/src/lib.rs
```diff
@@ -0,0 +1,263 @@
...
+    let networkid_string = format!("{:2X}", networkid);
+    ApiResult {
+        error: 0,
+        value: get_buf(&networkid_string),
+    }
+}
+
+/// # Safety
+/// this function should not be called inside rust
+pub unsafe extern "C" fn do_create_hdwallet(
```
tomtau: For HD wallet, start with **stateless** functionality -- just like https://github.com/crypto-com/cro-nodelib/pull/14/files#diff-60a0e806fe678c1085ea774b2f44d37eR6

Just expose functions for generating seed / mnemonic and deriving addresses **without a global mutable state**

cro-clib/src/lib.rs
```diff
@@ -0,0 +1,263 @@
...
+#[repr(C)]
+#[derive(Copy, Clone)]
+pub struct HDWallet {
+    pub error: ::std::os::raw::c_int,
+    pub name: Buf,
+    pub value: Buf,
+    pub mnemonics: Buf,
+    pub viewkey: Buf,
+    pub seed: Buf,
+    pub enckey: Buf,
```
tomtau: I'll repeat the same feedback -- these are different types, reflect that in the binding.
For example, for view key, just use `secp256k1::ffi::PublicKey`

cro-clib/src/lib.rs
```diff
@@ -0,0 +1,263 @@
...
+
+fn get_buf_vec(src: &[u8]) -> Buf {
+    let mut ret = Buf {
+        buf: [0; MAX_STRING_LENGTH as usize],
+        length: 0,
+    };
+    ret.buf[..src.len()].clone_from_slice(&src[..]);
+    ret.length = src.len() as i32;
+    ret
+}
```
tomtau: these are very error-prone and dangerous.

Perhaps when different types are used for inputs-outputs, instead of one fixed array.

See the previous feedback:

https://github.com/crypto-com/chain/pull/891#discussion_r366741011
https://github.com/crypto-com/chain/pull/891#discussion_r366738727

Let the caller manage allocation in any way they like, so just accept `*mut ...` + length... if the length is not enough, return an error instead of silently copying only a part of data

cro-clib/src/lib.rs
yihuang: Maybe make some fields opaque(`mnemonics`, `viewkey`, `seed`, `enckey`), expose related functionality through APIs.
What's the meaning of `value` field?

cro-clib/src/lib.rs
leejw51crypto: ok, i got it.
global singleton will be removed.
user will maintain his own context.
chain_wallet_delete
chain_wallet_new

cro-clib/src/lib.rs
tomtau: > What's the meaning of value field?
that's a mystery to me too -- in the code, it only copied the same value as `mnemonics`

cro-clib/Cargo.toml
leejw51crypto: will remove it

cro-clib/main.c
leejw51crypto: yes

cro-clib/src/lib.rs
leejw51crypto: yes

cro-clib/src/lib.rs
leejw51crypto: it's for general result value,
int error;
char* value;
but i'll change it to type for each function

cro-clib/src/lib.rs
leejw51crypto: because main data will be rust object and handled in rust side.
that code will be removed.

cro-clib/Makefile
leejw51crypto: yes

cro-clib/chain.h
```diff
@@ -0,0 +1,65 @@
...
+} HDWalletCreateResult;
+
+/**
+ * # Safety
+ * this function should not be called inside rust
+ */
+int32_t create_hdwallet(const char *_name,
+                        const char *_passphrase,
+                        HDWalletPtr *wallet_out,
+                        HDWalletCreateResult *create_result);
```
yihuang: Why not `char* mnemonic_buf, size_t len`.

cro-clib/chain.h
```diff
@@ -0,0 +1,65 @@
...
+ * # Safety
+ * this function should not be called inside rust
+ */
+int32_t destroy_hdwallet(HDWalletPtr hdwallet);
+
+/**
+ * # Safety
+ * this function should not be called inside rust
+ */
+int32_t restore_hdwallet(const char *_name, const char *_passphrase, HDWalletPtr *wallet_out);
```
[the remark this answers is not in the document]
leejw51crypto: yes, he can provide mnemonic by restore_hdwallet

cro-clib/chain.h
leejw51crypto: ok,i will change it

cro-clib/src/lib.rs
```diff
@@ -0,0 +1,123 @@
...
+
+/// # Safety
+/// this function should not be called inside rust
+#[allow(dead_code)]
+unsafe fn get_string(src: *const c_char) -> String {
+    CStr::from_ptr(src).to_string_lossy().into_owned()
+}
+
+fn copy_string(src: &str, dst: &mut [u8]) {
+    dst[..src.len()].clone_from_slice(&src.as_bytes()[..src.len()]);
```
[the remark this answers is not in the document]
leejw51crypto: ok

cro-clib/src/lib.rs
leejw51crypto: wallet, account will opaque pointer.
which only exists as "struct .." , actually void*

cro-clib/chain.h
tomtau: For `\0`-terminated strings -- use this type: https://doc.rust-lang.org/std/ffi/struct.CString.html

leejw51crypto:
ok

cro-clib/chain.h
leejw51crypto: ok

cro-clib/src/wallet.rs
```diff
@@ -0,0 +1,233 @@
...
+    let address = address_ptr.as_mut().expect("get address");
+    let address_string = address.address.to_string();
+    let mut src_string = address_string.as_bytes().to_vec();
+    src_string.push(0);
+    let src = &src_string[..];
+    assert!(src.len() < buf_length as usize);
+    ptr::copy_nonoverlapping(src.as_ptr(), buf, src.len());
+
+    SUCCESS
+}
```
tomtau: This function is unclear from the C header -- the name is "get_address", so my expectation would be that it copies the raw address data (i.e. 20 bytes in staking addresses, 32 bytes in transfer addresses)... but it's actually a sort of "sprintf".

cro-clib/src/wallet.rs
```diff
@@ -0,0 +1,233 @@
...
+    dst[src.len()] = 0;
+}
+
+/// hdwallet creating using bip44 hdwallet
+#[no_mangle]
+pub unsafe extern "C" fn cro_create_hdwallet(
+    wallet_out: *mut HDWalletPtr,
+    mnemonics: *mut u8,
+    mnemonics_length: i32,
+) -> i32 {
```
tomtau: perhaps you can create a type alias:
```
pub type CroResult = i32;
```

and use that alias instead of i32, so it's then easier to change in the future  (e.g. if it's changed to return a struct with error code + message as in the earlier iteration)

cro-clib/src/wallet.rs
```diff
@@ -0,0 +1,233 @@
...
+    dst[..src.len()].copy_from_slice(&src.as_bytes()[..src.len()]);
+    dst[src.len()] = 0;
+}
+
+/// hdwallet creating using bip44 hdwallet
+#[no_mangle]
+pub unsafe extern "C" fn cro_create_hdwallet(
+    wallet_out: *mut HDWalletPtr,
+    mnemonics: *mut u8,
+    mnemonics_length: i32,
```
tomtau: for length, usize (or u32) instead of i32?

cro-clib/src/wallet.rs
```diff
@@ -0,0 +1,233 @@
...
+    println!("{}", address.address.to_string());
+    SUCCESS
+}
+
+/// print address information
+#[no_mangle]
+pub unsafe extern "C" fn cro_get_address(
+    address_ptr: AddressPtr,
+    buf: *mut u8,
+    buf_length: i32,
```
tomtau: buf is a bit generic name -- perhaps "address_output" ? + length type can be usize/u32 instead of i32

cro-clib/src/wallet.rs
leejw51crypto: ok, i'll change as below

get_printed_address -> print address strng
get_raw_address-> get address as bytes

cro-clib/src/wallet.rs
leejw51crypto: changed to CroResult

cro-clib/src/wallet.rs
leejw51crypto: ok

cro-clib/src/wallet.rs
leejw51crypto: ok

cro-clib/src/types.rs
```diff
@@ -0,0 +1,67 @@
...
+pub const SUCCESS: i32 = 0;
+pub const FAIL: i32 = -1;
+
+/// network kinds
+#[repr(C)]
+pub enum CroNetwork {
+    Mainnet,
+    Testnet,
+    Devnet,
+}
```
tomtau: This is identical to `chain_core::init::network::Network;`

-- perhaps just add `#[repr(C)]` on `chain_core::init::network::Network` and here `pub use chain_core::init::network::Network;`

cro-clib/src/wallet.rs
```diff
@@ -0,0 +1,241 @@
...
+/// # Safety
+pub unsafe extern "C" fn cro_get_printed_address(
+    address_ptr: CroAddressPtr,
+    address_output: *mut u8,
+    address_output_length: u32,
+) -> CroResult {
+    let address = address_ptr.as_mut().expect("get address");
+    let address_string = address.address.to_string();
+    let mut src_string = address_string.as_bytes().to_vec();
+    src_string.push(0);
```
tomtau: use `CString`: https://doc.rust-lang.org/std/ffi/struct.CString.html

cro-clib/.gitignore
```diff
@@ -0,0 +1 @@
+.storage
```
yihuang: Why this line?

cro-clib/chain.h
```diff
@@ -0,0 +1,102 @@
...
+  Devnet,
+} CroNetwork;
+
+typedef struct CroAddress CroAddress;
+
+typedef struct CroHDWallet CroHDWallet;
+
+typedef struct CroResult {
+  int result;
+} CroResult;
```
yihuang:
```suggestion
typedef int CroResult;
```

cro-clib/src/wallet.rs
tomtau: https://github.com/crypto-com/chain/pull/895#discussion_r367745914

cro-clib/src/types.rs
leejw51crypto: ok

cro-clib/src/types.rs
```diff
@@ -0,0 +1,67 @@
...
+
+#[allow(dead_code)]
+/// # Safety
+pub unsafe fn get_string(src: *const c_char) -> String {
+    CStr::from_ptr(src).to_string_lossy().into_owned()
+}
+
+#[allow(dead_code)]
+pub fn copy_string(src: &str, dst: &mut [u8]) {
```
yihuang: It seems no used anywhere.

cro-clib/src/types.rs
```diff
@@ -0,0 +1,67 @@
...
+impl CroResult {
+    pub fn success() -> CroResult {
+        CroResult { result: SUCCESS }
+    }
+    pub fn fail() -> CroResult {
+        CroResult { result: FAIL }
+    }
+}
+
+#[allow(dead_code)]
```
yihuang:
```suggestion
```

cro-clib/src/types.rs
leejw51crypto: yes, i'll delete it

cro-clib/src/wallet.rs
leejw51crypto: CString is owned by Rust, give C can write it,
but above api is writing string inside rust, which is owned in C

cro-clib/.gitignore
leejw51crypto: i'll remove it

cro-clib/src/wallet.rs
tomtau: I mean, you can still do ` ptr::copy_nonoverlapping` -- just that `src` will be `CString` instead of `Vec<u8>`

leejw51crypto: ok, got it

chain-core/src/init/network.rs
```diff
@@ -114,3 +115,10 @@ mod test {
         assert_eq!(1, get_bip44_coin_type_from_network(Network::Devnet));
     }
 }
+
+
+#[no_mangle]
+pub extern "C" fn export_dev_network(net: Network) -> Network  
+{
+    return net;
+}
```
tomtau: oh, this one is because of cbindgen? I guess you can just export a useful function instead -- e.g. `get_bip44_coin_type_from_network` (just add `#[no_mangle]` and `extern` to it)

leejw51crypto: yes, this is for cbindgen exporting.
i'll add useful function

Issue #910, opened 2020-01-16
Problem: unbonding any amount removes a validator


tomtau:
https://github.com/crypto-com/chain/blob/master/chain-abci/src/app/mod.rs#L393
Currently, if a validator removes any amount -- even if the bonded amount is greater than the minimal or the smallest validator -- it'll be removed from the validator set.

PR #913, opened 2020-01-16
Problem: Unbonding any amount removes a validator

devashishdxt:
Solution: Changed new voting power assignment logic. Fixes https://github.com/crypto-com/chain/issues/910

Note: Logic is slightly different than the one in issue description.

### DIFF
--- a/chain-abci/src/app/mod.rs
+++ b/chain-abci/src/app/mod.rs
@@ -387,12 +387,15 @@ impl<T: EnclaveProxy> abci::Application for ChainNodeApp<T> {
                             .get(&account.address)
                             .copied()
                             .unwrap_or_else(TendermintVotePower::zero);
-                        if new_power > old_power && new_power >= min_power {
-                            self.power_changed_in_block
-                                .insert(account.address, new_power);
-                        } else if old_power >= min_power && new_power < old_power {
-                            self.power_changed_in_block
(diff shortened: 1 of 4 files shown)


tomtau:
A few other test cases:
unbond large amount (less than min)
AND
A. deposit the same amount -> no validator update
B. deposit different amounts larger than min -> validator update to that amount
C. deposit amount smaller than min -> validator update to 0

lezzokafka:
> A few other test cases:
> unbond large amount (less than min)
> AND
> A. deposit the same amount -> no validator update
> B. deposit different amounts larger than min -> validator update to that amount
> C. deposit amount smaller than min -> validator update to 0

 Would it works only if the `Unbond` and `Deposit` Tx are in the same block?

tomtau:
> > A few other test cases:
> > unbond large amount (less than min)
> > AND
> > A. deposit the same amount -> no validator update
> > B. deposit different amounts larger than min -> validator update to that amount
> > C. deposit amount smaller than min -> validator update to 0
> 
> Would it works only if the `Unbond` and `Deposit` Tx are in the same block?
> 
> **Edit**: should be all right as validator update is not (and should not) happen often .

yes, that test was in the context of one block

devashishdxt:
Closing this PR. Will reopen once all the testcases are completed.

Issue #827, opened 2020-01-06
Problem: no import/export for non-HD keys


tomtau:
reported by @leejw51crypto 
"for resync, user remove .storage then also wallet information also gone so separate storage for wallet, sync

for reference).storage/wallet <- for wallet.storage/sync <- for sync.cro_storage <- for chain"

yihuang:
`WalletState` and `SyncState` should be stored together, other stuff should be stored together.

yihuang:
We also need a backup for HD wallet, other than the mnemonic words, there are staking address count, transfer address count needs to backup, so when we restore, we have the same wallet.

PR #916, opened 2020-01-16
Problem(Fix #915): no delete wallet API

### LINKED ISSUE #915: Problem: no delete wallet API

yihuang:
Solution:
- Add delete wallet api to both cli and rpc.

### DIFF
--- a/client-cli/src/command/wallet_command.rs
+++ b/client-cli/src/command/wallet_command.rs
@@ -34,6 +34,11 @@ pub enum WalletCommand {
         #[structopt(name = "name", short, long, help = "Name of wallet")]
         name: String,
     },
+    #[structopt(name = "delete", about = "Delete wallet")]
+    Delete {
+        #[structopt(name = "name", short, long, help = "Name of wallet")]
+        name: String,
+    },
 }
 
 impl WalletCommand {
@@ -45,6 +50,7 @@ impl WalletCommand {
             WalletCommand::List => Self::list_wallets(wallet_client),
             WalletCommand::Restore { name } => Self::restore_wallet(wallet_client, name),
             WalletCommand::AuthToken { name } => Self::auth_token(wallet_client, name),
+            WalletCommand::Delete { name } => Self::delete(wallet_client, name),
         }
     }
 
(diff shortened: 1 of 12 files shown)


yihuang:
> Delete wallet is a very important action. Once deleted it would mean the funds are gone forever. This is the reason we didn't implement it in the first place.
> 
> How about deleted wallet is kept for certain period at least.

If the user already saved his mnemonic words safely, then it's not so dangerous.

calvinlauco:
> > Delete wallet is a very important action. Once deleted it would mean the funds are gone forever. This is the reason we didn't implement it in the first place.
> > How about deleted wallet is kept for certain period at least.
> 
> If the user already saved his mnemonic words safely, then it's not so dangerous.

Yes I agree, but this is not an easy assumption we can make 🙈

tomtau:
> > Delete wallet is a very important action. Once deleted it would mean the funds are gone forever. This is the reason we didn't implement it in the first place.
> > How about deleted wallet is kept for certain period at least.
> 
> If the user already saved his mnemonic words safely, then it's not so dangerous.

The default wallet option isn't deterministic, so has no seed and mnemonic words -- perhaps a delete for that type should be disabled until there's a "import/export keys" function?

yihuang:
> can you rebase the changes on the latest master?

done

yihuang:
> > > Delete wallet is a very important action. Once deleted it would mean the funds are gone forever. This is the reason we didn't implement it in the first place.
> > > How about deleted wallet is kept for certain period at least.
> > 
> > 
> > If the user already saved his mnemonic words safely, then it's not so dangerous.
> 
> The default wallet option isn't deterministic, so has no seed and mnemonic words -- perhaps a delete for that type should be disabled until there's a "import/export keys" function?

Maybe open another issue to import/export the basic wallet?

tomtau:
> > > > Delete wallet is a very important action. Once deleted it would mean the funds are gone forever. This is the reason we didn't implement it in the first place.
> > > > How about deleted wallet is kept for certain period at least.
> > > 
> > > 
> > > If the user already saved his mnemonic words safely, then it's not so dangerous.
> > 
> > 
> > The default wallet option isn't deterministic, so has no seed and mnemonic words -- perhaps a delete for that type should be disabled until there's a "import/export keys" function?
> 
> Maybe open another issue to import/export the basic wallet?

https://github.com/crypto-com/chain/issues/827

PR #921, opened 2020-01-17
Problem(fix #847): no high-level client deposit tx workflow

### LINKED ISSUE #847: Problem: no high-level client deposit tx workflow
Currently, when one runs a command `transaction new -t deposit -n ...`, it's quite low-level, as one needs to input individual UTXOs manually which is a bit clumsy (and not easy, as there's no "give me all utxos" command).

It'll be good to have some high-level command that will:
- take staking address and amount
- prepare the "consolidation" transfer tx: this TX will have an output index 0 with the amount required for the deposit tx + fee
- (broadcast that tx and *wait for confirmation*)
- prepare (and broadcast) the deposit tx: just take the output index 0 of that consolidated transfer tx

linfeng-crypto:
solution:
1. create a transfer tx to make a UTXO
2. broadcast the transfer tx and waiting util it confirmed
3. create deposit transaction using the transfer tx output
4. broadcast the deposit tx

### DIFF
--- a/client-cli/src/command/transaction_command.rs
+++ b/client-cli/src/command/transaction_command.rs
@@ -11,7 +11,7 @@ use chain_core::tx::data::attribute::TxAttributes;
 use chain_core::tx::data::input::TxoPointer;
 use chain_core::tx::data::output::TxOut;
 use chain_core::tx::TxAux;
-use client_common::{Error, ErrorKind, PublicKey, Result, ResultExt, SecKey};
+use client_common::{Error, ErrorKind, PublicKey, Result, ResultExt, SecKey, Transaction};
 use client_core::types::TransactionPending;
 use client_core::WalletClient;
 use client_network::NetworkOpsClient;
@@ -25,7 +25,11 @@ use crate::{ask_seckey, coin_from_str};
 #[derive(Debug)]
 pub enum TransactionType {
     Transfer,
+    // deposit inputs in the wallet to a staking address
     Deposit,
+    // deposit any amount of Coins you want to a staking address
+    // it will build an UTXO worth that amount and then deposit to the staking address
+    DepositAmount,
     Unbond,
     Withdraw,
     Unjail,
@@ -40,6 +44,8 @@ impl FromStr for TransactionType {
             Ok(TransactionType::Transfer)
         } else if eq_ascii(s, "deposit") {
             Ok(TransactionType::Deposit)
+        } else if eq_ascii(s, "deposit-amount") {
+            Ok(TransactionType::DepositAmount)
         } else if eq_ascii(s, "unbond") {
             Ok(TransactionType::Unbond)
         } else if eq_ascii(s, "withdraw") {
@@ -98,8 +104,9 @@ impl TransactionCommand {
             } => new_transaction(wallet_client, network_ops_client, name, transaction_type),
             TransactionCommand::Export { name, id } => {
                 let enckey = ask_seckey(None)?;
-                let tx = wallet_client.export_plain_tx(name, &enckey, id)?;
-                success(&tx);
+                let tx_info = wallet_client.export_plain_tx(name, &enckey, id)?;
+                let tx_info_str = tx_info.encode()?;
+                success(&tx_info_str);
                 Ok(())
             }
             TransactionCommand::Import { name, tx } => {
@@ -127,8 +134,11 @@ fn new_transaction<T: WalletClient, N: NetworkOpsClient>(
             wallet_client.update_tx_pending_state(&name, &enckey, tx_aux.tx_id(), tx_pending)?;
         }
         TransactionType::Deposit => {
-            let tx_aux = new_deposit_transaction(network_ops_client, name, &enckey)?;
+            let tx_aux = new_deposit_transaction(wallet_client, network_ops_client, name, &enckey)?;
             wallet_client.broadcast_transaction(&tx_aux)?;
+        }
+        TransactionType::DepositAmount => {
+            new_deposit_amount_transaction(wallet_client, network_ops_client, name, &enckey)?;
         }
         TransactionType::Unbond => {
             let tx_aux = new_unbond_transaction(network_ops_client, name, &enckey)?;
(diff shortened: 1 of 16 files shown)


client-core/src/wallet.rs
```diff
@@ -224,13 +238,19 @@ pub trait WalletClient: Send + Sync {
     /// # Return
     ///
     /// base64 encoded of `Transaction` json string
-    fn export_plain_tx(&self, name: &str, passphras: &SecKey, txid: &str) -> Result<String>;
+    fn export_plain_tx(
+        &self,
+        name: &str,
+        passphras: &SecKey,
+        txid: &str,
+    ) -> Result<TransactionInfo>;
```
yihuang: I guess the doc(even the name) of this method needs to be changed now.

client-rpc/src/rpc/staking_rpc.rs
```diff
@@ -90,37 +92,102 @@ where
...
+    /// 1. build a transfer transaction to make a UTXO which amount is `deposit_amount + fee`
+    /// 2. broadcast the transfer transaction, waiting it confirmed
+    /// 3. use the `outputs[0]` of the transfer transaction to deposit
+    /// 4. broadcast the deposit transaction, return tx_id
     fn deposit_stake(
         &self,
         request: WalletRequest,
         to_address: String,
-        inputs: Vec<TxoPointer>,
+        amount: Coin,
```
yihuang: The old deposit api is still useful in some situation, maybe create a new api?

tomtau:
bors try

client-cli/src/command/transaction_command.rs
```diff
@@ -204,17 +205,89 @@ fn new_unbond_transaction<N: NetworkOpsClient>(
     network_ops_client.create_unbond_stake_transaction(name, enckey, address, value, attributes)
 }
 
-fn new_deposit_transaction<N: NetworkOpsClient>(
+fn new_deposit_transaction<T: WalletClient, N: NetworkOpsClient>(
```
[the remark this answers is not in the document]
DogLi: Ok, I'll fix today

integration-tests/client-rpc/test/network-ops.test.ts
```diff
@@ -163,6 +163,77 @@ describe("Staking", () => {
 		};
 	};
 
+	const testTransferAndStakeAmountBaseUnit = async (
```
calvinaco: I think you forgot to create a test and run this function

linfeng-crypto: you are right, I'll fix it, thanks

tomtau:
bors try

devashishdxt:
> 2. broadcast the transfer tx and waiting util it confirmed

Why do we wait for `TransferTx` to get confirmed. We use `broadcast_tx_sync` to broadcast transaction which returns only when transaction is committed. I don't think there is any need to wait for confirmation. We can directly use `TransferTx` data to create `DepositTx`. Is there any problem with this approach @tomtau?

client-cli/src/command/transaction_command.rs
```diff
@@ -26,6 +28,7 @@ use crate::{ask_seckey, coin_from_str};
 pub enum TransactionType {
     Transfer,
     Deposit,
+    DepositAmount,
```
leejw51crypto: some comment?
explaining deposit, deposit amount is different

tomtau:
> > 1. broadcast the transfer tx and waiting util it confirmed
> 
> Why do we wait for `TransferTx` to get confirmed. We use `broadcast_tx_sync` to broadcast transaction which returns only when transaction is committed. I don't think there is any need to wait for confirmation. We can directly use `TransferTx` data to create `DepositTx`. Is there any problem with this approach @tomtau?

 `broadcast_tx_sync` is waiting for CheckTx result, not for block commit

tomtau:
but a new version may have `broadcast_tx_commit`?

client-cli/src/command/transaction_command.rs
tomtau: yes, that would be good -- especially, it's not really a different transaction type... it's two transactions in a sequence (Transfer => Deposit)

linfeng-crypto: Ok

client-cli/src/command/transaction_command.rs
```diff
@@ -204,17 +213,120 @@ fn new_unbond_transaction<N: NetworkOpsClient>(
...
+            if current_block_height - block_height >= 50 {
+                return Err(Error::new(
+                    ErrorKind::InternalError,
+                    "broadcast transaction timeout",
+                ));
+            }
+            continue;
+        }
+        break;
+    }
```
tomtau: I haven't check if in `Tendermint 0.32`, there's a RPC call `broadcast_tx_commit` that will wait until tx is committed... if there is, it'll be simpler

linfeng-crypto:
> We can directly use `TransferTx` data to create `DepositTx`

if use data directly without waiting the TransferTx comfirmed, the `DepositTx` will fail, I have made a test, tendermint will give an error.

client-core/src/wallet/default_wallet_client.rs
```diff
@@ -99,6 +100,71 @@ where
...
+
+        self.broadcast_transaction(&transaction)?;
+        //update the wallet state
+        let tx_pending = TransactionPending {
+            used_inputs: selected_inputs,
+            block_height: current_block_height,
+            return_amount,
+        };
+
+        self.update_tx_pending_state(name, enckey, transaction.tx_id(), tx_pending)?;
```
tomtau: if that's possible, this could be simplified with `tx_broadcast_commit`

client-rpc/src/rpc/staking_rpc.rs
```diff
@@ -96,31 +107,150 @@ where
...
                 &request.enckey,
-                inputs,
-                addr,
+                transactions,
+                to_staking_address,
                 attr,
             )
             .map_err(to_rpc_error)?;
 
+        // 4. broadcast the deposit transaction and waiting it confirmed
```
tomtau: this functionality is +- duplicate with client-cli -- could this be put to client-core?

client-cli/src/command/transaction_command.rs
devashishdxt: I don't think we should use it even if it is present. Tendermint documentation says that it should only be used for development and testing.

tomtau: ok, I didn't see that documentation

tomtau:
bors try

client-rpc/src/rpc/staking_rpc.rs
linfeng-crypto: Ok

PR #928, opened 2020-01-20
Problem (Fix #777): Order of client list address api is not sorted by creation order.

### LINKED ISSUE #777: (CRO-677) Problem: The result of the list address API is not ordered by creation order
The address list returned by the client should be ordered stable, e.g. by creation order.

Python integration tests use address list API to get the bonded/unbonded stacking address, it’s not stable with current API.

yihuang:
Solution:
- change `BTreeSet` to `IndexSet`

This is required by new integration tests, also more intuitive to user.


client-cli/src/command/address_command.rs
```diff
@@ -139,7 +139,11 @@ impl AddressCommand {
 
         let pub_keys = match address_type {
             AddressType::Staking => wallet_client.staking_keys(name, &enckey)?,
-            AddressType::Transfer => wallet_client.public_keys(name, &enckey)?,
+            AddressType::Transfer => wallet_client
+                .public_keys(name, &enckey)?
+                .iter()
```
devashishdxt: Why not use `.into_iter()` instead of `.iter().cloned()`?

client-core/src/service/wallet_service.rs
```diff
@@ -224,7 +217,7 @@ where
         staking_key: &PublicKey,
     ) -> Result<()> {
         self.modify_wallet(name, enckey, move |wallet| {
-            wallet.staking_keys.insert(staking_key.clone());
+            wallet.staking_keys.push(staking_key.clone());
```
devashishdxt: We now need to check here if given `staking_key` already exists and don't push if it does.

client-core/src/service/wallet_service.rs
```diff
@@ -255,7 +248,7 @@ where
     /// Adds a multi-sig address to given wallet
     pub fn add_root_hash(&self, name: &str, enckey: &SecKey, root_hash: H256) -> Result<()> {
         self.modify_wallet(name, enckey, move |wallet| {
-            wallet.root_hashes.insert(root_hash);
+            wallet.root_hashes.push(root_hash);
```
devashishdxt: We now need to check here if given `root_hash` already exists and don't push if it does.

tomtau:
isn't the order stable with TreeSet? Always sorted?

devashishdxt:
> isn't the order stable with TreeSet? Always sorted?

We can also use [`IndexSet`](https://crates.io/crates/indexmap) if we want to preserve insertion order in set.

client-cli/src/command/address_command.rs
yihuang: Good catch, fixed.

yihuang:
> isn't the order stable with TreeSet? Always sorted?

I need it sorted by creation order, just edited the title.

yihuang:
> > isn't the order stable with TreeSet? Always sorted?
> 
> We can also use [`IndexSet`](https://crates.io/crates/indexmap) if we want to preserve insertion order in set.

`IndexSet` is good as long as we don't need to remove, it seems we won't have removal requirement?

devashishdxt:
> > > isn't the order stable with TreeSet? Always sorted?
> > 
> > 
> > We can also use [`IndexSet`](https://crates.io/crates/indexmap) if we want to preserve insertion order in set.
> 
> `IndexSet` is good as long as we don't need to remove, it seems we won't have removal requirement?

I can only foresee `clear()` (highly improbable that we'll need this) which is ok with `IndexSet`.

yihuang:
> > > > isn't the order stable with TreeSet? Always sorted?
> > > 
> > > 
> > > We can also use [`IndexSet`](https://crates.io/crates/indexmap) if we want to preserve insertion order in set.
> > 
> > 
> > `IndexSet` is good as long as we don't need to remove, it seems we won't have removal requirement?
> 
> I can only foresee `clear()` (highly improbable that we'll need this) which is ok with `IndexSet`.

Done, changed to use `IndexSet`.

client-core/src/service/wallet_service.rs
```diff
@@ -15,16 +14,65 @@ use client_common::{
...
+    pub public_keys: IndexSet<PublicKey>,
     /// public keys of staking addresses
-    pub staking_keys: BTreeSet<PublicKey>,
+    pub staking_keys: IndexSet<PublicKey>,
     /// root hashes of multi-sig transfer addresses
-    pub root_hashes: BTreeSet<H256>,
+    pub root_hashes: IndexSet<H256>,
+}
+
+impl Encode for Wallet {
```
devashishdxt: Can you write some tests for testing `Encode` and `Decode` implementations?

yihuang: It is covered in `check_flow` test case, most of the service APIs involves encoding/decoding of the `Wallet`.

devashishdxt: Ok

PR #933, opened 2020-01-20
Problem: Unbonding any amount removes a validator

devashishdxt:
Solution: Changed new voting power assignment logic. Fixes https://github.com/crypto-com/chain/issues/910

Note: 
- Logic is slightly different than the one in issue description.
- This PR only includes basic test scenarios and does not cover all the possible cases.


tomtau:
bors try

Issue #930, opened 2020-01-20
Problem: unbonded or unjailed validator cannot rejoin the validator set


tomtau:
related https://github.com/crypto-com/chain/issues/910
https://github.com/crypto-com/chain/issues/856

PR #932, opened 2020-01-20
Problem (Fix #853): multi-node integration tests not executed on CI

yihuang:
Solution:
- new multi-node integration tests for jail and join.
- add a multinode-tests step in default drone pipeline.

Known problem:
- The last step of `jail` and `join` test doesn't succeed, because after `unjail` and `joinnode` request, the count of validators don't increase. They are commented out currently, uncomment them after the bug fixed.

PR #955, opened 2020-01-21
Problem: Unbonded or unjailed validator cannot rejoin the validator set

devashishdxt:
Solution: Modified transaction verification to allow existing validators with zero voting power. Fixes https://github.com/crypto-com/chain/issues/930

### DIFF
--- a/chain-tx-validation/src/lib.rs
+++ b/chain-tx-validation/src/lib.rs
@@ -493,15 +493,35 @@ pub fn verify_node_join(
         return Err(Error::NotEnoughStake);
     }
     let validator_address = TendermintValidatorAddress::from(&maintx.node_meta.consensus_pubkey);
-    // checks that validator hasn't joined yet
+
+    // checks that validator hasn't joined yet. If the validator has joined earlier, they can only rejoin if their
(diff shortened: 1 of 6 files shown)


yihuang:
After https://github.com/crypto-com/chain/pull/932 merged, you can uncomment these lines of code:
https://github.com/crypto-com/chain/blob/5b25430c1be007886892643e10f95e0b5b553bd0/integration-tests/multinode/jail_test.py#L79

chain-tx-validation/src/lib.rs
```diff
@@ -493,15 +493,35 @@ pub fn verify_node_join(
...
+
+        // Checks if current voting power is set to zero (i.e., it is not currently a validator)
+        if node_info
+            .validator_voting_power
+            .contains_key(&maintx.address)
+            && *node_info
+                .validator_voting_power
+                .get(&maintx.address)
+                .unwrap()
+                != TendermintVotePower::zero()
```
yihuang: What will happen if unbonding validator to make it bonded coins lower then voting_power_required? Does the power recorded in `validator_voting_power` is zero or non-zero value?

PR #959, opened 2020-01-22
Problem: Ambiguous outputs from client-cli

### LINKED ISSUE #951: Problem: Ambiguous outputs from the `client-cli`
**Describe the bug**
Multiple ambiguous outputs of the commands should be improved to better guide the users.

1. Missing explanations of the input params:
- `./target/debug/client-cli wallet new --name foo --type`
image
It's better to include all the available types and explain each when the type is missing.

- `./target/debug/client-cli state`
image
It's better to further explain the `<name>` as `<wallet name>`

- etc.

2. Duplicate outputs:
- `client-cli balance --name Default`
image
The object of the balance is duplicated with the table liked output

devashishdxt:
Solution: Fixed all the issues described in https://github.com/crypto-com/chain/issues/951. Fixes #951

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@ __pycache__
 
 # Storage
 /.storage
+/.storage.bak
 /.cro-storage
 
 # OS
--- a/client-cli/src/command.rs
+++ b/client-cli/src/command.rs
@@ -66,19 +66,34 @@ pub enum Command {
     },
     #[structopt(name = "view-key", about = "Shows the view key of a wallet")]
     ViewKey {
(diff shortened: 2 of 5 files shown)


yihuang:
The enum values have capitalized initial character before.

devashishdxt:
> The enum values have capitalized initial character before, I think better to keep it backward compatible, or we need to change many integration tests code.

It is case insensitive. I think I also need to add `case_insensitive` option in `structopt` macro.

devashishdxt:
bors try

devashishdxt:
bors try

PR #964, opened 2020-01-23
Problem:(cro-945) no c-binding for normal wallet

leejw51crypto:
Solution: add c-binding for normal wallet
tidy up

### DIFF
--- a/cro-clib/chain.h
+++ b/cro-clib/chain.h
@@ -15,9 +15,48 @@ typedef struct CroResult {
   int result;
 } CroResult;
 
+typedef CroAddress *CroAddressPtr;
+
 typedef CroHDWallet *CroHDWalletPtr;
 
-typedef CroAddress *CroAddressPtr;
+/**
+ * create staking address
+ * # Safety
+ */
+CroResult cro_basic_create_staking_address(CroAddressPtr *address_out);
+
+/**
+ * create staking address
+ * # Safety
+ */
+CroResult cro_basic_create_transfer_address(CroAddressPtr *address_out);
+
+/**
+ * create viewkey, which is for encrypted tx
+ * # Safety
+ */
+CroResult cro_basic_create_viewkey(CroAddressPtr *address_out);
+
+/**
+ * restore staking address
+ * 32 bytes
+ * # Safety
+ */
+CroResult cro_basic_restore_staking_address(CroAddressPtr *address_out, const uint8_t *input);
+
+/**
+ * restore transfer address
+ * 32 bytes
+ * # Safety
+ */
+CroResult cro_basic_restore_transfer_address(CroAddressPtr *address_out, const uint8_t *input);
+
+/**
+ * restore viewkey
+ * 32 bytes
+ * # Safety
+ */
+CroResult cro_basic_restore_viewkey(CroAddressPtr *address_out, const uint8_t *input);
 
 /**
  * create hd wallet
@@ -68,28 +107,29 @@ CroResult cro_destroy_address(CroAddressPtr addr);
 CroResult cro_destroy_hdwallet(CroHDWalletPtr hdwallet);
 
 /**
- * print address information
+ * export privatekey as raw bytes
+ * 32 bytes
+ * # Safety
+ */
+CroResult cro_export_private(CroAddressPtr address_ptr, uint8_t *dst);
+
+/**
+ * extract address as raw bytes
+ * minimum 32 length is necessary
+ * # Safety
+ */
+CroResult cro_extract_raw_address(CroAddressPtr address_ptr,
+                                  uint8_t *address_output,
+                                  uint32_t *address_output_length);
+
+/**
+ * get address as string
  * minimum byte length 100 is necessary
  * # Safety
  */
 CroResult cro_get_printed_address(CroAddressPtr address_ptr,
                                   uint8_t *address_output,
                                   uint32_t address_output_length);
-
-/**
- * print address information
- * minimum 32 length is necessary
- * # Safety
- */
-CroResult cro_get_raw_address(CroAddressPtr address_ptr,
-                              uint8_t *address_output,
-                              uint32_t *address_output_length);
-
-/**
- * print address information
- * # Safety
- */
-CroResult cro_print_address(CroAddressPtr address_ptr);
 
 /**
  * # Safety
--- a/cro-clib/src/address.rs
+++ b/cro-clib/src/address.rs
@@ -1,0 +1,60 @@
+use crate::types::{CroAddressPtr, CroResult};
+use std::ffi::CString;
+use std::ptr;
+
+/// export privatekey as raw bytes
+/// 32 bytes
+#[no_mangle]
+/// # Safety
+pub unsafe extern "C" fn cro_export_private(address_ptr: CroAddressPtr, dst: *mut u8) -> CroResult {
+    let address = address_ptr.as_mut().expect("get address");
+    let src = address.privatekey.serialize();
+    assert!(32 == src.len());
+    ptr::copy_nonoverlapping(src.as_ptr(), dst, src.len());
+    CroResult::success()
+}
+
+/// get address as string
+/// minimum byte length 100 is necessary
+#[no_mangle]
+/// # Safety
+pub unsafe extern "C" fn cro_get_printed_address(
+    address_ptr: CroAddressPtr,
+    address_output: *mut u8,
+    address_output_length: u32,
+) -> CroResult {
+    let address = address_ptr.as_mut().expect("get address");
+    let src_string = CString::new(address.address.as_bytes()).expect("get cstring");
+    let src = src_string.to_bytes_with_nul();
+    if src.len() > address_output_length as usize {
+        return CroResult::fail();
+    }
+    ptr::copy_nonoverlapping(src.as_ptr(), address_output, src.len());
+    CroResult::success()
+}
+
+/// extract address as raw bytes
+/// minimum 32 length is necessary
+#[no_mangle]
+/// # Safety
+pub unsafe extern "C" fn cro_extract_raw_address(
+    address_ptr: CroAddressPtr,
+    address_output: *mut u8,
+    address_output_length: *mut u32,
+) -> CroResult {
+    if address_output.is_null() {
+        return CroResult::fail();
+    }
+    if address_output_length.is_null() {
+        return CroResult::fail();
+    }
+    let address = address_ptr.as_mut().expect("get address");
+    let src_bytes = address.raw.clone();
+    let src = &src_bytes[..];
+    if src.len() > address_output_length as usize {
+        return CroResult::fail();
+    }
+    ptr::copy_nonoverlapping(src.as_ptr(), address_output, src.len());
+    *address_output_length = src.len() as u32;
+    CroResult::success()
+}
--- a/cro-clib/src/basic_wallet.rs
+++ b/cro-clib/src/basic_wallet.rs
@@ -1,0 +1,149 @@
+use crate::types::{CroAddress, CroAddressPtr, CroResult};
+use chain_core::init::address::RedeemAddress;
+
+use chain_core::state::account::StakedStateAddress;
+use chain_core::tx::data::address::ExtendedAddr;
+use client_common::MultiSigAddress;
+use client_common::{PrivateKey, PublicKey};
+use std::ptr;
+
+/// # Safety
+fn do_cro_basic_create_staking_address(private: PrivateKey) -> Box<CroAddress> {
+    let public: PublicKey = PublicKey::from(&private);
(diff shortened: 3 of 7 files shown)


cro-clib/src/basic_wallet.rs
```diff
@@ -0,0 +1,166 @@
...
+use chain_core::init::address::RedeemAddress;
+
+use chain_core::state::account::StakedStateAddress;
+use chain_core::tx::data::address::ExtendedAddr;
+use client_common::MultiSigAddress;
+use client_common::{PrivateKey, PublicKey};
+use std::ptr;
+
+/// # Safety
+unsafe extern "C" fn do_cro_basic_create_staking_address(
```
tomtau: why `do` ? 

perhaps comment should mention that this creates a non-HD from the provided private key?


And does this need to be unsafe if it's no public and only internal function?

cro-clib/src/basic_wallet.rs
```diff
@@ -0,0 +1,166 @@
...
+}
+
+/// restore staking address
+/// input_length: maximum size of input
+#[no_mangle]
+/// # Safety
+pub unsafe extern "C" fn cro_basic_restore_transfer_address(
+    address_out: *mut CroAddressPtr,
+    input: *const u8,
+    input_length: u32,
```
tomtau: does it need length argument if it only can be 32?

cro-clib/src/basic_wallet.rs
```diff
@@ -0,0 +1,166 @@
...
+    assert!(32 == input_length);
+    let array: &[u8] = std::slice::from_raw_parts(input, input_length as usize);
+    match PrivateKey::deserialize_from(array) {
+        Ok(deserialized) => do_cro_basic_create_staking_address(address_out, deserialized),
+        Err(_) => CroResult::fail(),
+    }
+}
+
+/// # Safety
+unsafe extern "C" fn do_cro_basic_create_transfer_address(
```
tomtau: see https://github.com/crypto-com/chain/pull/964/files#r372179264

cro-clib/src/basic_wallet.rs
```diff
@@ -0,0 +1,166 @@
...
+    assert!(32 == input_length);
+    let array: &[u8] = std::slice::from_raw_parts(input, input_length as usize);
+    match PrivateKey::deserialize_from(array) {
+        Ok(deserialized) => do_cro_basic_create_transfer_address(address_out, deserialized),
+        Err(_) => CroResult::fail(),
+    }
+}
+
+/// # Safety
+unsafe extern "C" fn do_cro_basic_create_viewkey(
```
tomtau: see https://github.com/crypto-com/chain/pull/964/files#r372179264

cro-clib/src/address.rs
```diff
@@ -0,0 +1,77 @@
...
+    CroResult::success()
+}
+
+/// print address information
+#[no_mangle]
+/// # Safety
+pub unsafe extern "C" fn cro_export_private(
+    address_ptr: CroAddressPtr,
+    dst: *mut u8,
+    dst_length: *mut u32,
```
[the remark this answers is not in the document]
leejw51: ok

cro-clib/src/basic_wallet.rs
leejw51: it's for handling rust errors

cro-clib/src/basic_wallet.rs
leejw51: i'll remove it , because it's always 32 bytes

cro-clib/src/address.rs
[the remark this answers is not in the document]
leejw51crypto: yes, i'll remove it

cro-clib/src/basic_wallet.rs
leejw51crypto: it's for handling rust errors , because interface function cannot use rust type error handling.

cro-clib/src/basic_wallet.rs
leejw51crypto: i'll remove it

cro-clib/src/basic_wallet.rs
leejw51crypto: ok

cro-clib/src/basic_wallet.rs
leejw51crypto: ok

cro-clib/src/address.rs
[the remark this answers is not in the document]
leejw51crypto: prefix only works in variable names and types, not for functions.

cro-clib/src/address.rs
```diff
@@ -0,0 +1,77 @@
+use crate::types::{CroAddressPtr, CroResult};
+use std::ffi::CString;
+use std::ptr;
+/// print address information
+#[no_mangle]
+/// # Safety
+pub unsafe extern "C" fn cro_print_address(address_ptr: CroAddressPtr) -> CroResult {
```
yihuang: More flexible to return a formatted string then call print in rust.

leejw51crypto: i agree, i'll change it

cro-clib/src/basic_wallet.rs
tomtau: but this function isn't public, it's not exposed -- it could just be a normal Rust function?

leejw51crypto: yes, i agree, i'll change it

leejw51crypto:
ok

cro-clib/src/address.rs
```diff
@@ -0,0 +1,84 @@
...
+    address_ptr: CroAddressPtr,
+    buf: *mut u8,
+    buf_length: u32,
+) -> CroResult {
+    let address = address_ptr.as_mut().expect("get address");
+    let address_string = address.address.to_string();
+    if address_string.as_bytes().len() >= buf_length as usize {
+        return CroResult::fail();
+    }
+    ptr::write_bytes(buf, 0, buf_length as usize);
```
tomtau: the names / comments are a bit misleading -- it doesn't "print" address, it copies the textual/printed address into "buf.... perhaps "get_printed_address" instead of "print_address" and "printed_address_output" instead of "buf" + comment for the expected length?

cro-clib/src/address.rs
```diff
@@ -0,0 +1,84 @@
...
+/// 32 bytes
+#[no_mangle]
+/// # Safety
+pub unsafe extern "C" fn cro_export_private(address_ptr: CroAddressPtr, dst: *mut u8) -> CroResult {
+    let address = address_ptr.as_mut().expect("get address");
+    let src = address.privatekey.serialize();
+    assert!(32 == src.len());
+    ptr::copy_nonoverlapping(src.as_ptr(), dst, src.len());
+    CroResult::success()
+}
```
tomtau: this doesn't "print address information", it copies a serialised private key

cro-clib/src/address.rs
```diff
@@ -0,0 +1,84 @@
...
+
+/// print address information
+/// minimum byte length 100 is necessary
+#[no_mangle]
+/// # Safety
+pub unsafe extern "C" fn cro_get_printed_address(
```
tomtau: what's the difference from https://github.com/crypto-com/chain/pull/964/files#diff-5cae6b4dce11d0a9bef53664a29417dfR7 ? The only difference I see is that this uses "CString"?

cro-clib/src/address.rs
```diff
@@ -0,0 +1,84 @@
...
+#[no_mangle]
+/// # Safety
+pub unsafe extern "C" fn cro_get_raw_address(
+    address_ptr: CroAddressPtr,
+    address_output: *mut u8,
+    address_output_length: *mut u32,
+) -> CroResult {
+    if address_output.is_null() {
+        return CroResult::fail();
+    }
```
tomtau: perhaps not "print", but "extract" (as this is raw, not printable / textual format)?

cro-clib/src/basic_wallet.rs
```diff
@@ -0,0 +1,160 @@
...
+            let raw = redeem.to_vec();
+
+            let ret = CroAddress {
+                privatekey: private,
+                publickey: public,
+                raw,
+                address: address.to_string(),
+            };
+            let address_box = Box::new(ret);
+            ptr::write(address_out, Box::into_raw(address_box));
```
tomtau: this is the only "unsafe" part of this internal function -- perhaps the function could return `CroAddress` and the caller will then do `ptr::write` ?

cro-clib/src/basic_wallet.rs
```diff
@@ -0,0 +1,160 @@
...
+) -> CroResult {
+    let public = PublicKey::from(&private);
+
+    let public_keys = vec![public.clone()];
+    let multi_sig_address = MultiSigAddress::new(public_keys, public.clone(), 1).unwrap();
+
+    let address: ExtendedAddr = multi_sig_address.into();
+
+    match address {
+        ExtendedAddr::OrTree(hash) => {
```
tomtau: same comment as with https://github.com/crypto-com/chain/pull/964/files#diff-b485cdb75580dbce0f2515011b9f1e97R30

cro-clib/src/basic_wallet.rs
```diff
@@ -0,0 +1,160 @@
...
+    }
+}
+
+/// # Safety
+unsafe fn do_cro_basic_create_viewkey(
+    address_out: *mut CroAddressPtr,
+    private: PrivateKey,
+) -> CroResult {
+    let public = PublicKey::from(&private);
+
```
tomtau: same as https://github.com/crypto-com/chain/pull/964/files#diff-b485cdb75580dbce0f2515011b9f1e97R30

cro-clib/src/address.rs
leejw51crypto: i agree, i will change and rebase

cro-clib/src/address.rs
leejw51crypto: ok

cro-clib/src/basic_wallet.rs
leejw51crypto: ok

cro-clib/src/address.rs
leejw51crypto: in c api, user handles privatekey for normal wallet.
unlike hdwallet, normal wallet is not deterministic.
so user manages privatekeys in c side.

cro-clib/src/address.rs
leejw51crypto: removed

leejw51crypto:
rebased

cro-clib/chain.h
```diff
@@ -69,27 +108,28 @@ CroResult cro_destroy_hdwallet(CroHDWalletPtr hdwallet);
 
 /**
  * print address information
```
tomtau: can you update the comments? they all say "print address information"

cro-clib/src/basic_wallet.rs
```diff
@@ -0,0 +1,185 @@
...
+use chain_core::init::address::RedeemAddress;
+
+use chain_core::state::account::StakedStateAddress;
+use chain_core::tx::data::address::ExtendedAddr;
+use client_common::MultiSigAddress;
+use client_common::{PrivateKey, PublicKey};
+use std::ptr;
+
+/// # Safety
+fn do_cro_basic_create_staking_address(private: PrivateKey) -> Result<Box<CroAddress>, ()> {
```
tomtau: can this fail? why `Result`, and not the `Box` directly?

cro-clib/src/basic_wallet.rs
```diff
@@ -0,0 +1,185 @@
...
+    let public: PublicKey = PublicKey::from(&private);
+    let address: StakedStateAddress = StakedStateAddress::BasicRedeem(RedeemAddress::from(&public));
+    match address {
+        StakedStateAddress::BasicRedeem(redeem) => {
+            assert!(20 == redeem.0.len());
```
tomtau: this dynamic assert is redundant -- the type of `redeem.0` is `[u8; REDEEM_ADDRESS_BYTES]`.

If you want to have a piece of mind, you can have a compile-time assertion:
```
const_assert!(REDEEM_ADDRESS_BYTES, 20)
```

cro-clib/chain.h
leejw51crypto: ok

cro-clib/src/basic_wallet.rs
leejw51crypto: ok

cro-clib/src/basic_wallet.rs
leejw51crypto: got it, i'll check

leejw51crypto:
modified

PR #966, opened 2020-01-28
Problem: No specifications for interaction between Validator Set Changes, Punishments and Rewards Pool

devashishdxt:
Solution: Created an ADR to decide on specifications.

Note: Related to https://github.com/crypto-com/chain-docs/issues/58

### DIFF
--- a/architecture-docs/adr-002.md
+++ b/architecture-docs/adr-002.md
@@ -1,0 +1,490 @@
+# ADR 002: Finalize specifications for interaction between Validator Set Changes, Punishments and Rewards
+
+## Changelog
+- 23-01-2020: Initial Draft
+
+## Context
+
+Currently, `ValidatorState` is used by punishments and rewards as a source of data to compute their respective slashing
+ratio and rewards amount. Besides this, `ValidatorState` is also used to manage dynamic validator set (i.e.,
+validators can be added or removed based on their network operations).
+
+### Problems with current design
+
+1. Our current design does not delete validator information from `ValidatorState` when a validator either unbonds their
+   funds or gets removed from validator set because of some punishment. The reason for not deleting this information is
+   because it is still required for computing `SlashRatio` while slashing a punished validator. Because of this,
+   validators cannot rejoin validator set (because their information already exists and we cannot add a duplicate
+   entry).
+1. During rewards calculation, we calculate `total_stake` based of the active validator at the time of rewards
+   distribution. This does not take into account all the validator set changes that took place in current rewards
+   period. For example, when a validator unbonds or deposits funds just before rewards distribution (let us assume that
+   they unbonded their funds at 23rd hour), their stake is not considered in rewards calculation.
+
+Besides this, validator set, rewards pool and punishments are three different components of our system and should
+maintain their own data instead of relying on `ValidatorState` for information. There should be a specified
+interface/behavior for interaction between these components.
+
+## Decision
+
+This document proposes new design for Dynamic Validator Set, Punishments and Rewards Pool. At the end, it specifies the
+interaction between all three components. 
+
+### Dynamic Validator Set
+
+`Validator` struct holds all the information related to a validator/council node.
+
+```rust
+/// A validator in Crypto.com Chain (this data structure can be merged with existing `CouncilNode`)
+pub struct Validator {
+    /// Name of validator
+    pub name: String,
+    /// Staking address of validator's staking account
+    pub staking_address: StakedStateAddress,
+    /// Tendermint's public key of validator
+    pub tendermint_pub_key: TendermintPubKey,
+    /// Current voting power of validator
+    pub voting_power: TendermintVotePower,
+    /// Optional security contact information
+    pub security_contact: Option<String>,
+}
+```
+
+Current validator set is maintained by `ValidatorState` struct. It exposes functions to add, remove and update current
+validator set. Besides this, `ValidatorState` also exposes function to calculate total voting power of current validator
+set which is used to calculate slash ratio while slashing a validator.
+
+```rust
+/// Runtime state of validator set
+pub struct ValidatorState {
+    /// All the validators corresponding to their staking address
+    validators: BTreeMap<StakedStateAddress, Validator>,
+    /// Staking address of validators based on their tendermint validator address (can be used to find validator
+    /// information from tendermint validator address)
+    tendermint_addresses: BTreeMap<TendermintValidatorAddress, StakedStateAddress>,
+}
+
+impl ValidatorState {
+    /// Calculates the total voting power of current active validators (used to calculate slash ratio)
+    pub fn get_total_voting_power(&self) -> TendermintVotePower {
+        todo!()
+    }
+
+    /// Retrieves details of a validator with given staking address (we can filter out inactive validators if required)
+    pub fn get_validator(&self, address: StakedStateAddress) -> Option<&Validator> {
+        todo!()
+    }
+
+    /// Adds a validator to validator set
+    ///
+    /// # Note
+    ///
+    /// - Returns `Err` when a validator with `staking_address` already exists
+    /// - Returns `Err` when validator's `voting_power` is set to zero
+    pub fn add_validator(&mut self, validator: Validator) -> Result<()> {
+        todo!()
+    }
+
+    /// Removes a validator from validator set
+    ///
+    /// # Note
+    ///
+    /// - Returns `Err` when validator with given address does not exist
+    pub fn remove_validator(&mut self, address: StakedStateAddress) -> Result<Validator> {
+        todo!()
+    }
+
+    /// Retrieves a mutable reference to validator details of given staking address (this can be used to modify an
+    /// existing validator's voting power when they unbond or deposit)
+    ///
+    /// # Note
+    ///
+    /// - Removes validator when `new_voting_power` is zero
+    /// - Returns `Err` when validator does not exist
+    pub fn update_voting_power(
+        &mut self,
+        address: &StakedStateAddress,
+        new_voting_power: TendermintVotePower,
+    ) -> Result<()> {
+        todo!()
+    }
+
+    /// Returns a list of active validators in descending order of their voting power
+    ///
+    /// # Note
+    ///
+    /// Sorts the list of validators in descending order of their voting power and returns top `max_validators`.
+    /// Implementation can be optimized so that we don't have to perform sort every time.
+    pub fn get_active_validators(&self) -> Vec<Validator> {
+        todo!()
+    }
+}
+```
+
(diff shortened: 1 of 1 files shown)


architecture-docs/adr-002.md
```diff
@@ -0,0 +1,490 @@
...
+`ProposerTracker` tracks the number of blocks proposed by any validator and also maintains all the validator set changes
+in current reward period.
+
+```rust
+/// Tracks the number of blocks proposed by validators and also their voting power changes
+pub struct ProposerTracker {
+    /// Number of blocks a validator proposed in current reward period
+    proposer_count: BTreeMap<StakedStateAddress, u64>,
+    /// Changes in active validator set in current reward period
+    validator_set_changes: Vec<(Timespec, Vec<Validator>)>,
```
tomtau: what is this one for @lezzokafka @yihuang @devashishdxt ? In https://crypto-com.github.io/getting-started/staking.html#rewards -- it says "number of blocks proposed by the validator / total number of blocks"

architecture-docs/adr-002.md
```diff
@@ -0,0 +1,490 @@
...
+- `UnbondTx`
+  - If the staking address in `UnbondTx` is associated with a validator (i.e., `ValidatorState::get_validator()` returns
+    `Some(_)`), then add the new power of validator (power should be zero if remaining stake is less than minimum
+    required stake) in `power_changed_in_block`.
+- `DepositTx`
+  - If the staking address in `UnbondTx` is associated with a validator (i.e., `ValidatorState::get_validator()` returns
+    `Some(_)`), then add the new power of validator in `power_changed_in_block`.
+
+#### End Block
+
```
tomtau: what about the *interactions*? when should they be removed / cleared?

architecture-docs/adr-002.md
```diff
@@ -0,0 +1,490 @@
...
+### Positive
+
+- This design is more modular and separates concerns of each individual component.
+- Having a modular design makes code more comprehensible and easy to modify in future.
+- Specified interface for interaction between different components makes it the whole design more robust and less prone
+  to bugs.
+
+### Negative
+
+- More code and more abstractions (but I don't thing this should be a concern because Rust compiler should get rid of
```
tomtau: typo

architecture-docs/adr-002.md
devashishdxt: This is to solve second point mentioned in `Problems with current design`. We should consider all the changes in active validator set when calculating total stake.

architecture-docs/adr-002.md
devashishdxt: In the code comments it is mentioned that whenever voting power is updated to zero, the code should remove validator from set. See the comments of `ValidatorState::update_voting_power()`.

devashishdxt:
> the draft so far captures the current behaviour and doesn't fix #930



> the draft so far captures the current behaviour and doesn't fix #930

This ADR also proposes a new design which will automatically fix #930.

architecture-docs/adr-002.md
tomtau: > For example, when a validator unbonds or deposits funds just before rewards distribution (let us assume that they unbonded their funds at 23rd hour), their stake is not considered in rewards calculation.

I guess that's a feature / can be specified this way (if it makes things less complex and faster)

architecture-docs/adr-002.md
tomtau: I think that could be clarified here.
One thing that's implicit or omitted is "how they should be updated".

And what about `LivenessTracker`? When/how is that updated when the validator is no longer active?

architecture-docs/adr-002.md
```diff
@@ -0,0 +1,490 @@
...
+validators can be added or removed based on their network operations).
+
+### Problems with current design
+
+1. Our current design does not delete validator information from `ValidatorState` when a validator either unbonds their
+   funds or gets removed from validator set because of some punishment. The reason for not deleting this information is
+   because it is still required for computing `SlashRatio` while slashing a punished validator. Because of this,
+   validators cannot rejoin validator set (because their information already exists and we cannot add a duplicate
+   entry).
+1. During rewards calculation, we calculate `total_stake` based of the active validator at the time of rewards
```
tomtau: as mentioned in https://github.com/crypto-com/chain/pull/966/files#r372212411 this may be OK -- if I understand it correctly, the only drawback is that in one period, slightly lower amount is emitted, and if one leaves the validator set, they won't get rewarded in other reward period.

architecture-docs/adr-002.md
```diff
@@ -0,0 +1,490 @@
...
+        self.validator_set_changes.push((time, active_validator_set))
+    }
+
+    /// Calculates the sum of average voting powers of each validator (takes voting power changes in account)
+    ///
+    /// # Example
+    ///
+    /// Let us assume that rewards period is 24 hours and there is only one active validator. For the first 23 hours,
+    /// their voting power was 50 and for the last hour, their voting power was 100. Then, the sum average stake should
+    /// be ((23 * 50) + (1 * 100)) / 24. This should be calculated for each validator and their sum should be returned.
```
tomtau: https://crypto-com.github.io/getting-started/staking.html#rewards

> S: total stakings at current block when reward distribution happens

not average

architecture-docs/adr-002.md
devashishdxt: If we want that behaviour, then I guess this isn’t required. But, I think we should add this averaging to calculate total stake, this’ll be more _fair_ for validators. I discussed this with @lezzokafka before leaving, he may be able to explain motivation behind this.

architecture-docs/adr-002.md
devashishdxt: I discussed this with @lezzokafka and we both agreed that we should average the staking power over rewards period.

architecture-docs/adr-002.md
devashishdxt: Yeah. I forgot to add `LivenessTracker` here. Can’t update PR now, I don’t have my laptop with me.

architecture-docs/adr-002.md
devashishdxt: I think we should take validator set changes into account, both unbonding and deposit, and calculate sum of averages.

architecture-docs/adr-002.md
lezzokafka: @devashishdxt  @tomtau  Sure, the current way of calculating the total reward ("minted") being sent to the pool is based on the _daily_reward_rate * total_staking_ at the cut off time.
For example: 
- a huge validator joins the validator set at the very last second; the reward being sent to the pool will be calculated by the _total_staking_ at the cut off time.

architecture-docs/adr-002.md
devashishdxt: Just to clarify, if any validator was removed in current block, i.e., their power was changed to zero, they should be removed from `LivenessTracker`.

architecture-docs/adr-002.md
```diff
@@ -0,0 +1,490 @@
...
+   validators cannot rejoin validator set (because their information already exists and we cannot add a duplicate
+   entry).
+1. During rewards calculation, we calculate `total_stake` based of the active validator at the time of rewards
+   distribution. This does not take into account all the validator set changes that took place in current rewards
+   period. For example, when a validator unbonds or deposits funds just before rewards distribution (let us assume that
+   they unbonded their funds at 23rd hour), their stake is not considered in rewards calculation.
+
+Besides this, validator set, rewards pool and punishments are three different components of our system and should
+maintain their own data instead of relying on `ValidatorState` for information. There should be a specified
+interface/behavior for interaction between these components.
```
yihuang: If we keep several copied of the same data, extra care should be taken to keep them in sync.

architecture-docs/adr-002.md
```diff
@@ -0,0 +1,490 @@
...
+
+```rust
+/// Runtime state of validator set
+pub struct ValidatorState {
+    /// All the validators corresponding to their staking address
+    validators: BTreeMap<StakedStateAddress, Validator>,
+    /// Staking address of validators based on their tendermint validator address (can be used to find validator
+    /// information from tendermint validator address)
+    tendermint_addresses: BTreeMap<TendermintValidatorAddress, StakedStateAddress>,
+}
```
yihuang: How do we calculate validator power updates(`Vec<(ValidatorPubKey, VotingPower)>`) based on this structure?

architecture-docs/adr-002.md
```diff
@@ -0,0 +1,490 @@
...
+/// A validator in Crypto.com Chain (this data structure can be merged with existing `CouncilNode`)
+pub struct Validator {
+    /// Name of validator
+    pub name: String,
+    /// Staking address of validator's staking account
+    pub staking_address: StakedStateAddress,
+    /// Tendermint's public key of validator
+    pub tendermint_pub_key: TendermintPubKey,
+    /// Current voting power of validator
+    pub voting_power: TendermintVotePower,
```
yihuang: Does this field guarantee the stored value either greater than minimal voting power or zero,
Or we do the minimal voting power filtering at query time?

devashishdxt: While writing this ADR, I assumed that whoever implements it will put all the necessary bound checks.

architecture-docs/adr-002.md
devashishdxt: Can you explain a bit more? Why do you want to calculate this information? It is already available to us in `end_block` as `power_changed_in_block`.

architecture-docs/adr-002.md
devashishdxt: If you go through this document, you’ll see that we’re not keeping separate copies of same data, we’re storing data where it is needed and deleting when it is no longer needed.

architecture-docs/adr-002.md
```diff
@@ -0,0 +1,490 @@
...
+- Non-live validators should be added to punishment queue.
+- All the validators in punishment queue should be added to slashing queue (their slash ratio should be updated if
+  they're already present in slashing queue).
+- If current rewards period is over, then rewards should be distributed and proposer tracker should be re-initialized
+  for new rewards period with current active validators.
+
+#### Deliver Transaction
+
+- `NodeJoinTx`
+  - After successful validation, add the validator to `power_changed_in_block`.
```
yihuang: So the `Validator`/`ValidatorState` structure mentioned above is not mutated before `end_block`? All the voting power changes are stored temporarily in `power_changed_in_block`?
I think the details of `power_changed_in_block` should also be described in specification.

devashishdxt: `power_changed_in_block` is present in current implementation and this document does not propose any changes to that. We can add it for reference here.

devashishdxt:
Maybe this is not the best way forward. Closing this PR.

PR #982, opened 2020-02-01
Problem (Fix #954): Reward is lower than expected

### LINKED ISSUE #954: Problem: Reward is lower than expected
**Example:**
`Total_staking: 1.5 Billion` 
`"monetary_expansion_r0\":350`
`"monetary_expansion_tau\":14500000000000000 `
`"distribution_period\":86400` ( One day) 
`"monetary_expansion_decay\":999860`

At the end of the first day, 1688331667524 (16.8 K of CRO ) was "minted" and sent to the reward pool. One of the validator's (with 33% of the voting power) staking address gain from 500,000,000 to 500,005,639 
  
**Expected behaviour:**
The expected size  of the reward distribution should be ~ 1.2 M CRO for that period (about 28% p.a. for 1.5 B CRO staking)

yihuang:
Solution:
- Use coin minting formular provided by @lezzokafka


chain-core/src/common/fixed.rs
```diff
@@ -25,45 +26,196 @@ fn continued_fraction(
...
+    let it_a = vec![FixedNumber::from_num(1), y * 2 - x]
         .into_iter()
-        .chain((2..).map(move |n| x + FixedNumber::from_num(n)));
-    let it_b = iter::once(1).chain((1..).map(|n| -n)).map(move |n| x * n);
+        .chain((0..).map(move |i| y * (i * 4 + 6)));
+    let it_b = iter::once(2 * x).chain(iter::repeat(x2));
     (it_a, it_b)
 }
 
 /// Compute exponencial function with continued fraction method
```
tomtau: perhaps expand a comment to clarify that this is computing `exp(x/y)`

chain-core/src/common/fixed.rs
```diff
@@ -25,45 +26,196 @@ fn continued_fraction(
 }
 
 // https://en.wikipedia.org/wiki/Exponential_function#Continued_fractions_for_ex
-fn exp_series(
+fn exp2_series(
     x: FixedNumber,
+    y: FixedNumber,
 ) -> (
     impl Iterator<Item = FixedNumber>,
     impl Iterator<Item = FixedNumber>,
```
tomtau: as this return type is a bit long, I think it'll be good to at least give it a type alias with some comment how it's meant to be used, e.g.:
```
type FractionSeries = (impl Iterator<Item = FixedNumber>, impl Iterator<Item = FixedNumber>);
```

And my guess is that the usage / implementation is a bit different than the RosettaCode with macros -- so will be good to capture / documentation that (I guess one first needs to remove the head of the first iterator, and then zip it with the second iterator?)

chain-core/src/common/fixed.rs
```diff
@@ -25,45 +26,196 @@ fn continued_fraction(
...
 ) {
-    let it_a = vec![FixedNumber::from_num(1), FixedNumber::from_num(1)]
+    let x2 = x * x;
+    let it_a = vec![FixedNumber::from_num(1), y * 2 - x]
         .into_iter()
-        .chain((2..).map(move |n| x + FixedNumber::from_num(n)));
-    let it_b = iter::once(1).chain((1..).map(|n| -n)).map(move |n| x * n);
+        .chain((0..).map(move |i| y * (i * 4 + 6)));
```
tomtau: I find this a bit difficult to unpack / reason about  -- perhaps this sub-expression `(0..).map(move |i| y * (i * 4 + 6))` could be taken out and given some name?

One other perhaps helpful thing for me would be if a (top-level) comment or a unit test showed the first N expected expressions in each iterator.
is it this?
`it_a = [1, y * 2 - x, 6*y, 10*y, 14*y ....]`
`it_b = [2*x, x^2, x^2 ... ]`
in order to relate that to the wiki article

chain-core/src/common/fixed.rs
```diff
@@ -25,45 +26,196 @@ fn continued_fraction(
...
+fn log2_series(
+    x: FixedNumber,
+    y: FixedNumber,
+) -> (
+    impl Iterator<Item = FixedNumber>,
+    impl Iterator<Item = FixedNumber>,
+) {
+    let n1 = 2 * y + x;
+    let x2 = x * x;
+    let it_a = iter::once(FixedNumber::from_num(0)).chain((1..).step_by(2).map(move |i| n1 * i));
```
tomtau: similar comment as with `exp2`

chain-core/src/common/fixed.rs
```diff
@@ -25,45 +26,196 @@ fn continued_fraction(
...
+    let it_b = iter::once(2 * x).chain((1..).map(move |i| -x2 * (i * i)));
+    (it_a, it_b)
+}
+
+/// Compute natural logarithm function with continued fraction method
+pub fn log2(x: FixedNumber, y: FixedNumber) -> FixedNumber {
+    continued_fraction(log2_series(x, y), 10)
+}
+
+pub fn monetary_expansion(staking: u64, tau: u64, r0: u64, period: u64) -> u64 {
```
tomtau: perhaps some top-level comment / explanation + richer types than `u64` (either type aliases or opaque types) could make it a bit clearer and less-error prone (so that one doesn't pass in wrong arguments in wrong places; or if it's refactored) 

e.g. `staking` / `staking_` could be `Coin`? I guess?
period could be some `Timespec` or some clarification it's in seconds?
`tau` and `r0` could be directly `FixedNumber` or some opaque type that captures their safe ranges -- e.g. have `pub struct Tau` that guarantees range ` [1, 100_00000000_00000000]`, `pub struct R0` ([0, 1000]) ...

tomtau: those opaque types could be used in network parameters as well to make it consistent

tomtau:
bors try

chain-abci/src/app/rewards.rs
```diff
@@ -155,15 +142,20 @@ mod tests {
...
-            total_staking,
-            top_level.rewards_pool.minted,
-            &top_level.network_params,
+            top_level
+                .network_params
+                .get_rewards_monetary_expansion_r0()
+                .as_millis(),
+            top_level.network_params.get_rewards_reward_period_seconds() as u64,
         );
+        println!("reward1: {}", reward1);
```
leejw51crypto: info! , also more explanation

chain-abci/src/app/rewards.rs
```diff
@@ -155,15 +142,20 @@ mod tests {
...
+        let reward1 = Coin::new(reward1).unwrap();
         let mut req = env.req_begin_block(1, 0);
         req.mut_header().set_time(seconds_to_timestamp(
             state.block_time
-                + top_level.network_params.get_rewards_distribution_period() as Timespec,
+                + top_level.network_params.get_rewards_reward_period_seconds() as Timespec,
```
leejw51crypto: since , when some comments?

chain-abci/src/app/rewards.rs
```diff
@@ -182,14 +174,18 @@ mod tests {
...
-            (total_staking + reward1).unwrap(),
-            top_level.rewards_pool.minted,
-            &top_level.network_params,
+            top_level
+                .network_params
+                .get_rewards_monetary_expansion_r0()
+                .as_millis(),
+            top_level.network_params.get_rewards_reward_period_seconds() as u64,
         );
+        let reward2 = Coin::new(reward2).unwrap();
```
leejw51crypto: .expect("get reward2")

chain-core/src/common/fixed.rs
tomtau: The return type could perhaps be `Coin`?

lezzokafka:
Thanks @yihuang  for helping me out with the testing with the config: 
```
      "rewards_config": {
        "monetary_expansion_r0": 350,  
        "monetary_expansion_tau": 10000000000000000, 
       }, 
```
 with applying the new function: 
`math.exp(86400 * math.log(1 + 0.35 * math.exp(-S/10)) / 31536000.0)`
for total staking `S` from 1B to 25B

Left: results from the new function;  Right: expected values

chain-core/src/common/fixed.rs
yihuang: It will complains:
```
rust says `impl Trait` not allowed outside of function and inherent method return types
```

If I only make an alias for `impl Iterator<Item = FixedNumber>`, it still complains:
```
`impl Trait` in type aliases is unstable
note: for more information, see https://github.com/rust-lang/rust/issues/63063
```

tomtau: Something like `struct FractionSeries<T: impl Iterator<Item = FixedNumber>>(pub T, pub T)` would complain about trait objects?

tomtau: if it's wrapped in a struct, I guess one other alternative would be:

```
impl FractionSeries ... {
   pub fn continued_fraction(self, iteration_count: usize) -> FixedNumber {
      ... continued_fraction code...
  }

} 
```

chain-core/src/common/fixed.rs
yihuang: I've changed the staking parameter and return type to `Coin`, `r0` to `Millis`. Inside the computation they are all converted to `Fixed`.
I'm not sure about using `Timespec` for `period`, because I remember not long ago, `Timespec` is still a signed type, maybe better to use `u64` directly here.

chain-core/src/common/fixed.rs
yihuang: Added this comment.

chain-core/src/common/fixed.rs
yihuang: Done

chain-abci/src/app/rewards.rs
```diff
@@ -69,12 +48,20 @@ impl<T: EnclaveProxy> ChainNodeApp<T> {
...
+                total_staking.into(),
+                top_level.rewards_pool.tau,
+                params.get_rewards_monetary_expansion_r0().as_millis(),
+                params.get_rewards_reward_period_seconds() as u64,
+            );
+            Coin::new(min(minted, can_mint)).unwrap()
+        } else {
+            Coin::zero()
+        };
+        println!("minted: {} {}", minted, total_staking);
```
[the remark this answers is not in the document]
yihuang: Done

chain-abci/src/app/rewards.rs
yihuang: Removed this line since it's only for debugging purposes.

chain-abci/src/app/rewards.rs
yihuang: What do you mean?

chain-abci/src/app/rewards.rs
yihuang: Removed now.

chain-core/src/common/fixed.rs
tomtau: we can come back to that in a different PR

chain-core/src/common/fixed.rs
yihuang: And `tau`/`r0`/`period ` has no inherent upper range for this formula, except some too big value might make the computation overflow.

chain-abci/src/app/rewards.rs
leejw51crypto: simple  explanation of "get_rewards_reward_period_seconds"

PR #995, opened 2020-02-04
Problem (Fix #929): total/available balance for ongoing deposit transaction is not correct

### LINKED ISSUE #929: Problem: Ongoing deposit transaction don't have pending value in balance.
Ongoing withdraw and transfer transactions are reflected in `pending` value of balance, but deposit transaction is not.

yihuang:
Solution:
- Handle `TransactionPending` for deposit transaction
- Add thorough integration tests for pending balance


yihuang:
> seems all right, but it'll be good to document this semantics assumptions around total/pending/available (some part suggested in [#929 (comment)](https://github.com/crypto-com/chain/issues/929#issuecomment-581786721) ) -- perhaps adding a few unit tests that show the intended logic?

Added some comments to the `WalletBalance` definition.

I find it hard to write unit testing for the high-level logic because it needs to construct all kinds of tx and execute them, better to do them in integration-tests.

PR #998, opened 2020-02-04
Problem: (fix #967) MultiSig logic is coupled with MultiSig service

### LINKED ISSUE #967: Problem: MultiSig is highly coupled with storage
The high coupling with storage makes it not possible to be exposed to cro-nodelib

calvinlauco:
Solution: Decouple MultiSig into separate module for re-use

### DIFF
--- a/client-core/src/multi_sig/builder.rs
+++ b/client-core/src/multi_sig/builder.rs
@@ -1,0 +1,293 @@
+use parity_scale_codec::{Decode, Encode};
+use secp256k1::schnorrsig::SchnorrSignature;
+
+use chain_core::common::H256;
+use client_common::{ErrorKind, PrivateKey, PublicKey, Result, ResultExt};
+
+use super::MultiSigSession;
+
+/// MultiSig session builder tailored for Crypto.com chain flow
+///
+/// WARNING: Never restore MultiSig builder from an untrusted source or
+/// from a source not created by you.
+/// Restoring from such a source can result in complete lost of your funds.
+/// If you are in doubt, always create a new session builder and restart
+/// the whole process from scratch.
+pub struct MultiSigBuilder {
+    session: MultiSigSession,
+}
+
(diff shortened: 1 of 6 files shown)


client-core/src/multi_sig/builder.rs
```diff
@@ -0,0 +1,277 @@
...
+        let session = MultiSigSession::decode(&mut bytes.as_slice()).chain(|| {
+            (
+                ErrorKind::DeserializationError,
+                "Unable to deserialize MultiSig session",
+            )
+        })?;
+
+        Ok(MultiSigBuilder { session })
+    }
+}
```
tomtau: Is this needed? I don't see it used anywhere.
If it's needed, comments should mention a warning for potential foot guns:

- the underlying multi-party protocol should be aborted if there's a problem detected at any step -- the expected recovery is to start a fresh session
- if there is an API to persist/serialize incomplete sessions, the user should better know to be careful about ensuring the latest session state, otherwise there's a potential "footgun":
  - user starts a session -- incomplete session state 0
  - ...
  - user continues a session -- incomplete session state N
  - an attacker (somehow) manages to force the user to restart the session from some incomplete session state before N
  - e.g. an attacker observed the user's nonce commitment and nonce, tweaked its own nonce and force the user to restart from the state before seeing the attacker's nonce commitment -- a variant of https://medium.com/blockstream/insecure-shortcuts-in-musig-2ad0d38a97da

foreseaz:
LGTM, I will update nodelib to call the decoupled methods after this PR is merged.

client-core/src/multi_sig/builder.rs
calvinaco: Yes it is intended for the `cro-nodelib`, but the scenario is a concern. I think @foreseaz should be aware of too when designing the `cro-nodelib`.

And I will add some warnings to the comment.

PR #1010, opened 2020-02-06
Problem:(cro-947) no C bindings for transaction building

leejw51crypto:
Solution:add c-binding for tx
this pr is for tx building only

### DIFF
--- a/client-core/src/transaction_builder/raw_transfer_transaction_builder.rs
+++ b/client-core/src/transaction_builder/raw_transfer_transaction_builder.rs
@@ -20,10 +20,14 @@ use crate::signer::{DummySigner, SignCondition, Signer};
 use crate::TransactionObfuscation;
 
 /// Unspent transaction output with witness data
-#[derive(Debug, Decode, Encode)]
+#[derive(Debug, Decode, Encode, Clone)]
 pub struct WitnessedUTxO {
+    /// which utxo?
     pub prev_txo_pointer: TxoPointer,
+    /// utxo
     pub prev_tx_out: TxOut,
+    /// signature and merkleproof of rawpubkey
+    /// rawpubkey: combinded key of publickeys for multisig
     pub witness: Option<TxInWitness>,
 }
 
--- a/cro-clib/chain.h
+++ b/cro-clib/chain.h
@@ -9,7 +9,11 @@
 
 typedef struct CroAddress CroAddress;
 
+typedef struct CroFee CroFee;
+
 typedef struct CroHDWallet CroHDWallet;
+
+typedef struct CroTx CroTx;
 
 typedef struct CroResult {
   int result;
@@ -17,7 +21,11 @@ typedef struct CroResult {
 
 typedef CroAddress *CroAddressPtr;
 
+typedef CroFee *CroFeePtr;
+
 typedef CroHDWallet *CroHDWalletPtr;
+
+typedef CroTx *CroTxPtr;
 
 /**
  * create staking address
@@ -59,6 +67,14 @@ CroResult cro_basic_restore_transfer_address(CroAddressPtr *address_out, const u
 CroResult cro_basic_restore_viewkey(CroAddressPtr *address_out, const uint8_t *input);
 
 /**
+ * create fee algorithm
+ * # Safety
+ */
+CroResult cro_create_fee_algorithm(CroFeePtr *fee_out,
+                                   const char *constant_string,
+                                   const char *coeff_string);
+
+/**
  * create hd wallet
  * minimum  300 byte-length is necessary
  * # Safety
@@ -86,6 +102,12 @@ CroResult cro_create_transfer_address(CroHDWalletPtr wallet_ptr,
                                       uint32_t index);
 
 /**
+ * create tx
+ * # Safety
+ */
+CroResult cro_create_tx(CroTxPtr *tx_out, uint8_t network);
+
+/**
  * create viewkey, which is for encrypted tx
  * # Safety
  */
@@ -101,10 +123,29 @@ CroResult cro_create_viewkey(CroHDWalletPtr wallet_ptr,
 CroResult cro_destroy_address(CroAddressPtr addr);
 
 /**
+ * destroy fee
+ * # Safety
+ */
+CroResult cro_destroy_fee_algorithm(CroFeePtr fee);
+
+/**
  * destroy bip44 hdwallet
  * # Safety
  */
 CroResult cro_destroy_hdwallet(CroHDWalletPtr hdwallet);
+
+/**
+ * destroy tx
+ * # Safety
+ */
+CroResult cro_destroy_tx(CroTxPtr tx);
+
+/**
+ * estimate fee
+ * tx_payload_size: in bytes
+ * # Safety
+ */
+uint64_t cro_estimate_fee(CroFeePtr fee_ptr, uint32_t tx_payload_size);
 
 /**
  * export privatekey as raw bytes
@@ -135,3 +176,74 @@ CroResult cro_get_printed_address(CroAddressPtr address_ptr,
  * # Safety
  */
 CroResult cro_restore_hdwallet(const char *mnemonics_string, CroHDWalletPtr *wallet_out);
+
+/**
+ * add txin
+ * txid_string: 64 length hex-char , 32 bytes
+ * addr_string: transfer address
+ * coin: carson unit  for example) 1_0000_0000 carson = 1 cro, 1 carson = 0.0000_0001 cro
+ * # Safety
+ */
+CroResult cro_tx_add_txin(CroTxPtr tx_ptr,
+                          const char *txid_string,
+                          uint16_t txindex,
+                          const char *addr_string,
+                          uint64_t coin);
+
+/**
+ * add txin in bytes
+ * txid: txid in raw bytes, it's 32 bytes
+ * txindex: which utxo in tx which txid_hex points
+ * addr, coin: txid_hex + txindex points this utxo (address, coin value)
+ * # Safety
+ */
+CroResult cro_tx_add_txin_raw(CroTxPtr tx_ptr,
+                              uint8_t txid[32],
+                              uint16_t txindex,
+                              uint8_t addr[32],
+                              uint64_t coin);
+
+/**
+ * add txout , this makes utxo
+ * addr_string: which address in string?
+ * coin: value to send in carson unit , 1 carson= 0.0000_0001 cro
+ * # Safety
+ */
+CroResult cro_tx_add_txout(CroTxPtr tx_ptr, const char *addr_string, uint64_t coin);
+
+/**
+ * add txout with bytes
+ * addr: which address in bytes
+ * coin: value to send in carson unit , 1 carson= 0.0000_0001 cro
+ * # Safety
+ */
+CroResult cro_tx_add_txout_raw(CroTxPtr tx_ptr, uint8_t addr[32], uint64_t coin);
+
+/**
+ * add viewkey in string, which you can get from client-cli
+ * # Safety
+ */
+CroResult cro_tx_add_viewkey(CroTxPtr tx_ptr, const char *viewkey_string);
+
+/**
+ * add viewkey in bytes
+ * # Safety
+ */
+CroResult cro_tx_add_viewkey_raw(CroTxPtr tx_ptr, uint8_t viewkey[33]);
+
+/**
+ * extract bytes from signed tx
+ * this output is encrypted with tx-query-app
+ * can be broadcast to the network
+ * # Safety
+ */
+CroResult cro_tx_complete_signing(CroTxPtr tx_ptr, uint8_t *output, uint32_t *output_length);
+
+/**
+ * sign for each txin
+ * address_ptr: privatekey which will sign
+ * tx_ptr: which tx to sign?
+ * which_tx_in_user: which txin inside tx?
+ * # Safety
+ */
+CroResult cro_tx_sign_txin(CroAddressPtr address_ptr, CroTxPtr tx_ptr, uint16_t which_tx_in_user);
--- a/cro-clib/examples/tx.c
+++ b/cro-clib/examples/tx.c
@@ -1,0 +1,80 @@
+
+#include <stdio.h>
+#include <inttypes.h>
+#include <stdlib.h>
+#include <string.h>
+#include <wchar.h>
+#include <assert.h>
+#include "../chain-core.h"
+#include "../chain.h"
+
+void print_address(CroAddressPtr a);
+void print_hex(char *tmp, int length)
+{
+    int i;
+    for (i = 0; i < length; i++)
+    {
+        printf("%X", (unsigned char)tmp[i]);
+    }
+    printf("\n");
+}
+int test_tx()
+{
+    const char *mnemonics = "annual dinosaur deliver hour loop food buddy lift alert obvious thank scorpion young amused climb defy erode blur drip gun require clerk beef armed";
+
+    CroHDWalletPtr hdwallet = NULL;
+    CroAddressPtr staking = NULL;
+    CroAddressPtr staking2 = NULL;
+    CroAddressPtr transfer = NULL;
+    CroAddressPtr transfer2 = NULL;
+    CroAddressPtr viewkey = NULL;
+    CroAddressPtr viewkey2 = NULL;
+    CroFeePtr fee = NULL;
+    cro_create_fee_algorithm(&fee, "5", "1.0");
+
+    cro_restore_hdwallet(mnemonics, &hdwallet);
+    cro_create_staking_address(hdwallet, Devnet, &staking, 0);
+    cro_create_staking_address(hdwallet, Devnet, &staking2, 1);
+    cro_create_transfer_address(hdwallet, Devnet, &transfer, 0);
+    cro_create_transfer_address(hdwallet, Devnet, &transfer2, 1);
+    cro_create_viewkey(hdwallet, Devnet, &viewkey, 0);
+    cro_create_viewkey(hdwallet, Devnet, &viewkey2, 1);
+    print_address(staking);
+    print_address(staking2);
+    print_address(transfer);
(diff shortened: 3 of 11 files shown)


client-core/src/transaction_builder/raw_transfer_transaction_builder.rs
```diff
@@ -22,8 +22,11 @@ use crate::TransactionObfuscation;
 /// Unspent transaction output with witness data
 #[derive(Debug, Decode, Encode)]
 pub struct WitnessedUTxO {
+    /// which utxo?
     pub prev_txo_pointer: TxoPointer,
+    /// utxo
     pub prev_tx_out: TxOut,
+    /// signature
```
tomtau: not only a signature

cro-clib/src/fee.rs
```diff
@@ -0,0 +1,58 @@
...
+use chain_core::tx::fee::{LinearFee, Milli};
+use std::ffi::CString;
+use std::os::raw::c_char;
+use std::ptr;
+use std::str::FromStr;
+
+/// create fee
+#[no_mangle]
+/// # Safety
+pub unsafe extern "C" fn cro_create_fee(
```
tomtau: this is not creating a "fee", but "fee policy" (something that can be used to compute fees)

cro-clib/src/fee.rs
```diff
@@ -0,0 +1,58 @@
...
+    ptr::write(fee_out, Box::into_raw(fee_box));
+    CroResult::success()
+}
+
+/// estimate fee
+#[no_mangle]
+/// # Safety
+pub unsafe extern "C" fn cro_estimate_fee(
+    fee_ptr: CroFeePtr,
+    length: u32,
```
tomtau: perhaps clarify it's a total tx payload length in bytes

cro-clib/src/fee.rs
tomtau: so better adjust names and comments to make it clear

cro-clib/src/fee.rs
```diff
@@ -0,0 +1,58 @@
...
+    fee_output: *mut u8,
+    fee_output_length: u32,
```
tomtau: Fee is an opaque type around Coin/u64 and has a fixed length.

For C bindings, the equivalent should ideally use C-style opaque types (something like `struct fee { uint64_t x; }`). 

(a more messy approach is to take `*mut u64` directly, but one will lose compile-time checking for APIs that, for example, may take amount + fee where one may by mistake switch the arguments)

cro-clib/src/types.rs
```diff
@@ -46,7 +64,62 @@ impl CroResult {
...
+        }
+    }
+
+    std::str::from_utf8(&src[0..n]).expect("utf8").to_string()
+}
+
+#[repr(C)]
+pub struct CroUtxo {
+    pub address: [u8; 100], // holder
+    pub coin: [u8; 100],    // amount in carson unit
```
tomtau: address should be 32 bytes and amount is u64? what about time locking?

cro-clib/src/types.rs
```diff
@@ -46,7 +64,62 @@ impl CroResult {
...
+    // out amount - in amount = fee
+    // in
+    // if signed, these 3 counts are the same
+    pub txin_pointer: Vec<TxoPointer>,          // which utxo?
+    pub txin_witness: Vec<Option<TxInWitness>>, // sinature
+    pub txin: Vec<TxOut>,                       // utxo which will be consumed
+    // out
+    pub txout: Vec<TxOut>, // new utxo
+    // viewkeys which can view thix tx
+    pub viewkey: Vec<String>,
```
tomtau: this data is identical to what's in `Tx` / `PlainTxAux` -- why not to use that directly?

cro-clib/src/types.rs
```diff
@@ -46,7 +64,62 @@ impl CroResult {
...
+    let mut n = 0;
+    for (i, item) in src.iter().enumerate() {
+        if 0 == (*item) as u8 {
+            n = i;
+            break;
+        }
+    }
+
+    std::str::from_utf8(&src[0..n]).expect("utf8").to_string()
+}
```
tomtau: Why this custom check, and not just use the standard library -- `CString::new(src)` or `CStr::from_bytes_with_nul(src)` ?

cro-clib/src/types.rs
```diff
@@ -34,6 +42,16 @@ pub struct CroAddress {
 }
 
 #[repr(C)]
+#[derive(Clone)]
+pub struct CroStakedState {
+    pub nonce: u64,
+    pub bonded: [u8; 100],
+    pub unbonded: [u8; 100],
```
tomtau: u64? they are amounts

tomtau: and what is the associated address?

cro-clib/examples/tx.c
```diff
@@ -0,0 +1,80 @@
...
+    print_address(staking2);
+    print_address(transfer);
+    print_address(transfer2);
+    print_address(viewkey);
+    print_address(viewkey2);
+
+    CroTxPtr tx = NULL;
+    cro_create_tx(&tx);
+    // compose tx
+    cro_tx_add_txin(tx, "1483dadcae2e6e44bc3248623c7dce749f08270dc9967c6a3ba57f1ce4416b29", 0, "dcro14yj2fc3u3ssvlsfdrr6fu0qmvjx97d405jw9jnwe50ge4y6v53hqr20xzd",
```
tomtau: I think this may be fine -- @calvinaco may comment how he saw it other APIs / bindings.
but it may also be good to add `_raw(...)` functions that, instead of taking these textual representations (hex, bech32, ...), would take directly the byte arrays -- as many of them are 32 bytes (txid, transfer address...), they may be wrapped in some opaque types

cro-clib/src/fee.rs
yihuang: Or all components share a config structure, which is inited with genesis file directly. So no need to copy-paste individuals parameters from genesis file.

cro-clib/chain.h
```diff
@@ -100,12 +120,33 @@ CroResult cro_create_viewkey(CroHDWalletPtr wallet_ptr,
...
+ */
+CroResult cro_destroy_tx(CroTxPtr tx);
+
+/**
+ * estimate fee
+ * # Safety
+ */
+CroResult cro_estimate_fee(CroFeePtr fee_ptr,
+                           uint32_t length,
+                           uint8_t *fee_output,
```
yihuang: What's the content of `fee_output` look like.

client-core/src/transaction_builder/raw_transfer_transaction_builder.rs
leejw51crypto: ok

cro-clib/chain.h
```diff
@@ -135,3 +176,46 @@ CroResult cro_get_printed_address(CroAddressPtr address_ptr,
...
 CroResult cro_restore_hdwallet(const char *mnemonics_string, CroHDWalletPtr *wallet_out);
+
+/**
+ * add txin
+ * txid_string: 64 length hex-char , 32 bytes
+ * addr_string: transfer address
+ * cro_string: cro unit , example 0.0001
+ * # Safety
+ */
+CroResult cro_tx_add_txin(CroTxPtr tx_ptr,
```
yihuang: I guess other transaction types (deposit/unbund/withdraw etc.) are postponed for later pr?

cro-clib/chain.h
leejw51crypto: fee is coin amount, so api gives back as "string",  i think u64 is better.
i will change it

cro-clib/examples/tx.c
leejw51crypto: agree, i will change to raw bytes

cro-clib/src/fee.rs
leejw51crypto: ok

cro-clib/src/fee.rs
leejw51crypto: ok

cro-clib/src/fee.rs
leejw51crypto: it will just return u64, i'll modify

cro-clib/src/types.rs
leejw51crypto: yes, i think "u64" is better.

cro-clib/src/types.rs
leejw51crypto: ok

cro-clib/src/types.rs
leejw51crypto: i'll add time locking

cro-clib/src/types.rs
leejw51crypto: it's for communication, don't need to the same with rust.
i'll check PlainTxAux

cro-clib/chain.h
leejw51crypto: yes, after this pr is reviewed

cro-clib/src/types.rs
leejw51crypto: associated address is staking address

cro-clib/src/types.rs
leejw51crypto: added timelock

cro-clib/src/transaction_build.rs
```diff
@@ -0,0 +1,294 @@
...
+    tx.txin_pointer.push(txin_pointer);
+    tx.txin_witness.push(None);
+    tx.txin.push(txin);
+    Ok(())
+}
+
+/// add txin
+/// txid_string: 64 length hex-char , 32 bytes
+/// addr_string: transfer address
+/// cro_string: cro unit , example 0.0001
```
tomtau: there's no `cro_string` anymore -- so just clarify it's the amount in carsons / base units (1 = 0.0000_0001)

cro-clib/src/types.rs
```diff
@@ -34,6 +42,16 @@ pub struct CroAddress {
 }
 
 #[repr(C)]
+#[derive(Clone)]
+pub struct CroStakedState {
+    pub nonce: u64,
```
tomtau: what about the address, jailed, ...? but I guess this may be replaced/redesigned later

cro-clib/src/types.rs
```diff
@@ -46,7 +64,50 @@ impl CroResult {
...
+    // out amount - in amount = fee
+    // in
+    // if signed, these 3 counts are the same
+    pub txin_pointer: Vec<TxoPointer>,          // which utxo?
+    pub txin_witness: Vec<Option<TxInWitness>>, // sinature
+    pub txin: Vec<TxOut>,                       // utxo which will be consumed
+    // out
+    pub txout: Vec<TxOut>, // new utxo
+    // viewkeys which can view thix tx
+    pub viewkey: Vec<secp256k1::PublicKey>,
```
tomtau: this structure duplicates https://github.com/crypto-com/chain/blob/master/chain-core/src/tx/mod.rs#L60
As the internal structure of `CroTx` is oblivious to C, this could just contain `Tx` and `TxWitness` and mutate these (`(something).tx.add_input(..)`, `(something).tx.add_output(..)`: https://github.com/crypto-com/chain/blob/master/chain-core/src/tx/data/mod.rs#L111 etc.)

cro-clib/src/transaction_build.rs
```diff
@@ -0,0 +1,294 @@
...
+#[no_mangle]
+/// # Safety
+pub unsafe extern "C" fn cro_tx_complete_signing(
+    tx_ptr: CroTxPtr,
+    output: *mut u8,
+    output_length: *mut u32,
+) -> CroResult {
+    let user_tx: &mut CroTx = tx_ptr.as_mut().expect("get tx");
+    let encoded: Vec<u8> = user_tx.tx.as_ref().expect("get tx core").encode();
+    ptr::copy_nonoverlapping(encoded.as_ptr(), output, encoded.len());
```
tomtau: this doesn't extract the signature though -- perhaps construct PlainTxAux and serialise that?

cro-clib/src/transaction_build.rs
```diff
@@ -0,0 +1,294 @@
...
+    }
+}
+
+fn schnorr_sign(
+    message: &TxId,
+    public_key: &PublicKey,
+    private_key: &PrivateKey,
+) -> Result<TxInWitness> {
+    let public_keys: Vec<PublicKey> = vec![public_key.clone()];
+    let multi_sig_address = MultiSigAddress::new(public_keys.to_vec(), public_keys[0].clone(), 1)?;
```
tomtau: this assumes outputs to be locked 1-of-1 -- it's ok to start with, but document this assumption

cro-clib/src/transaction_build.rs
leejw51crypto: ok

cro-clib/src/types.rs
leejw51crypto: yes, i'll add more state.

cro-clib/src/types.rs
leejw51crypto: ok, i'll try to use it directly.

cro-clib/src/transaction_build.rs
leejw51crypto: ok

cro-clib/src/transaction_build.rs
leejw51crypto: yes, multi signature will be added

cro-clib/src/transaction_build.rs
leejw51crypto: ok, i'll change it

cro-clib/src/types.rs
yihuang: > it's for communication, don't need to the same with rust.
> i'll check PlainTxAux

How about expose pointer only?

cro-clib/src/types.rs
```diff
@@ -46,7 +64,43 @@ impl CroResult {
     }
 }
 
+/// copy from c array to rust array
+/// # Safety
+pub unsafe fn copy_string(src: &str, dst: &mut [u8]) {
+    dst[..src.len()].copy_from_slice(&src.as_bytes()[..src.len()]);
+    dst[src.len()] = 0;
+}
+
```
tomtau: this function doesn't seem to be used?

cro-clib/src/types.rs
```diff
@@ -46,7 +64,43 @@ impl CroResult {
...
+#[repr(C)]
+pub struct CroUtxo {
+    pub address: [u8; 32], // holder
+    pub value: u64,        // amount in carson unit
+    pub valid_from: u64,   // time lock
+}
+
+#[derive(Clone, Default)]
+pub struct CroTx {
+    pub txin: Vec<TxOut>,
```
tomtau: what is the purpose of `txin` ? It adds extra information, but I don't see it used anywhere -- only the length is used? Having only `tx.inputs` should be enough?

cro-clib/src/types.rs
```diff
@@ -46,7 +64,43 @@ impl CroResult {
...
+    pub txin_witness: Vec<Option<TxInWitness>>,
+    pub viewkey: Vec<secp256k1::PublicKey>,
```
tomtau: similarly this -- why not just update the attributes in Tx?

cro-clib/src/types.rs
```diff
@@ -46,7 +64,43 @@ impl CroResult {
...
+pub struct CroUtxo {
+    pub address: [u8; 32], // holder
+    pub value: u64,        // amount in carson unit
+    pub valid_from: u64,   // time lock
+}
+
+#[derive(Clone, Default)]
+pub struct CroTx {
+    pub txin: Vec<TxOut>,
+    pub txin_witness: Vec<Option<TxInWitness>>,
```
tomtau: One thing I see in the code is that it expects this vector be initialised with None before signing -- perhaps make fields of `crotx` private and encapsulate these update operations in ```impl CroTx {
pub fn add_txin(&mut self, ...) ...
}```

cro-clib/src/types.rs
leejw51crypto: it's no longer used, i'll remove it.

cro-clib/src/types.rs
leejw51crypto: this is TxoPointer's utxo, supplied from user.

cro-clib/src/types.rs
leejw51crypto: because , it's temporary holder , c-code fills by one by one.

cro-clib/src/types.rs
leejw51crypto: ok

leejw51crypto:
rebased

cro-clib/src/types.rs
tomtau: but why not to hold in in `CroTx.tx` ?

cro-clib/src/types.rs
tomtau: but why not to hold in in `CroTx.tx` ?

leejw51crypto:
use WitnessedUTxO directly and rebased

cro-clib/src/types.rs
leejw51crypto: ok, i'll modify it

leejw51crypto:
add viewkey directly

leejw51crypto:
ok, i'll rebase now

cro-clib/src/types.rs
```diff
@@ -50,3 +67,32 @@ impl CroResult {
...
+    pub valid_from: u64,   // time lock
+}
+
+#[derive(Clone)]
+pub struct CroTx {
+    pub txin: Vec<WitnessedUTxO>, // TxoPointer, TxOut, TxInWitness
+    // contains TxoPointer(in), TxOut(out), attribites(viewkey)
+    pub tx: Tx,
+    /// network-id
+    pub network: u8,
```
tomtau: again this is something duplicating tx attributes

leejw51crypto: i'll modify now.

Issue #851, opened 2020-01-08
Problem: no integration test for unbonding validator


tomtau:
check that it left the validator set

@leejw51crypto Jason can help with this issue

yihuang:
The plan it to integration this into the join test, the test sequence is like this:
- normal three-node setup, with equal bonded coins.
- init validators == 3
- unbond from node0
- check validators == 2
- withdraw and deposit node0
- join node
- check validators == 3

PR #1039, opened 2020-02-10
Problem (Fix #850 #851): no unbonding validator test, no byzantine jail test

### LINKED ISSUE #850: Problem: no automated test for jailing using byzantine faults
two validators at the same time – double sign byzantine fault, one of the validators was “duplicate” (had the same validator key)
3 nodes

### LINKED ISSUE #948: Problem: Client has no stable way to wait for successful execution of transaction
Currently, there are different ways to archive this:
- Wait for several blocks(1 is not enough) (using `latest_block_height` of `status` endpoint), not stable.
- Wait for the pending value of wallet balance to become zero, only works for some types of transactions, and assuming there are no concurrent tasks on the same wallet.
- Query `abci_query` with path `meta` and txid. There's no way to detect failure except timeout. Also not all kinds of transactions(deposit) can be queried with it.

It would be good for client-rpc/client-cli to provide the behaviour of `broadcast_tx_commit`.

yihuang:
Solution:
- integrate unbond test into join test.
- add byzantine test
- use tx_search to wait for transaction (Fix #948)
- change node stopping sequence, try to fix app_hash mismatch problem

### DIFF
--- a/integration-tests/bot/chainbot.py
+++ b/integration-tests/bot/chainbot.py
@@ -129,7 +129,7 @@ def tendermint_cfg(moniker, app_port, rpc_port, p2p_port, peers):
         },
         'tx_index': {
             'indexer': 'kv',
-            'index_tags': '',
+            # 'index_tags': '',
             'index_all_tags': True
         },
         'instrumentation': {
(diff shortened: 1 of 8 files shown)


tomtau:
Two problems = two solutions = two PRs

tomtau:
seems ok, but please separate out

PR #1041, opened 2020-02-10
Problem (Fix #851): no unbonding validator test

### LINKED ISSUE #851: Problem: no integration test for unbonding validator

### LINKED ISSUE #948: Problem: Client has no stable way to wait for successful execution of transaction

yihuang:
Solution:
- integrate unbond test into join test.
- use tx_search to wait for transaction (Fix #948)
- change node stopping sequence, try to fix app_hash mismatch problem

### DIFF
--- a/integration-tests/bot/chainbot.py
+++ b/integration-tests/bot/chainbot.py
@@ -129,7 +129,7 @@ def tendermint_cfg(moniker, app_port, rpc_port, p2p_port, peers):
         },
         'tx_index': {
             'indexer': 'kv',
-            'index_tags': '',
+            # 'index_tags': '',
             'index_all_tags': True
         },
         'instrumentation': {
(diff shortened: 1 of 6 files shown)


tomtau:
bors try

yihuang:
> looks all right -- not sure if #948 is fixed, as it depends on the node configuration.... in the default configuration I guess the client would need to track / return the tendermint tx id

The original motivation for this issue is for integration tests, maybe when we have cases for other kinds of tasks who don't want to enable index tags, open new issues.

PR #1042, opened 2020-02-10
Problem (Fix #1034): node with a jailed staked state remains/returns as an active validator

### LINKED ISSUE #1034: Problem: node with a jailed staked state remains/returns as an active validator
Observed on 0.2.1 testnet:

* eastus2 node validator was restarted at about 1.30 AM on 2020-02-08 and due to a missing /dev/sgx, it failed to start up (tx-validation-app didn't start up -> chain-abci didn't start up -> tendermint didn't start up) ... between blocks 71904 and 71905
* (some time later, the associated staked state was jailed -- not sure exactly when @lezzokafka )
* around 10 AM on 2020-02-08, I fixed the node (reinstalled SGX driver) -- all running and syncing ... at that time, the node didn't appear to be in the validator set (e.g. not in `/validators` or with 0 in `abci_query` of `council-nodes`)
* now, the node is a part of the validator set (address `AECD188A8F9372847665D655F15D2BBDAE1494D0` / pub key: `nUNb5erLD6n61TQ7g6ZlRk067kwzs2tuH8bSro07JlI=`), yet its associated stake state (address: `0x6dbd5b8fe0dad494465aa7574defba711c184102`) is jailed

yihuang:
Solution:
- Don't reward jailed account

### DIFF
--- a/chain-abci/src/app/app_init/validator_state.rs
+++ b/chain-abci/src/app/app_init/validator_state.rs
@@ -337,6 +337,13 @@ impl ValidatorState {
                 let mut state = get_account(addr, &root, &accounts)
                     .expect("io error or validator account not exists");
 
+                if state.is_jailed() {
+                    log::error!(
+                        "Jailed validator should not have reward stats, will cause coins burned"
+                    );
+                    continue;
+                }
+
                 let amount = (share * count).unwrap();
                 let _balance = state.add_reward(amount).unwrap();
                 root = update_staked_state(state.clone(), &root, accounts).0;


chain-abci/src/app/app_init/validator_state.rs
```diff
@@ -337,6 +337,11 @@ impl ValidatorState {
                 let mut state = get_account(addr, &root, &accounts)
                     .expect("io error or validator account not exists");
 
+                if state.is_jailed() {
+                    // don't reward jailed validator
+                    continue;
+                }
+
                 let amount = (share * count).unwrap();
```
tomtau: won't the shares/count then be incorrect?

yihuang: > won't the shares/count then be incorrect?

The share might be smaller than it should be, but if in normal case we will remove reward_stat when jailing validator, this code is only an insurance.

tomtau: log error / warn then if this shouldn't happen?

yihuang: Done

PR #1048, opened 2020-02-11
Problem (Fix #850): no automated test for jailing using byzantine faults

### LINKED ISSUE #850: Problem: no automated test for jailing using byzantine faults

yihuang:
Solution:
- test procedure:
      - 3 nodes
      - node2 has more than 2/3 voting powers
      - node0 and node1 have same validator seed, both have 0 bonded coins initially, only unbonded coins.
      - start the nodes, and stop node1 immediately.
      - withdraw, deposit and join node0.
      - start node1, check node0 punished with byzantine fault.

### DIFF
--- a/integration-tests/multinode/byzantine_test.py
+++ b/integration-tests/multinode/byzantine_test.py
@@ -1,0 +1,67 @@
+#!/usr/bin/env python3
+import os
+import time
+from chainrpc import RPC
+from chainbot import SigningKey
+from common import UnixStreamXMLRPCClient, wait_for_validators, wait_for_port, wait_for_blocks, stop_node, wait_for_tx
+
+'''
+- 3 nodes
+- node2 has more than 2/3 voting powers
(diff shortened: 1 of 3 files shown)


integration-tests/multinode/byzantine_test.py
```diff
@@ -0,0 +1,78 @@
...
+procedure:
+- restore node2 wallet on node0
+- stop node2
+- wait for missed_block_threshold blocks
+- check punishment state on the first staking address
+- start node2
+- wait for node2 port
+- do unjail, expect failure.
+- wait until jail_time + jail_duration
+- do unjail
```
tomtau: this description doesn't seem to match? What is the byzantine fault here?

yihuang: The docstring is copied from jail_test = =".
Updated.

yihuang:
Added the sleep before stop node.

PR #1072, opened 2020-02-14
Problem(fix #865)  no way to run client in air-gapped settings

### LINKED ISSUE #865: Problem: no way to run client in air-gapped settings
https://en.wikipedia.org/wiki/Air_gap_(networking)

e.g. in this setting:

A. Online machine (doesn't have keys)

B. Offline isolated machine (has keys)

we may wish to have the following workflow (using client-cli or equivalents):

1. On A., dump inputs (staked state given an address, network params etc.) to a file F-A

(transfer F-A to B, e.g. on a USB flash drive)

2. On B. read F-A, display it, create a signed transaction using it -- dump the plain signed transaction payload to a file F-B

(transfer F-B to A)

3. On A., (obfuscate if relevant), display and broadcast the transaction created from F-B

linfeng-crypto:
solution:
* on online wallet, create a watch-only wallet, dump the inputs and other needed info as base64 string
* on offline wallet,  create a signed transaction using the dumped info
* send the signed transaction on the online wallet, wallet state be will updated on watch-only wallet

### DIFF
--- a/client-cli/src/command/transaction_command.rs
+++ b/client-cli/src/command/transaction_command.rs
@@ -1,5 +1,8 @@
 use std::collections::BTreeSet;
 use std::convert::TryInto;
+use std::fs::File;
+use std::io::{Read, Write};
+use std::path::PathBuf;
 use std::str::FromStr;
 
 use chain_core::common::{Timespec, HASH_SIZE_256};
@@ -14,6 +17,7 @@ use chain_core::tx::data::input::TxoPointer;
 use chain_core::tx::data::output::TxOut;
 use chain_core::tx::TxAux;
 use client_common::{Error, ErrorKind, PublicKey, Result, ResultExt, SecKey, Transaction};
+use client_core::transaction_builder::SignedTransferTransaction;
 use client_core::types::{BalanceChange, TransactionPending};
 use client_core::WalletClient;
 use client_network::NetworkOpsClient;
@@ -27,6 +31,7 @@ use structopt::StructOpt;
 use unicase::eq_ascii;
 
 use crate::{ask_seckey, coin_from_str};
+use client_core::transaction_builder::UnsignedTransferTransaction;
 
 const TRANSACTION_TYPE_VARIANTS: [&str; 7] = [
     "transfer",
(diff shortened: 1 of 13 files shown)


client-cli/src/command/transaction_command.rs
```diff
@@ -131,6 +133,55 @@ pub enum TransactionCommand {
...
+            short = "n",
+            long = "name",
+            help = "Name of wallet"
+        )]
+        name: String,
+        #[structopt(
+            name = "tx_unsigned",
+            short = "t",
+            long = "tx_unsigned",
+            help = "base64 encoded raw transaction info"
```
leejw51crypto: how about adding option for just json?

client-cli/src/command/transaction_command.rs
```diff
@@ -157,6 +208,34 @@ impl TransactionCommand {
                 success(format!("import amount: {}", imported_amount).as_str());
                 Ok(())
             }
+            TransactionCommand::Build { name } => {
+                let enckey = ask_seckey(None)?;
+                let to_address = ask_transfer_address()?;
+                ask("Enter transfer amount (in CRO): ");
```
leejw51crypto: make sure that user enters  "." in cro

client-cli/src/command/transaction_command.rs
linfeng-crypto: base58 is better I think.

client-cli/src/command/transaction_command.rs
linfeng-crypto: yes, the function [`coin_from_str`](https://github.com/crypto-com/chain/blob/master/client-cli/src/main.rs#L75) can deal with that I think.

client-cli/src/command/transaction_command.rs
tomtau: base58 was meant for potential human input -- this is more for copy-paste.
base64 is ok for that.

another option is just to dump to a binary file and provide a path to that file

client-rpc/src/rpc/wallet_rpc.rs
```diff
@@ -80,6 +82,25 @@ pub trait WalletRpc: Send + Sync {
...
+        to_address: String,
+        amount: Coin,
+        view_keys: Vec<String>,
+    ) -> Result<String>;
+
+    #[rpc(name = "wallet_signRawTransferTransaction")]
+    fn sign_raw_transfer_tx(&self, request: WalletRequest, unsigned_tx: String) -> Result<String>;
+
+    #[rpc(name = "wallet_broadcastSignedTransferTransaction")]
+    fn broadcast_signed_transfer_tx(
```
yihuang: Is the result of `build_raw_transfer_tx` can be sent to tendermint broadcast directly? I think no need this api in `client-rpc`?

linfeng-crypto: no, I'll remove it, just for test

linfeng-crypto:
> `broadcast_signed_transfer_tx` seems no necessary if the raw tx can be sent to tendermint for broadcast directly?

I think we need it, it a hight-level method and update watch-only wallet's status such as update the balance, it looks like that we can send coin from the watch-only wallet just like the normal wallet, but we signed on the offline machine.

yihuang:
> > `broadcast_signed_transfer_tx` seems no necessary if the raw tx can be sent to tendermint for broadcast directly?
> 
> I think we need it, it a hight-level method and update watch-only wallet's status such as update the balance, it looks like that we can send coin from the watch-only wallet just like the normal wallet, but we signed on the offline machine.

👌

client-rpc/src/rpc/wallet_rpc.rs
tomtau: It does an extra step of storing / tracking the tx before broadcast: https://github.com/crypto-com/chain/pull/1072/files#diff-56d5986577a1416a78d11832b08af3c0R865 but not sure if necessary?

linfeng-crypto:
> would it be possible to add a test of this functionality, e.g. in integration tests?

I've talked about the integration tests  with @leejw51crypto before, I have to prepare two environments, online and offline, which is different than the current one(only need one environment), he suggested that we do it in a new issue.

Actually, for the offline environment, we only need `client-cli`, but `client-cli` can not start up if there is no tendermint(and `chain-abci` for it), maybe we can add `--offline` to skip the tendermint check when the `client-cli` start, then the environment for the integration test would be same with current one and can do easily.

tomtau:
and with `chain-rpc`?

linfeng-crypto:
> and with `chain-rpc`?

I can do in client-clie

PR #1035, opened 2020-02-10
Problem: Public network state retrievals require wallet information.

devashishdxt:
Solution: Removed wallet details requirement. Fixes #974.

PR #1074, opened 2020-02-14
Problem(fix #1043) watch-only mode cannot be used in client-cli

### LINKED ISSUE #1043: Problem: watch-only mode cannot be used in client-cli
https://github.com/crypto-com/chain/pull/698
may be useful for https://github.com/crypto-com/chain/issues/865

linfeng-crypto:
solution: add `restore basic wallet ` and `create watch address` in client-cli

### DIFF
--- a/client-cli/src/command/address_command.rs
+++ b/client-cli/src/command/address_command.rs
@@ -1,10 +1,10 @@
 use std::str::FromStr;
 
-use quest::{ask, success};
+use quest::{ask, success, text};
 use structopt::StructOpt;
 use unicase::eq_ascii;
 
-use client_common::{Error, ErrorKind, Result};
+use client_common::{error::ResultExt, Error, ErrorKind, PublicKey, Result};
 use client_core::WalletClient;
 
 use crate::ask_seckey;
@@ -14,6 +14,7 @@ const ADDRESS_TYPE_VARIANTS: [&str; 2] = ["transfer", "staking"];
 #[derive(Debug)]
 pub enum AddressType {
     Transfer,
+    TransferWatch,
     Staking,
 }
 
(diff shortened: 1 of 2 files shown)


client-cli/src/command/address_command.rs
```diff
@@ -123,11 +129,26 @@ impl AddressCommand {
                 success(&format!("New address: {}", address));
                 Ok(())
             }
+            AddressType::StakingWatch => {
+                let enckey = ask_seckey(None)?;
+                let public_key = ask_public_key(None)?;
```
yihuang: public key is not secret stuff, maybe pass it in the argument?

linfeng-crypto: sure

linfeng-crypto: But when creating a normal staking address or transfer address, no need to pass the pubkey, if we pass the pubkey in the arguments, it may make some confusion.

yihuang: It's more like import an watch-only address, rather than create an address?

lezzokafka: The staking address and its state is public, anyone with the staking address is able to check its state.

client-cli/src/command/address_command.rs
```diff
@@ -14,7 +14,9 @@ const ADDRESS_TYPE_VARIANTS: [&str; 2] = ["transfer", "staking"];
 #[derive(Debug)]
 pub enum AddressType {
     Transfer,
+    TransferWatch,
     Staking,
+    StakingWatch,
```
tomtau: I think no need for this, as it's public information -- just use "state" command to retrieve it: https://github.com/crypto-com/chain/pull/1035

linfeng-crypto: Ok

PR #1080, opened 2020-02-17
Problem:(cro-947) no C bindings for transaction building

leejw51crypto:
Solution: added unbond, withdrawal, transfer, deposit c-bindings

### DIFF
--- a/cro-clib/chain.h
+++ b/cro-clib/chain.h
@@ -9,6 +9,8 @@
 
 typedef struct CroAddress CroAddress;
 
+typedef struct CroDepositTx CroDepositTx;
+
 typedef struct CroFee CroFee;
 
 typedef struct CroHDWallet CroHDWallet;
@@ -26,6 +28,18 @@ typedef CroFee *CroFeePtr;
 typedef CroHDWallet *CroHDWalletPtr;
 
 typedef CroTx *CroTxPtr;
+
+typedef CroDepositTx *CroDepositTxPtr;
+
+/**
+ * TODO: other states (jailed, unjail) will be added
+ */
+typedef struct CroStakedState {
+  uint64_t nonce;
+  uint64_t bonded;
+  uint64_t unbonded;
+  uint64_t unbonded_from;
+} CroStakedState;
 
 /**
  * create staking address
@@ -67,6 +81,16 @@ CroResult cro_basic_restore_transfer_address(CroAddressPtr *address_out, const u
 CroResult cro_basic_restore_viewkey(CroAddressPtr *address_out, const uint8_t *input);
 
 /**
+ * staked -> utxo
+ * tendermint_url: ws://localhost:26657/websocket
+ * user_data: tx data to send
+ * # Safety
+ */
+CroResult cro_broadcast(const char *tenermint_url_string,
+                        const uint8_t *user_data,
+                        uint32_t data_length);
+
+/**
  * create fee algorithm
  * # Safety
  */
@@ -103,9 +127,20 @@ CroResult cro_create_transfer_address(CroHDWalletPtr wallet_ptr,
 
 /**
  * create tx
+ * tx_out: previous allocated Tx
  * # Safety
  */
 CroResult cro_create_tx(CroTxPtr *tx_out, uint8_t network);
+
+/**
+ * create deposit tx
+ * network: network id  ex) 0xab
+ * to_address_user: staking address, null terminated string  , ex) 0x1ad06eef15492a9a1ed0cfac21a1303198db8840
+ * # Safety
+ */
+CroResult cro_create_tx_deposit(CroDepositTxPtr *tx_out,
+                                uint8_t network,
+                                const char *to_address_user);
 
 /**
  * create viewkey, which is for encrypted tx
@@ -119,6 +154,7 @@ CroResult cro_create_viewkey(CroHDWalletPtr wallet_ptr,
 /**
  * destroy address
  * # Safety
+ * addr: previously allocated address
  */
 CroResult cro_destroy_address(CroAddressPtr addr);
 
@@ -131,6 +167,7 @@ CroResult cro_destroy_fee_algorithm(CroFeePtr fee);
 /**
  * destroy bip44 hdwallet
  * # Safety
+ * hdwallet: previously allocated hdwallet
  */
 CroResult cro_destroy_hdwallet(CroHDWalletPtr hdwallet);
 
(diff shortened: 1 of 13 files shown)


leejw51crypto:
this is for 
unbond, withdrawal, transfer, deposit c-bindings

leejw51crypto:
ok

cro-clib/src/transaction.rs
```diff
@@ -0,0 +1,318 @@
...
+use std::ptr;
+use std::str::FromStr;
+use std::string::ToString;
+fn do_cro_unbond(
+    network: u8,
+    nonce: u64,
+    from_address: &CroAddress,
+    to_address_user: &str,
+    amount: u64,
+) -> Result<Vec<u8>> {
```
tomtau: just by looking at the name `do_cro_unbond`, the arguments and the return type `Result<Vec<u8>>` without any comments one cannot understand what this function is supposed to do:
1. `to_address_user: &str` -- what format is this supposed to be in? Hex?
2. `amount: u64` -- is this in base units? 
3. `Result<Vec<u8>>` -- when can this fail / return `Err`? What is in that `Vec<u8>`?
...

Even though these are internal functions (not exposed in the bindings), it'll be good to make them more maintainable -- e.g. add a comment explaining such things, give a better name (`create_encoded_signed_unbond`?), have more precise types for the arguments...

cro-clib/src/transaction.rs
```diff
@@ -0,0 +1,318 @@
...
+/// to_address_user:staking address
+/// amount: carson unit   1 carson= 0.0000_0001 cro
+/// output: signed tx encoded
+#[no_mangle]
+/// # Safety
+pub unsafe extern "C" fn cro_unbond(
+    network: u8,
+    nonce: u64,
+    from_ptr: CroAddressPtr,
+    to_address_user: *const std::os::raw::c_char,
```
tomtau: `use std::os::raw::c_char`
and here just: `*const c_char` ?

cro-clib/src/transaction.rs
```diff
@@ -0,0 +1,318 @@
...
+    let result = TxAux::UnbondStakeTx(transaction, signature);
+    let encoded = result.encode();
+    Ok(encoded)
+}
+
+/// staked -> staked
+/// network: networkid
+/// nonce: nonce of the staked state, use cro_get_staked_state to get this nonce
+/// from_ptr: staking address
+/// to_address_user:staking address
```
tomtau: I guess for this one, it's expecting a null-terminated C-string? Hex-encoded?

cro-clib/src/transaction.rs
```diff
@@ -0,0 +1,318 @@
...
+/// amount: carson unit   1 carson= 0.0000_0001 cro
+/// output: signed tx encoded
+#[no_mangle]
+/// # Safety
```
tomtau: https://doc.rust-lang.org/book/ch14-02-publishing-to-crates-io.html#making-useful-documentation-comments
> Safety: If the function is unsafe to call (we discuss unsafety in Chapter 19), there should be a section explaining why the function is unsafe and covering the invariants that the function expects callers to uphold.

I think rather than having these empty sections, it'd be good to explain what is expected -- I guess it's expecting that pointers point to valid / previously created data with `cro_create_*`? No null pointers? That string is null-terminated?

cro-clib/src/transaction.rs
```diff
@@ -0,0 +1,318 @@
...
+/// viewkeys: viewkey list, this is string list
+/// output: minimum 1000 bytes, signed tx encoded
+#[no_mangle]
+/// # Safety
+pub unsafe extern "C" fn cro_withdraw(
+    tenermint_url_string: *const c_char,
+    network: u8,
+    from_ptr: CroAddressPtr,
+    to_ptr: CroAddressPtr,
+    viewkeys: *const *const c_char, // viewkeys
```
tomtau: hex encoded?

tomtau: should clarified in the comments -- how the pubkey is serialised and encoded that's expected in this input

cro-clib/src/transaction.rs
```diff
@@ -0,0 +1,318 @@
...
+    viewkey_count: i32,
```
tomtau: total size of the the array or the the number of items (as it can be calculated if each item is assumed to have the same size)? 
should be clarified

cro-clib/src/fee.rs
```diff
@@ -41,6 +41,25 @@ pub unsafe extern "C" fn cro_estimate_fee(fee_ptr: CroFeePtr, tx_payload_size: u
     let coin_value: u64 = fee_value.to_coin().into();
     coin_value
 }
+
+/// estimate fee after encryption
+/// tx_payload_size: in bytes
+#[no_mangle]
+/// # Safety
+pub unsafe extern "C" fn cro_estimate_fee_after_encrypt(
```
tomtau: not sure if used?

cro-clib/src/wallet.rs
```diff
@@ -108,9 +108,10 @@ pub unsafe extern "C" fn cro_create_transfer_address(
     let (public, private) = wallet
         .seed
         .derive_key_pair(network, CroAccount::Transfer as u32, index)
-        .unwrap();
+        .expect("dereive key pair");
```
linfeng-crypto: `derive` not `dereive`

cro-clib/src/transaction.rs
leejw51crypto: ok

cro-clib/src/transaction.rs
leejw51crypto: ok

cro-clib/src/transaction.rs
leejw51crypto: i'll add more explanation for arguments

cro-clib/src/transaction.rs
leejw51crypto: ok

cro-clib/src/transaction.rs
leejw51crypto: ok

cro-clib/src/transaction.rs
leejw51crypto: i'll add more explanation, it's count of viewkeys.
viewkey is null-terminated strings.

cro-clib/src/fee.rs
leejw51crypto: if there is fee, it will be used.
because user need to compute fee, decide the amount to return to the sender.

cro-clib/src/wallet.rs
leejw51crypto: ok

cro-clib/chain.h
```diff
@@ -66,6 +80,16 @@ CroResult cro_basic_restore_transfer_address(CroAddressPtr *address_out, const u
...
 
+/**
+ * staked -> utxo
+ * tendermint_url: ws://localhost:26657/websocket
+ * user_data: tx data to send
+ * # Safety
+ */
+CroResult cro_broadcast(const char *tenermint_url_string,
+                        const uint8_t *user_data,
+                        uint32_t data_length);
```
yihuang: I have some worry about simply encapsulating everything(including network and storage) into the library, it will put many limitations on the user.
For example, on the mobile device, the user might need to use the native network stack, to handle platform-specific things, like permission, network status switch, etc.
Also, it can't be used in event-driven or async language or framework easily, like python's async, or nodejs.

tomtau: Yeah, there are two sides to this. I think we can stop with the C-bindings work at this stage, so that we don't overdo it, and wait for @ben-crypto feedback when they try to use it on iOS / Android -- either they'd want more of it (so that they don't need to worry about syncs, utxo management etc.) or less of it (due to the platform-specific things, as you mentioned)

leejw51crypto: ok, i agree

yihuang: For those users who want the whole thing and don't care about networking stack and blocking IO,  we can encapsulate the `client-rpc` with a single API: `spawn_client_process`, they can access the functionality through JSON-RPC.

tomtau: the problem with JSON-RPC is that its HTTP port is open to other applications on the same host (or even the outside world if one isn't careful with settings and it's running on a machine with a public IP)

yihuang: > the problem with JSON-RPC is that its HTTP port is open to other applications on the same host (or even the outside world if one isn't careful with settings and it's running on a machine with a public IP)

We can support the UNIX domain socket then.

tomtau: Not sure if unix domain sockets can work on non-jailbroken iOS.

> In iOS 10, daemons/processes with the "seatbelt" entitlement (e.g. mediaserverd) don't allow you to create or connect to unix domain sockets anymore.

yihuang: > Not sure if unix domain sockets can work on non-jailbroken iOS.
> 
> > In iOS 10, daemons/processes with the "seatbelt" entitlement (e.g. mediaserverd) don't allow you to create or connect to unix domain sockets anymore.

Or we can expose the `client-rpc` as an single API like `fn client_execute(json_request) -> json_response`.

PR #1090, opened 2020-02-18
Problem (Fix #1089): the upper bound of reward parameters are too restrictive

### LINKED ISSUE #1089: Problem: the upper bound of parameters of reward distribution are too restrictive
Currently, there are checks:
```
        if self.monetary_expansion_tau > 100_00000000_00000000_u64 {
            return Err("tau too big");
        }
```
It's actually pointless, and there are needs to set a higher value than that, @lezzokafka .

yihuang:
Solution:
  - Add more reasoning into the range of parameters.

The reasoning behind the choice:
https://github.com/crypto-com/chain/issues/1089#issuecomment-587327223

### DIFF
--- a/chain-abci/tests/validator_changes.rs
+++ b/chain-abci/tests/validator_changes.rs
@@ -147,10 +147,10 @@ fn check_rejoin() {
         |parameters| {
             // tweaking times + parameters, something more than 0.0... gets minted
             parameters.unbonding_period = 3600 * 24 * 10000;
-            parameters.rewards_config.reward_period_seconds = 3600 * 12 * 10000;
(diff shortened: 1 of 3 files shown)


yihuang:
> ok to merge, but need to understand more context.
> current docs have this upper bound mentioned there: https://crypto-com.github.io/getting-started/staking.html#rewards
> my assumption was that most of these bounds were there for "safe" (no panic or strange results) operations with fixed points?

Yes, there is no logical constraint to the upper bound of both `r0` and `tau`, I'll check the safety of fixed point computation again, check sent pr to update docs first before we merge this one.

tomtau:
bors retry

PR #1092, opened 2020-02-18
Problem (Fix #1091): Inactivated validator can still appear in block proposer

### LINKED ISSUE #1091: Problem: Inactivated validator can still recorded in reward stats
Because validator is inactivated in `end_block`, and rewards stats are cleaned up immediately, but validator removal in tendermint only takes effect in `N+2` blocks, so the validator can still appear in block proposer or `last_commit_info`. When recording reward stats, needs to check if the validator if active.

yihuang:
Solution:
- Check validator not active when record reward stats


tomtau:
bors retry

PR #1099, opened 2020-02-18
Problem:(fix #1027) inputted coin value in client-cli is not confirmed

### LINKED ISSUE #1027: Problem: inputted coin value in client-cli is not confirmed
coin value
1 carson== 0.0000_0001 cro
1 cro = 1_0000_0000 carson
is very important
1. double confirm the value
2. double confirm for the unit (carson or cro)

linfeng-crypto:
solution: confirm coin input in client-cli

### DIFF
--- a/client-cli/src/command/transaction_command.rs
+++ b/client-cli/src/command/transaction_command.rs
@@ -467,11 +467,7 @@ fn new_unbond_transaction<N: NetworkOpsClient>(
 ) -> Result<TxAux> {
     let attributes = StakedStateOpAttributes::new(get_network_id());
     let address = ask_staking_address()?;
-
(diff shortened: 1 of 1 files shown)


linfeng-crypto:
> looks all right -- it may also be good to add the expected form ("with a decimal point, e.g. 1.5") after "in CRO" for a more explanatory description.
> and there are a few formatting issues (I think they can be resolved with "cargo fmt"):
> https://travis-ci.org/crypto-com/chain/jobs/651842522#L1290

Printout the result in CRO, like "1.5 CRO", and use `[Y|N]` to confirm the result. If not confirmed, make user input again in a loop.

PR #1100, opened 2020-02-18
Problem (Fix #1095): Reward distribution don't increase staking account's nonce

### LINKED ISSUE #1095: Problem: Reward distribution don't increase staking account's nonce

yihuang:
Solution:
- Increase nonce by one when add reward.


yihuang:
> unit test fails -- I think it on L203 you can increase the nonce that the transaction is created with.
> also, it can be good to add a unit test that only captures this bug (e.g. do a reward distribution and check that nonces of affected accounts were updated)

Done, add checking of nonce in the standard reward distribution test case.

Issue #914, opened 2020-01-16
Problem: Unable to restore wallet with the same seed (or mnemonic words) as existing wallet.


yihuang:
Steps to reproduce:
```
$ chainrpc.py wallet restore "pony thank pluck sweet bless tuna couple eight stove fluid essay debate cinnamon elite only" --name Default
$ chainrpc.py wallet restore "pony thank pluck sweet bless tuna couple eight stove fluid essay debate cinnamon elite only" --name test
Traceback (most recent call last):
    ...
jsonrpcclient.exceptions.ReceivedErrorResponseError: Decryption error: Incorrect enckey: Unable to unlock stored values
```

The reason is different wallets have different encryption keys now, so it can't load the key pair created by the other wallet.

tomtau:
@yihuang you get this error `jsonrpcclient.exceptions.ReceivedErrorResponseError: Decryption error: Incorrect enckey: Unable to unlock stored values`'

after executing the second `restore`? or after some other command (not sure if to interpret "..." as other commands or a trimming of the error message)

yihuang:
> @yihuang you get this error `jsonrpcclient.exceptions.ReceivedErrorResponseError: Decryption error: Incorrect enckey: Unable to unlock stored values`'
> 
> after executing the second `restore`? or after some other command (not sure if to interpret "..." as other commands or a trimming of the error message)

It's trim of error messages, I edited a bit to make it clearer.

yihuang:
A potential solution might be to prefix the key of `KeyService` (maybe other services as well if necessary) with the wallet name.

tomtau:
I guess both are fine -- what are the impacts (positive/negative) of each?

yihuang:
Option two:
Positives:
- Able to iterate key pairs of a wallet without referencing other information.
Negatives:
- Have to load/save all key pairs in bulk, even if we only want to operate on a single one. Since we already go with this structure in `Wallet`, so I think it's fine.
- Store view key pair with a different storage key if we want to encrypt it with a different token.

leejw51crypto:
i think option 1 is good
KEYSPACE+ wallet_name prefix

yihuang:
> i think option 1 is good
> KEYSPACE+ wallet_name prefix
> 
> a little in-efficient, but will be no problem

Option two would allow import/export wallet as a whole, without decrypting information.

leejw51crypto:
as long as we use enckey per wallet, backup & restore with one shot is not possible

[...]

PR #1106, opened 2020-02-19
Problem:(fix #827)no import/export for non-HD keys

### LINKED ISSUE #827: Problem: no import/export for non-HD keys

linfeng-crypto:
Solution: 
* add `export` method to backup a wallet, and `import` method to store the wallet again, both in client-rpc and client-cli
* add unit test

### DIFF
--- a/client-cli/src/command/wallet_command.rs
+++ b/client-cli/src/command/wallet_command.rs
@@ -1,4 +1,4 @@
-use quest::{ask, success, text};
+use quest::{ask, error, success, text};
 use secstr::SecUtf8;
 use structopt::StructOpt;
 
@@ -6,7 +6,12 @@ use client_common::{Error, ErrorKind, PrivateKey, Result, ResultExt};
 use client_core::types::WalletKind;
 use client_core::{Mnemonic, WalletClient};
 
-use crate::ask_passphrase;
+use crate::{ask_passphrase, ask_seckey};
+use client_core::service::WalletInfo;
+use client_core::wallet::WalletRequest;
+use std::fs::File;
+use std::io::{Read, Write};
+use std::path::PathBuf;
 
 const WALLET_KIND_VARIANTS: [&str; 2] = ["basic", "hd"];
 
@@ -30,6 +35,43 @@ pub enum WalletCommand {
             case_insensitive = true
         )]
         wallet_type: WalletKind,
+    },
+    #[structopt(name = "export", about = "Backup wallet to a file")]
+    Export {
+        #[structopt(
+            name = "wallet name",
+            short = "n",
+            long = "name",
+            help = "Name of wallet (comma separated if many) "
+        )]
+        name: Option<String>,
+        #[structopt(
+            name = "from_file",
+            short = "f",
+            long = "from_file",
+            parse(from_os_str),
+            help = r#"json file contains a list of {"name": wallet_name, "auth_token": auth_token}"#
+        )]
+        from_file: Option<PathBuf>,
+        #[structopt(
+            name = "to_file",
+            short = "t",
+            long = "to_file",
+            parse(from_os_str),
+            help = "file to dump the wallet information"
+        )]
+        to_file: Option<PathBuf>,
+    },
+    #[structopt(name = "import", about = "Import a wallet from a file")]
+    Import {
+        #[structopt(
+            name = "file",
+            short = "f",
+            long = "file",
+            parse(from_os_str),
+            help = "file stored the wallet info"
+        )]
+        file: PathBuf,
     },
     #[structopt(name = "list", about = "List all wallets")]
     List,
@@ -86,6 +128,12 @@ impl WalletCommand {
             WalletCommand::RestoreBasic { name } => Self::restore_basic_wallet(wallet_client, name),
             WalletCommand::AuthToken { name } => Self::auth_token(wallet_client, name),
             WalletCommand::Delete { name } => Self::delete(wallet_client, name),
+            WalletCommand::Export {
+                name,
+                from_file,
+                to_file,
+            } => Self::export(wallet_client, name, from_file, to_file),
+            WalletCommand::Import { file } => Self::import(wallet_client, file),
         }
     }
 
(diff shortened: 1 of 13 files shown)


client-cli/src/command/wallet_command.rs
```diff
@@ -31,6 +35,35 @@ pub enum WalletCommand {
...
+        )]
+        name: String,
+        #[structopt(
+            name = "file",
+            short = "f",
+            long = "file",
+            parse(from_os_str),
+            help = "file to dump wallet"
+        )]
+        file: PathBuf,
```
yihuang: I suggest to treat a special value "-" as stdout and make it the default value, so the command can work with pipes in shell.

client-cli/src/command/wallet_command.rs
```diff
@@ -31,6 +35,35 @@ pub enum WalletCommand {
...
+    #[structopt(name = "import", about = "Import a wallet from a file")]
+    Import {
+        #[structopt(
+            name = "file",
+            short = "f",
+            long = "file",
+            parse(from_os_str),
+            help = "file stored the wallet info"
+        )]
+        file: PathBuf,
```
yihuang: I suggest to treat a special value "-" as stdin and make it the default value, so the command can work with pipes in shell.

yihuang:
We had an issue before to import/export all wallets at once, which is closed for duplication of this one, so I presume this one should support import/export all wallets at once?

tomtau:
> We had an issue before to import/export all wallets at once, which is closed for duplication of this one, so I presume this one should support import/export all wallets at once?

@leejw51crypto is this one ok as is, or should it take say several names of wallets, or a flag `--all` and ask for individual details (and then dump all in one file or ask for several ?)

client-cli/src/command/wallet_command.rs
linfeng-crypto: Currently, I serialized the wallet info as multiline json string using the `serde_json::to_string_pretty`. 

If inputted from the terminal, I have to serialize the wallet info into one line json string. By the way, the output is colored and there is a waring messge like `Warning! `CRYPTO_CHAIN_ID` environment variable is not set. Setting network to devnet and network-id to 0`. I don't think the the pipes can work well.

leejw51crypto:
import and export all wallets would be better

tomtau:
> import and export all wallets would be better
> dump all to one file also using one password?
> if decoding fails, we can ask again
> 
> consider that wallets can be over 100 or thousands

1. will anyone have client-cli with that many wallets? there'll be other issues arising for that, so it can be left for now

2. each wallet may a different passphrase

leejw51crypto:
yes, that's possible scenario for special use case

client-cli/src/command/wallet_command.rs
yihuang: I mean something like this:
```
let output = if name == "-" {
  std::io::stdout
} else {
  File::open(name)...
};
output.write_all(...);
```

yihuang:
Is it better to let export/import the encrypted data as is, not to decrypt and re-encrypt? This will work for multiple wallets with different passphrases.
And it's safer because information is still encrypted?

linfeng-crypto:
> Is it better to let export/import the encrypted data as is, not to decrypt and re-encrypt? This will work for multiple wallets with different passphrases.
> And it's safer because the information is not decrypted?

I think it's user's response to encrypt data and keep the file safe, just like the  mnemonics. It's not the client's business.

client-cli/src/command/wallet_command.rs
```diff
@@ -31,6 +37,43 @@ pub enum WalletCommand {
...
+            name = "wallet name",
+            short = "n",
+            long = "name",
+            help = "Name of wallet"
+        )]
+        name: Option<String>,
+        #[structopt(
+            name = "from_file",
+            short = "f",
+            long = "from_file",
```
leejw51crypto: how about just form

client-cli/src/command/wallet_command.rs
```diff
@@ -31,6 +37,43 @@ pub enum WalletCommand {
...
+            parse(from_os_str),
+            help = r#"json file contains a list of {"name": wallet_name, "auth_token": auth_token}"#
+        )]
+        from_file: Option<PathBuf>,
+        #[structopt(
+            name = "to_file",
+            short = "t",
+            long = "to_file",
```
leejw51crypto: just 'to'?

client-cli/src/command/wallet_command.rs
leejw51crypto: - is too short , "console"?

leejw51crypto:
also addresses, roothashes(mutlsig addresses) need to export, import

yihuang:
> > Is it better to let export/import the encrypted data as is, not to decrypt and re-encrypt? This will work for multiple wallets with different passphrases.
> > And it's safer because the information is not decrypted?
> 
> I think it's user's response to encrypt data and keep the file safe, just like the mnemonics. It's not the client's business.

I mean if we don't decrypt the information, we can simply:
```
client-cli export --all
```
No need to worry about different passphrases of different wallets.

client-cli/src/command/wallet_command.rs
yihuang: > `-` is too short , "console"?

`-` is like a conventional name for this.

client-cli/src/command/wallet_command.rs
tomtau: I think it's ok -- if it's too long, one can just use `-f ... -t ...`?

client-cli/src/command/wallet_command.rs
tomtau: https://github.com/crypto-com/chain/pull/1106#discussion_r382356917

client-cli/src/command/wallet_command.rs
linfeng-crypto: Ok

linfeng-crypto:
> > > Is it better to let export/import the encrypted data as is, not to decrypt and re-encrypt? This will work for multiple wallets with different passphrases.
> > > And it's safer because the information is not decrypted?
> > 
> > 
> > I think it's user's response to encrypt data and keep the file safe, just like the mnemonics. It's not the client's business.
> 
> I mean if we don't decrypt the information, we can simply:
> 
> ```
> client-cli export --all
> ```
> 
> No need to worry about different passphrases of different wallets.

sure

yihuang:
> > > > Is it better to let export/import the encrypted data as is, not to decrypt and re-encrypt? This will work for multiple wallets with different passphrases.
> > > > And it's safer because the information is not decrypted?
> > > 
> > > 
> > > I think it's user's response to encrypt data and keep the file safe, just like the mnemonics. It's not the client's business.
> > 
> > 
> > I mean if we don't decrypt the information, we can simply:
> > ```
> > client-cli export --all
> > ```
> > 
> > 
> > No need to worry about different passphrases of different wallets.
> 
> sure, to get the private key, the workflow is `wallet_name -> public key -> private key`, but we should give a enckey to get the public key, or we have to store the un-encrypted map`wallet_name->pubkey`, is it safe?

Yeah, that's a block, the structure of the key service is to be changed in #914, after that it probably will be changed to `wallet_name -> key_pairs`. 
Before that, I guess we just have to stick to the current solution?

linfeng-crypto:
> > > > > Is it better to let export/import the encrypted data as is, not to decrypt and re-encrypt? This will work for multiple wallets with different passphrases.
> > > > > And it's safer because the information is not decrypted?
> > > > 
> > > > 
> > > > I think it's user's response to encrypt data and keep the file safe, just like the mnemonics. It's not the client's business.
> > > 
> > > 
> > > I mean if we don't decrypt the information, we can simply:
> > > ```
> > > client-cli export --all
> > > ```
> > > 
> > > 
> > > No need to worry about different passphrases of different wallets.
> > 
> > 
> > sure, to get the private key, the workflow is `wallet_name -> public key -> private key`, but we should give a enckey to get the public key, or we have to store the un-encrypted map`wallet_name->pubkey`, is it safe?
> 
> Yeah, that's a block, the structure of the key service is to be changed in #914, after that it probably will be changed to `wallet_name -> key_pairs`.
> Before that, I guess we just have to stick to the current solution?

sure

client-core/src/service/wallet_service.rs
```diff
@@ -7,12 +7,70 @@ use chain_core::init::address::RedeemAddress;
...
+    /// name the the wallet
+    pub name: String,
+    /// wallet meta data
+    #[serde(deserialize_with = "deserde_from_str", serialize_with = "serde_to_str")]
+    pub wallet: Wallet,
+    /// private key of the wallet
+    #[serde(deserialize_with = "deserde_from_str", serialize_with = "serde_to_str")]
+    pub private_key: PrivateKey,
+    /// passphrase used when import wallet
+    pub passphrase: String,
```
tomtau: `String` -> `SecUtf8`?

tomtau: Or Option<SecUtf8>, as it's optional (the client would ask for it if not provided?)?

client-cli/src/command/wallet_command.rs
```diff
@@ -122,6 +171,135 @@ impl WalletCommand {
...
+                for setting in wallet_settings {
+                    let name = setting["name"]
+                        .as_str()
+                        .chain(|| (ErrorKind::InvalidInput, "expect name"))?;
+                    let key = setting["auth_token"]
+                        .as_str()
+                        .chain(|| (ErrorKind::InvalidInput, "Expect auth_token"))?;
+                    let enckey = parse_hex_enckey(&key).map_err(|_e| {
+                        error(&format!("error auth_token of wallet {}", name));
+                    });
```
tomtau: probably giving it a data type "Vec<WalletExportSettings>" with serde implementations would be less error-prone ?

linfeng-crypto: Ok

client-core/src/service/wallet_service.rs
```diff
@@ -7,12 +7,71 @@ use chain_core::init::address::RedeemAddress;
...
+    /// name of the the wallet
+    pub name: String,
+    /// wallet meta data
+    #[serde(deserialize_with = "deserde_from_str", serialize_with = "serde_to_str")]
+    pub wallet: Wallet,
+    /// private key of the wallet
+    #[serde(deserialize_with = "deserde_from_str", serialize_with = "serde_to_str")]
+    pub private_key: PrivateKey,
+    /// passphrase used when import wallet
+    pub passphrase: Option<SecUtf8>,
```
calvinaco: Is this passphrase supposed to be manually filled on import?

linfeng-crypto: yes, if the passphrase is None, the client will ask for it.

client-cli/src/command/wallet_command.rs
linfeng-crypto: This is one way to use it, but only some developers know it.

yihuang: > How does the command work with `-`, like this?
> 
> ```
> cat wallet_file | ./client-cli wallet import -
> ```

Yes, with pipes it can work with other tools together, for example, compression tool:
```
$ ./client-cli wallet export --all | gzip > wallets.export.gz
$ zcat wallets.export.gz | ./client-cli wallet import
```

PR #1117, opened 2020-02-21
Problem (Fix #1069): blocks with multiple txs may fail light client app hash check

### LINKED ISSUE #1069: Problem: blocks with multiple txs may fail light client app hash check
Error messages:

```
1) Staking
--
806 | should support staking, unbonding and withdrawing:
807 | Error: Error when synchronizing wallet: Verify error: state app hash don't match block header
808 | at Object.exports.asyncMiddleman (test/core/utils.ts:108:9)
809 | at process._tickCallback (internal/process/next_tick.js:68:7)
810 |  
811 | 2) Wallet Auto-sync
812 | can auto-sync unlocked wallets:
813 | Error: Error when synchronizing receiver wallet: Verify error: state app hash don't match block header
814 | at Object.exports.asyncMiddleman (test/core/utils.ts:108:9)
815 | at process._tickCallback (internal/process/next_tick.js:68:7)
816 |  
817 | 3) Wallet transaction
818 | Zero Fee
819 | can transfer funds between two wallets:
820 | Error: Error when synchronizing receiver wallet: Verify error: state app hash don't match block header
821 | at Object.exports.asyncMiddleman (test/core/utils.ts:108:9)
822 | at process._tickCallback (internal/process/next_tick.js:68:7)


```

yihuang:
Solution:
- Change BTreeMap to IndexMap


yihuang:
> would be good to add a small test that demonstrates the bug

Yeah, I just want to unblock @calvinlauco 's work first, we can add the integration tests like discussed on slack.

calvinlauco:
> > would be good to add a small test that demonstrates the bug
> 
> Yeah, I just want to unblock @calvinlauco 's work first, we can add the integration tests like discussed on slack.

Thanks @yihuang , Let me try with this first and get back to you.
@lezzokafka

yihuang:
Fixed the unit test build.

lezzokafka:
> Thanks @yihuang , Let me try with this first and get back to you.
> @lezzokafka

Tested, it passed the problematic block :D

tomtau:
> I also verified that it has solved the problem. Should we wait for the test case as it seems complicated to create one. Or we should merge this to fix the current issue first?

i think there could at least be a small unit test that captures the bug (it may need isolating out some of the sync functionality)

calvinaco:
> > would be good to add a small test that demonstrates the bug
> 
> Yeah, I just want to unblock @calvinlauco 's work first, we can add the integration tests like discussed on slack.

@yihuang 
I have extracted the block data and create a test here
https://github.com/calvinlauco/chain/commit/18a14ba54fb1d34f0c2f9717fedc195cd176bca2

and then I applied your fix here
https://github.com/calvinlauco/chain/commit/029a065c2ef95ed6fba11a6d8cd3d28f69eaa8c6

yihuang:
I just added an multinode integration test for this.

yihuang:
> 👍 can you squash the commits?

Done

client-common/src/tendermint/types/block_results.rs
```diff
@@ -56,11 +56,11 @@ pub struct Attribute {
 
 impl BlockResults {
     /// Returns transaction ids and the corresponding fees in block results
-    pub fn fees(&self) -> Result<BTreeMap<TxId, Fee>> {
+    pub fn fees(&self) -> Result<IndexMap<TxId, Fee>> {
```
leejw51crypto: could you add some explanation why "change to IndexMap"?

client-core/src/wallet/syncer.rs
```diff
@@ -484,7 +485,7 @@ pub(crate) struct FilteredBlock {
     /// Block time
     pub block_time: Time,
     /// List of successfully committed transaction ids in this block and their fees
-    pub valid_transaction_fees: BTreeMap<TxId, Fee>,
+    pub valid_transaction_fees: IndexMap<TxId, Fee>,
```
leejw51crypto: how about
Vec< (TxId, Fee) >
because order should be preserved

leejw51crypto: index map is actually
IndexMap == Hashtable + List

if we port to other language (haskell, c),
it would much simpler if we use Vec

tomtau: changing to `IndexMap` seems to be the simplest solution to #1069 -- changing to Vec (or other restructuring) would require more complex code changes

integration-tests/multinode/multitx_test.py
```diff
@@ -0,0 +1,73 @@
...
+    wait_for_tx(rpc, txid)
+    rpc.wallet.sync()
+
+print('Stop node1')
+stop_node(supervisor, 'node1')
+
+last_height = latest_block_height(rpc)
+
+print('Send multiple tx')
+pending_txs = [rpc.wallet.send(transfer1, amount) for _ in addresses]
```
tomtau: is `rpc.wallet.send` async?
It seems there needs to be some "wait", otherwise node1 may be started up before everything else?

yihuang: > is `rpc.wallet.send` async?
> It seems there needs to be some "wait", otherwise node1 may be started up before everything else?

Even if it's async, the tx should be processed eventually? Maybe `wait_for_tx` need some more timeout?

yihuang: Changed a little bit, let's see how that works?

tomtau: > Even if it's async, the tx should be processed eventually? 

yes, but my guess is that the goal is here is that these TXs end up in mempool while node1 is down, so that they all get picked up for the same block. If it starts up before that, they'll get picked up immediately for different blocks

yihuang: > > Even if it's async, the tx should be processed eventually?
> 
> yes, but my guess is that the goal is here is that these TXs end up in mempool while node1 is down, so that they all get picked up for the same block. If node1 starts up before that / while they are being submitted, they may get picked up immediately for different blocks

Yes, I added 1-second sleep for this.

PR #1118, opened 2020-02-21
Problem (Fix #949): deposit amount fee calculation is not correct

### LINKED ISSUE #949: Problem: "deposit amount" fee calculation is not correct
start tendermint with fee, deposit 500000, but get 500074.

deposited amount in the staked state is larger than what's specified in the command line.

the likely cause is dummysigner estimation

### LINKED ISSUE #1101: Problem: withdraw all coins estimates a wrong fee when extra view keys are entered
steps to reproduce (using client-cli):
1. configure network with non-zero fee -- let one address (A) as unbonded
2. in a different wallet, create transfer address + view key (B)
3. try withdraw all coins from A and provide B details

what happens: validation fails with
> Error: Tendermint RPC error: verification failed: transaction input output coin (plus fee) sums don't match

what should happen -- transaction goes through

yihuang:
Solution:
- Fix DummySigner for deposit transaction
- Fix mock_enc_dec payload padding
- Test withdraw with viewkeys in the multinode join test (Fix #1101  )

### DIFF
--- a/client-core/src/signer/dummy_signer.rs
+++ b/client-core/src/signer/dummy_signer.rs
@@ -22,10 +22,8 @@ pub struct DummySigner();
 impl DummySigner {
     /// pad payload to the multiples of 128bits
     fn pad_payload(&self, plain_txaux: PlainTxAux) -> Vec<u8> {
-        let unit = 16_usize;
-        let plain_payload_len = plain_txaux.encode().len();
-        // PKCS7 padding -- if the len is multiple of the block length, it'll still be padded
-        vec![0; plain_payload_len + unit - plain_payload_len % unit]
+        // https://tools.ietf.org/html/rfc8452
+        vec![0; plain_txaux.encode().len() + 16]
     }
 
     /// Creates a mock merkletree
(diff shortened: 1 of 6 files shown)


client-core/src/signer/dummy_signer.rs
```diff
@@ -25,7 +25,7 @@ impl DummySigner {
         let unit = 16_usize;
         let plain_payload_len = plain_txaux.encode().len();
         // PKCS7 padding -- if the len is multiple of the block length, it'll still be padded
-        vec![0; plain_payload_len + unit - plain_payload_len % unit]
+        vec![0; plain_payload_len + unit - plain_payload_len % unit + 5]
     }
```
tomtau: rfc8452 doesn't seem to have this padding -- seems the length should be `plain_payload_len + tag_len(16)`

tomtau: as Aad is txid, fixed to 32

yihuang: > as Aad is txid, fixed to 32

Yeah, `plain_len + 32` passes the test too, = ="

yihuang:
> also there's a fee computation function in c bindings

`cro_estimate_fee`? that one needs bigger change because it only takes a `tx_payload_size` as an argument, which I think it's the size of the tx before obfuscation, that's not enough to compute the fee.

yihuang:
#1101 is also solved, it's verified in the multinode join test, because normal integration tests will withdraw all at the beginning.

tomtau:
bors try

PR #1122, opened 2020-02-23
Problem (Fix #1121): Too tedious to create separated c apis for all client features

### LINKED ISSUE #1121: Problem: Too tedious to create separated c apis for all client features
Separated from #1115.

We can classify c apis into two categories:
- Stateless, these are pure functions, and we should provide them with normal c APIs, they can be embedded into user's code perfectly.
  - keypair derivation
  - transaction signing
  - fee policy
  - UTxO selection algorithm implemented in a stateless way
- Stateful features, there are features involves storage state and blocking network IO.
  These APIs should be taken extra care when used in UI thread or event-driven environment.
  - Wallet syncing.
  - Transaction building and broadcasting.
  - Basically most of the features in `client-rpc`.

I propose to expose the stateful features with a common JSON API:

```rust
fn jsonrpc_call(storage_path, tendermint_rpc_url, network_id, json_request) -> json_response
```

> It might be expanded a little bit as we investigate deeper, but the number of APIs should be a handful.

yihuang:
Solution:
- create `jsonrpc_call` api to encapulate all commands of `client-rpc`.

### DIFF
--- a/client-core/src/synchronizer.rs
+++ b/client-core/src/synchronizer.rs
@@ -1,5 +1,6 @@
 //! global polling synchronizer
 use std::collections::BTreeMap;
+use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::mpsc::channel;
 use std::sync::{Arc, Mutex};
 use std::thread;
@@ -19,6 +20,9 @@ type WalletPassphrase = SecKey;
 pub struct PollingSynchronizer {
     wallets: Arc<Mutex<BTreeMap<WalletName, WalletPassphrase>>>,
     progress: Arc<Mutex<SynchronizerProgress>>,
+    should_run: Arc<AtomicBool>,
+    sync_thread: Option<thread::JoinHandle<()>>,
+    progress_thread: Option<thread::JoinHandle<()>>,
 }
 
 impl PollingSynchronizer {
@@ -45,30 +49,37 @@ impl PollingSynchronizer {
         let wallets = self.wallets.clone();
         let progress = self.progress.clone();
 
+        self.should_run.store(true, Ordering::Relaxed);
+        let should_run = self.should_run.clone();
+
         let (sender, receiver) = channel();
-        thread::spawn(move || loop {
-            let wallets_to_synchronize = wallets
-                .lock()
-                .expect("Unable to acquire lock on wallets to synchronize in polling synchronizer")
-                .clone();
+        self.sync_thread = Some(thread::spawn(move || {
+            while should_run.load(Ordering::Relaxed) {
+                let wallets_to_synchronize = wallets
+                    .lock()
+                    .expect(
+                        "Unable to acquire lock on wallets to synchronize in polling synchronizer",
+                    )
+                    .clone();
 
-            for (name, enckey) in wallets_to_synchronize.iter() {
-                let result = WalletSyncer::with_obfuscation_config(
-                    config.clone(),
-                    Some(sender.clone()),
-                    name.clone(),
-                    enckey.clone(),
-                )
-                .and_then(|syncer| syncer.sync());
-                if let Err(e) = result {
-                    log::error!("Error while synchronizing wallet [{}]: {:?}", name, e);
+                for (name, enckey) in wallets_to_synchronize.iter() {
+                    let result = WalletSyncer::with_obfuscation_config(
+                        config.clone(),
+                        Some(sender.clone()),
+                        name.clone(),
+                        enckey.clone(),
+                    )
+                    .and_then(|syncer| syncer.sync());
+                    if let Err(e) = result {
+                        log::error!("Error while synchronizing wallet [{}]: {:?}", name, e);
+                    }
                 }
+
+                thread::sleep(time::Duration::from_millis(100));
             }
+        }));
 
-            thread::sleep(time::Duration::from_millis(100));
-        });
-
-        thread::spawn(move || {
+        self.progress_thread = Some(thread::spawn(move || {
             for progress_report in receiver.iter() {
                 match progress_report {
                     ProgressReport::Init {
@@ -106,7 +117,18 @@ impl PollingSynchronizer {
                     }
                 }
             }
-        });
+        }));
+    }
+
+    /// Stop the auto synchronizer thread
+    pub fn stop(&mut self) {
+        self.should_run.store(false, Ordering::Relaxed);
+        if let Some(thread) = self.sync_thread.take() {
+            thread.join().expect("join auto syncer thread failed");
+        }
+        if let Some(thread) = self.progress_thread.take() {
+            thread.join().expect("join progress report thread failed");
+        }
     }
 }
 
--- a/cro-clib/chain.h
+++ b/cro-clib/chain.h
@@ -272,6 +272,31 @@ CroResult cro_join(uint8_t network,
 
 /**
  * # Safety
+ *
+ * Should not be called with null pointers.
+ *
+ * c example:
+ *
+ * ```c
+ * char buf[BUFSIZE];
+ * const char* req = "{\"jsonrpc\": \"2.0\", \"method\": \"wallet_list\", \"params\": [], \"id\": 1}";
+ * int retcode = cro_jsonrpc_call("./data", "ws://...", 0xab, req, buf, sizeof(buf));
+ * if (retcode == 0) {
+ *     printf("response: %s\n", buf);
+ * } else {
+ *     printf("error: %s\n", buf);
+ * }
+ * ```
+ */
+CroResult cro_jsonrpc_call(const char *storage_dir,
+                           const char *websocket_url,
+                           uint8_t network_id,
+                           const char *request,
+                           char *buf,
+                           uintptr_t buf_size);
+
+/**
+ * # Safety
  */
 CroResult cro_restore_hdwallet(const char *mnemonics_string, CroHDWalletPtr *wallet_out);
 
--- a/cro-clib/src/jsonrpc.rs
+++ b/cro-clib/src/jsonrpc.rs
@@ -1,0 +1,194 @@
+use jsonrpc_core::IoHandler;
+use std::ffi::{CStr, CString};
+use std::os::raw::c_char;
+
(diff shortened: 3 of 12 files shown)


yihuang:
Python test:
```python
import json
from ctypes import *

lib = cdll.LoadLibrary('../target/debug/libcro_clib.dylib')
lib.jsonrpc_call.argtypes = (c_char_p, c_char_p, c_char, c_char_p, c_char_p, c_size_t)

storage_path = './data/node0/wallet'.encode()
wsurl = 'ws://127.0.0.1:26657/websocket'.encode()
req = json.dumps({'jsonrpc': '2.0', 'method': 'wallet_list', 'params': [], 'id': 1}).encode()
output = create_string_buffer(1024)
retcode = lib.jsonrpc_call(storage_path, wsurl, b'\xAB', req, output, len(output))
assert retcode == 0, output.value.decode()
print(json.loads(output.value))
# {'jsonrpc': '2.0', 'result': ['Default'], 'id': 1}
```

tomtau:
bors try

yihuang:
One known inefficiency with this API is it open/close the DB, also spawn/stop the auto-synchronizer thread for each API call. We can further optimize that if we agree on the overall direction.

client-core/src/synchronizer.rs
```diff
@@ -19,6 +20,9 @@ type WalletPassphrase = SecKey;
 pub struct PollingSynchronizer {
     wallets: Arc<Mutex<BTreeMap<WalletName, WalletPassphrase>>>,
```
leejw51: how about just using one mutex , for PollingSynchronizer

client-core/src/synchronizer.rs
```diff
@@ -19,6 +20,9 @@ type WalletPassphrase = SecKey;
...
     progress: Arc<Mutex<SynchronizerProgress>>,
+    should_run: Arc<AtomicBool>,
+    sync_thread: Option<thread::JoinHandle<()>>,
```
leejw51crypto: how about using one mutx for wallets, progress, should_run

client-core/src/synchronizer.rs
```diff
@@ -19,6 +20,9 @@ type WalletPassphrase = SecKey;
 pub struct PollingSynchronizer {
     wallets: Arc<Mutex<BTreeMap<WalletName, WalletPassphrase>>>,
     progress: Arc<Mutex<SynchronizerProgress>>,
+    should_run: Arc<AtomicBool>,
```
leejw51crypto: houw about just using bool?

client-core/src/synchronizer.rs
```diff
@@ -45,30 +49,37 @@ impl PollingSynchronizer {
...
+                    .clone();
+
+                for (name, enckey) in wallets_to_synchronize.iter() {
+                    let result = WalletSyncer::with_obfuscation_config(
+                        config.clone(),
+                        Some(sender.clone()),
+                        name.clone(),
+                        enckey.clone(),
+                    )
+                    .and_then(|syncer| syncer.sync());
```
leejw51crypto: in testnet, this sync will take many hours.
maybe days for slow network

so i think for a few wallets,
each thread for wallet ?

or one block data, multiplexing,
but that's beyond this PR

client-core/src/synchronizer.rs
```diff
@@ -45,30 +49,37 @@ impl PollingSynchronizer {
...
-                    config.clone(),
-                    Some(sender.clone()),
-                    name.clone(),
-                    enckey.clone(),
-                )
-                .and_then(|syncer| syncer.sync());
-                if let Err(e) = result {
-                    log::error!("Error while synchronizing wallet [{}]: {:?}", name, e);
+        self.sync_thread = Some(thread::spawn(move || {
+            while should_run.load(Ordering::Relaxed) {
```
leejw51crypto: while should_run {

cro-clib/src/jsonrpc.rs
```diff
@@ -0,0 +1,181 @@
...
+}
+
+fn do_jsonrpc_call(
+    storage_dir: &str,
+    websocket_url: &str,
+    network_id: u8,
+    json_request: &str,
+) -> Result<String> {
+    let mut io = IoHandler::new();
+    let storage = SledStorage::new(storage_dir)?;
```
leejw51crypto: how de-coupling storage?
so that user supply other database for storage

cro-clib/src/jsonrpc.rs
```diff
@@ -0,0 +1,181 @@
...
+        .err_kind(ErrorKind::InvalidInput, || {
+            "stateful command execute failed"
+        })
+}
+
+/// # Safety
+///
+/// Should not be called with null pointers.
+#[no_mangle]
+pub unsafe extern "C" fn jsonrpc_call(
```
leejw51crypto: this makes temporary rpc server,
so in this context, io port , should be found by using socket api (bind)

but rpc server everytime is very slow.
i suggest separating into two pars
1. initialize rpc server  <- initialize_json_rpc
2. use rpc server via rpc <- json_rpc_call

client-core/src/synchronizer.rs
yihuang: It’s shared between threads, so atomic is needed.

client-core/src/synchronizer.rs
```diff
@@ -106,7 +117,18 @@ impl PollingSynchronizer {
...
             }
-        });
+        }));
+    }
+
+    /// Stop the auto synchronizer thread
+    pub fn stop(&mut self) {
+        self.should_run.store(false, Ordering::Relaxed);
+        if let Some(thread) = self.sync_thread.take() {
+            thread.join().unwrap();
```
leejw51crypto: expect

client-core/src/synchronizer.rs
yihuang: No, they are for different purposes.

client-core/src/synchronizer.rs
yihuang: It's already recorded in #1060

client-core/src/synchronizer.rs
yihuang: https://github.com/crypto-com/chain/pull/1122#discussion_r383067007

cro-clib/src/jsonrpc.rs
yihuang: That's a potential feature for the future.

cro-clib/src/jsonrpc.rs
yihuang: No server gets started, directly calls the RPC functions, not go through the socket.

cro-clib/chain.h
```diff
@@ -437,3 +437,15 @@ CroResult cro_withdraw(const char *tenermint_url_string,
                        int32_t viewkey_count,
                        uint8_t *output,
                        uint32_t *output_length);
+
+/**
+ * # Safety
+ *
+ * Should not be called with null pointers
+ */
```
leejw51crypto: add some example for storage, tendermint url

cro-clib/chain.h
```diff
@@ -437,3 +437,15 @@ CroResult cro_withdraw(const char *tenermint_url_string,
...
+CroResult jsonrpc_call(const char *storage_dir,
```
leejw51crypto: how about json_rpc

yihuang: I'll add a `cro_` prefix.

yihuang:
> @yihuang Can you add more context in this PRs description? Why is it too tedious to create C APIs (root cause) and how does this PR tries to solve that root cause?

It's an alternative approach to current c API, with clear advantages and disadvantages if works for the user:
- Much smaller code to maintain, current cro-clib already has 1591 lines of rust code with incomplete features, this approach only need 181 lines of rust code to provide complete client features, and the consistency with `client-rpc` is kept automatically.
  ```
  $ cat cro-clib/src/*.rs|wc -l
   1591
  $ git checkout issue1121
  $ cat cro-clib/src/jsonrpc.rs|wc -l
    181
  ```

Disadvantages are relative inefficiency because of JSON serialization/deserialization, which I mentions in the previous comments.

cro-clib/chain.h
yihuang: Done, added a c example in the doc.

cro-clib/chain.h
yihuang: Done

client-core/src/synchronizer.rs
yihuang: Done

leejw51crypto:
i think both approaches are fine,
for c api, all control and ownership belongs to the user.
especially for wallet

yihuang:
> may be ok -- one important thing is that for "sync", it'll be pretty bad for the caller, as that may be long-running and blocking (and you don't have any feedback when that is supposed to finish ... or whether it got stuck or something).
> I suppose it could also take a callback function pointer that could be repeatedly called with progress?

Yes, there are quite a few things to be improved, the current one is like a minimal working prototype:
- Separate the open/execute/close, so we can share opened resources between multiple calls.
- There are logging codes in `client-rpc`, for an API library, the user probably not expect logs inside.
- For the sync progress report, we can have some common stream output support, something like this?:
  ```
  let stream_id = jsonrpc_call('add wallet to sync');
  jsonrpc_stream_watch(stream_id, callback);
  ```

leejw51crypto:
callback or polling is necessary for really long request

for sync, i'm thinking similar thing with yi huang
1. give basic apis
-open, close, get block
- user owns flow control, manages sync by himeself

2. automatic state machine which has its own thread
-open, close, callback or polling
-storage all take care of the state machine

3. simple method <- current json rpc
provide some polling or callback to get progress
just like light-weight client-rpc

tomtau:
> Yes, there are quite a few things to be improved,

@yihuang @leejw51crypto so shall this PR be merged? There can be separate issues on the improvements?

leejw51crypto:
i think, we can merge this pr,
and improve by another pr

tomtau:
@yihuang can you squash the commits?

yihuang:
> @yihuang can you squash the commits?

Done

PR #1129, opened 2020-02-25
Problem (Fix #960): client-cli requires inputting authentication token

### LINKED ISSUE #960: Problem: client-cli requires inputting authentication token
That makes sense for client-rpc.
As "cli" is a human interface, this is not ideal, especially if the token can be derived from the passphrase (once in the beginning).

yihuang:
Solution:
- Support reading optional environment variables AUTH_TOKEN and PASSPHRASE

### DIFF
--- a/client-cli/src/command.rs
+++ b/client-cli/src/command.rs
@@ -53,6 +53,8 @@ ENVIRONMENT VARIABLES:
     CRYPTO_CHAIN_ID                 Chain ID of Crypto.com Chain
     CRYPTO_CLIENT_STORAGE           Storage directory (Default: `.storage`)
     CRYPTO_CLIENT_TENDERMINT        Websocket endpoint for tendermint (Default: `ws://localhost:26657/websocket`)
+    PASSPHRASE                      Passphrase to use when needed
+    AUTH_TOKEN                      Auth token to use when needed
 "#
 )]
 pub enum Command {
--- a/client-cli/src/main.rs
+++ b/client-cli/src/main.rs
@@ -60,6 +60,10 @@ pub(crate) fn chain_id() -> Option<String> {
 }
 
 pub(crate) fn ask_passphrase(message: Option<&str>) -> Result<SecUtf8> {
+    if let Ok(passphrase) = std::env::var("PASSPHRASE") {
+        return Ok(passphrase.into());
+    }
+
     ask(message.unwrap_or("Enter passphrase: "));
     password()
         .map(Into::into)
@@ -67,6 +71,10 @@ pub(crate) fn ask_passphrase(message: Option<&str>) -> Result<SecUtf8> {
 }
 
 pub(crate) fn ask_seckey(message: Option<&str>) -> Result<SecKey> {
+    if let Ok(s) = std::env::var("AUTH_TOKEN") {
+        return Ok(parse_hex_enckey(&s)?);
+    }
+
     ask(message.unwrap_or("Enter authentication token: "));
     let key = password().err_kind(ErrorKind::InvalidInput, || "Unable to read enckey")?;
     parse_hex_enckey(&key)


yihuang:
> CLI's UX seems to remain quite error-prone and counter-intuitive:
> 
> 1. a new user may not know about these env variables, as they are optional --> one will keep inputting them anyway
> 2. a new user may not know that there's "wallet auth" hidden command
> 3. by never being asked about the passphrase again, one may forget it
> 4. having passphrase in environment variables leaves it in memory or static text files for any other process to see -- perfect, no need for key loggers or anything fancy like that
> 
> if there's an existing command-line wallet or password manager that has the same or similar UX, it's ok -- otherwise, it breaks [POLA](https://en.wikipedia.org/wiki/Principle_of_least_astonishment)

4. Yes, we can remove the `PASSPHRASE` env var.
For other points, isn't the purpose of auth token is for user to no need to input them every time? they can store them somewhere safe, like mnemonic words?

tomtau:
> For other points, isn't the purpose of auth token is for user to no need to input passphrase every time? they can store them somewhere safe, like mnemonic words?

No, "auth token" in the current implementation here is a derived encryption key from the hardened passphrase hash. Its purposes were:
1) a step in untangling client-* internals, so that more keys could be derived, e.g. for separating out authorization
2) mitigate inefficiencies in client-* internals where it was recomputing the hardened hash for every single store operation.

Keeping it long-term "plain" in an environment variable has the same consequences as keeping a passphrase there.

Exposing this to a non-developer (the context here is user interface / CLI) is a leaky implementation detail.

I was talking about UX of command-line applications that handle similar sensitive materials. The typical metaphor there is "lock/unlock" (e.g. in geth or command line key managers)  -- the user enters passphrase on "unlock" (typically it's also logged in some external audit log), the application then allows authorized operations to be executed and makes sure it's only unlocked temporarily.

client-cli/src/command.rs
```diff
@@ -53,6 +53,8 @@ ENVIRONMENT VARIABLES:
     CRYPTO_CHAIN_ID                 Chain ID of Crypto.com Chain
     CRYPTO_CLIENT_STORAGE           Storage directory (Default: `.storage`)
     CRYPTO_CLIENT_TENDERMINT        Websocket endpoint for tendermint (Default: `ws://localhost:26657/websocket`)
+    PASSPHRASE                      Passphrase to use when needed
+    AUTH_TOKEN                      Auth token to use when needed
```
leejw51crypto: users will not know difference between passphrase and auth_token
add some explanation for above

leejw51crypto: also names
CRYPTO_PASSPHRASE
CRYPTO_AUTH_TOKEN

tomtau:
Just to clarify / rephrase my comment https://github.com/crypto-com/chain/pull/1129#pullrequestreview-363995456 -- the "Problem" was that client-cli requires entering this internal key *at all*.
That can either be solved by reverting back to the common UX in such command-lines tools or by other means -- e.g. not entering it at all and only taking it from the environment (if that's common in other existing similar tools)... if that's done, then it needs to be 1) more intuitive (telling the user the steps needed), 2) less error-prone and making user aware of all the risks associated with leaving this information in the shell environment

calvinlauco:
> Just to clarify / rephrase my comment [#1129 (review)](https://github.com/crypto-com/chain/pull/1129#pullrequestreview-363995456) -- the "Problem" was that client-cli requires entering this internal key _at all_.
> That can either be solved by reverting back to the common UX in such command-lines tools or by other means -- e.g. not entering it at all and only taking it from the environment (if that's common in other existing similar tools)... if that's done, then it needs to be 1) more intuitive (telling the user the steps needed), 2) less error-prone and making user aware of all the risks associated with leaving this information in the shell environment

I think it really depends on the purpose of `client-cli`. it is not uncommon to see other CLI accepts ENV variables for API keys & credentials/looking for certain config files in the home directory etc. If the CLI is intended for interacting with a real human user, then this rule seems valid to me. 

A reference to how Geth did so in the account operations
https://geth.ethereum.org/docs/interface/managing-your-accounts#non-interactive-use
They have an explicit option to provide a password file and warn about the risk of feeding passphrase directly in command line. They also mention you can always do tricky things to make it work.

I think what we can do is make sure the one who uses CLI knows about the risk of exposing passphrase/authkey all the time:
- If it is a command-line option to link a file, we add warnings to the description
- Maybe a command-line option even to directly accept enckey, but again the warnings
- No default ENV provided?, if the developer chooses to inject credentials via ENV, they can do the command line thing, but it also means they are an advanced user and should be aware of the risk they are taking, just like what they did when injecting API keys in ENV

tomtau:
> A reference to how Geth did so in the account operations
> https://geth.ethereum.org/docs/interface/managing-your-accounts#non-interactive-use
> They have an explicit option to provide a password file and warn about the risk of feeding passphrase directly in command line. They also mention you can always do tricky things to make it work.
> 

that Geth doc page with all the bold warnings and reminders about "DO NOT FORGET YOUR PASSWORD" reminded me that people were asking @lezzokafka how to recover their passphrase, so logged the issue here: https://github.com/crypto-com/chain/issues/1144

yihuang:
Closing this PR for now, the discussion can happen in the issue.

PR #1131, opened 2020-02-25
Problem:(fix #1004) client doesn't validate transactions it generates

### LINKED ISSUE #1004: Problem: client doesn't validate transactions it generates
add some context specific information in the error message
for example)
if incoins < total_outsum.unwrap() {
        let gap= total_outsum - incoins;
        let info=format!("more {} coin is necessary for fee in check_input_output_sums ", gap);
        return Err(Error::InputOutputDoNotMatch, gap);
}

linfeng-crypto:
solution: 
* Add fee check for transfer tx on client-side
 * check the staking address is jailed or not for the `deposit_bonded_stake` tx

### DIFF
--- a/client-core/src/transaction_builder/raw_transfer_transaction_builder.rs
+++ b/client-core/src/transaction_builder/raw_transfer_transaction_builder.rs
@@ -316,6 +316,7 @@ where
         self.verify_inputs()?;
         self.verify_outputs()?;
         self.verify_output_does_not_exceed_input_amount()?;
+        self.verify_fee()?;
         self.verify_input_witnesses()?;
 
         Ok(())
@@ -356,14 +357,22 @@ where
         let input_value = self.total_input_amount()?;
         let output_value = self.total_output_amount()?;
 
-        // TODO: Include fee in calculation?
         if input_value < output_value {
+            return Err(Error::new(ErrorKind::VerifyError, "Insufficient balance"));
+        }
+        Ok(())
+    }
+
+    fn verify_fee(&self) -> Result<()> {
+        let fee_expected = self.estimate_fee()?;
+        let fee_in_tx = self.fee()?;
+        if fee_in_tx < fee_expected {
+            let fee_gap = (fee_expected - fee_in_tx).unwrap();
             return Err(Error::new(
                 ErrorKind::VerifyError,
-                "Output amount exceed input amount",
+                format!("Insufficient fee, need more {:?}", fee_gap),
             ));
         }
-
         Ok(())
     }
 
@@ -588,7 +597,7 @@ mod raw_transfer_transaction_builder_tests {
 
             let err = builder.verify().unwrap_err();
             assert_eq!(err.kind(), ErrorKind::VerifyError);
-            assert_eq!(err.message(), "Output amount exceed input amount");
+            assert_eq!(err.message(), "Insufficient balance");
         }
 
         #[test]
@@ -1057,11 +1066,11 @@ mod raw_transfer_transaction_builder_tests {
 
         builder.add_input((
             TxoPointer::new(random(), 0),
-            TxOut::new(transfer_addr.clone(), Coin::new(100).unwrap()),
+            TxOut::new(transfer_addr.clone(), Coin::new(500).unwrap()),
         ));
         builder.add_input((
             TxoPointer::new(random(), 0),
-            TxOut::new(transfer_addr, Coin::new(200).unwrap()),
+            TxOut::new(transfer_addr, Coin::new(500).unwrap()),
         ));
 
         builder.add_output(TxOut::new(
--- a/client-network/src/network_ops/default_network_ops_client.rs
+++ b/client-network/src/network_ops/default_network_ops_client.rs
@@ -162,10 +162,23 @@ where
         to_address: StakedStateAddress,
         attributes: StakedStateOpAttributes,
     ) -> Result<(TxAux, TransactionPending)> {
+        // if the to_address belongs to current wallet, we do not check the state
+        let staking_addresses = self.wallet_client.staking_addresses(name, enckey)?;
+        if !staking_addresses.contains(&to_address) {
+            let staked_state = self.get_staked_state(&to_address)?;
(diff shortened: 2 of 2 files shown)


client-network/src/network_ops/default_network_ops_client.rs
```diff
@@ -166,6 +166,15 @@ where
             .iter()
             .map(|(input, _)| input.clone())
             .collect::<Vec<_>>();
+
+        let staked_state = self.get_staked_state(&to_address)?;
```
yihuang: I think `get_staked_state` return error if address not exists on chain, but should be ok to be the to_address of deposit tx.

client-core/src/transaction_builder/raw_transfer_transaction_builder.rs
```diff
@@ -356,14 +356,25 @@ where
...
                 ErrorKind::VerifyError,
                 "Output amount exceed input amount",
             ));
         }
 
+        // check fee
+        let fee_in_tx = (input_value - output_value)
+            .chain(|| (ErrorKind::VerifyError, "Invalid amount of output or input"))?;
+        let fee_expect = self.estimate_fee()?;
+        if fee_in_tx < fee_expect {
```
yihuang: How about just single check: `input_value < output_value + min_fee`?

client-network/src/network_ops/default_network_ops_client.rs
tomtau: For withdraw, it's ok.
For deposit, it could be the case that this staked state wasn't created in genesis and this is the first transaction

client-core/src/transaction_builder/raw_transfer_transaction_builder.rs
```diff
@@ -356,14 +356,25 @@ where
...
         if input_value < output_value {
             return Err(Error::new(
                 ErrorKind::VerifyError,
                 "Output amount exceed input amount",
             ));
         }
 
+        // check fee
+        let fee_in_tx = (input_value - output_value)
+            .chain(|| (ErrorKind::VerifyError, "Invalid amount of output or input"))?;
```
leejw51crypto: how about
insufficient amount ,
because intput, output can be difficult for users

client-network/src/network_ops/default_network_ops_client.rs
linfeng-crypto: When deposit, if the staking address belongs to the wallet and `staked_state` does not exist, we let it go so that the user can be a validator. In other cases, if the staking address not exists, we can prevent a user's loss.

client-core/src/transaction_builder/raw_transfer_transaction_builder.rs
linfeng-crypto: I'll fix it.

client-network/src/network_ops/default_network_ops_client.rs
yihuang: > When deposit, if the staking address belongs to the wallet and `staked_state` does not exist, we let it go so that the user can be a validator. In other cases, if the staking address not exists, we can prevent a user's loss.

I think this function is about deposit tx, so it's problematic to return error on non-exist staking address.

tomtau: >  In other cases, if the staking address not exists, we can prevent a user's loss.

it may be the case that it's really what you want to do -- e.g. @lezzokafka frequently deposits to addresses that are not managed by his wallet to distribute testnet tokens .

so rather than preventing it, it may just have an extra confirmation step

client-core/src/transaction_builder/raw_transfer_transaction_builder.rs
```diff
@@ -1057,11 +1066,11 @@ mod raw_transfer_transaction_builder_tests {
 
         builder.add_input((
             TxoPointer::new(random(), 0),
-            TxOut::new(transfer_addr.clone(), Coin::new(100).unwrap()),
+            TxOut::new(transfer_addr.clone(), Coin::new(500).unwrap()),
         ));
         builder.add_input((
             TxoPointer::new(random(), 0),
-            TxOut::new(transfer_addr, Coin::new(200).unwrap()),
+            TxOut::new(transfer_addr, Coin::new(500).unwrap()),
```
tomtau: why are the amount changes needed in tests?

tomtau: the fee was incorrect?

client-network/src/network_ops/default_network_ops_client.rs
```diff
@@ -162,10 +162,23 @@ where
         to_address: StakedStateAddress,
         attributes: StakedStateOpAttributes,
     ) -> Result<(TxAux, TransactionPending)> {
+        // if the to_address belongs to current wallet, we do not check the state
+        let staking_addresses = self.wallet_client.staking_addresses(name, enckey)?;
+        if !staking_addresses.contains(&to_address) {
+            let staked_state = self.get_staked_state(&to_address)?;
+            verify_unjailed(&staked_state).map_err(|e| {
```
leejw51crypto: how about using ".chain"?

client-core/src/transaction_builder/raw_transfer_transaction_builder.rs
```diff
@@ -356,14 +357,22 @@ where
...
 
-        // TODO: Include fee in calculation?
         if input_value < output_value {
+            return Err(Error::new(ErrorKind::VerifyError, "Insufficient balance"));
+        }
+        Ok(())
+    }
+
+    fn verify_fee(&self) -> Result<()> {
+        let fee_expect = self.estimate_fee()?;
```
leejw51crypto: how about "fee_expected"?

client-core/src/transaction_builder/raw_transfer_transaction_builder.rs
linfeng-crypto: the old one inputs value equals to outputs value,  after I add fee check, some tests will fail.

PR #1138, opened 2020-02-26
Problem (Fix #818): outdated README

### LINKED ISSUE #818: Problem: outdated README
e.g. one may need clang/g++

yihuang:
Solution:
- Update the build instructions in README
- Rename docker/Dockerfile.new to Dockerfile as default docker file
- Remove kcov part since it's not used and waste build time.


tomtau:
> Perhaps it will be good to have a separate task describing some of the make command supported by the Makefile and their usage.
> 
> Right now the Makefile is mentioned in https://crypto-com.github.io/getting-started/#build-binary-and-docker-images but not commands are used/missing.

perhaps just mention there's `make help` that prints out the options?

calvinlauco:
> > Perhaps it will be good to have a separate task describing some of the make command supported by the Makefile and their usage.
> > Right now the Makefile is mentioned in https://crypto-com.github.io/getting-started/#build-binary-and-docker-images but not commands are used/missing.
> 
> perhaps just mention there's `make help` that prints out the options?

This would also work.

yihuang:
> > Perhaps it will be good to have a separate task describing some of the make command supported by the Makefile and their usage.
> > Right now the Makefile is mentioned in https://crypto-com.github.io/getting-started/#build-binary-and-docker-images but not commands are used/missing.
> 
> perhaps just mention there's `make help` that prints out the options?

Done

PR #1150, opened 2020-02-27
Problem:(cro-1140) no android support for c-bindings

leejw51:
Solution:  add android support
add lib for example


leejw51crypto:
retry

Issue #1128, opened 2020-02-25
Problem: Makefile is trying to compile tx-validation-ap


calvinaco:
```
docker run -i --rm \
	-v /home/ubuntu/.cargo/git:/root/.cargo/git \
	-v /home/ubuntu/.cargo/registry:/root/.cargo/registry \
	-v `pwd`:/chain \
	--env RUSTFLAGS=-Ctarget-feature=+aes,+sse2,+sse4.1,+ssse3 \
	--workdir=/chain \
	cryptocom/chain:latest \
	bash -c '. /root/.docker_bashrc && \
	echo "========  build dev-utils   =========" && \
	cargo build --bin dev-utils && \
	echo "========  build chain-abci   =========" && \
	cd chain-abci && cargo build && \
	echo "========  build client-cli   =========" && \
	cd ../client-cli && cargo build&& \
	echo "========  build client-rpc   =========" && \
	cd ../client-rpc && cargo build'
========  build dev-utils   =========
    Finished dev [unoptimized + debuginfo] target(s) in 0.25s
========  build chain-abci   =========
    Finished dev [unoptimized + debuginfo] target(s) in 0.14s
========  build client-cli   =========
    Finished dev [unoptimized + debuginfo] target(s) in 0.17s
========  build client-rpc   =========
    Finished dev [unoptimized + debuginfo] target(s) in 0.15s
compile sgx query
    Finished dev [unoptimized + debuginfo] target(s) in 0.12s
make: Entering directory '/chain/chain-tx-enclave/tx-query'
    Finished dev [unoptimized + debuginfo] target(s) in 0.10s
CARGO => ../../target/debug/libtx_query_enclave.a
make: Leaving directory '/chain/chain-tx-enclave/tx-query'
compile sgx validation
error: package ID specification `tx-validation-app` matched no packages
Makefile:178: recipe for target 'build-sgx-validation' failed
make: *** [build-sgx-validation] Error 101
```

tomtau:
docker/build.sh

PR #1152, opened 2020-02-27
Problem: Makefile is trying to compile tx-validation-ap #1128

allthatjazzleo:
Solution: remove tx-validation-app component from makefile and dockerfile
fix  enclave.socket issue for run-abci and run-sgx-query to share same unix domain socket
replace run_tx_validation.sh with ci-scripts/run_chain_abci.sh to make sure abci can replace the deprecated tx-validation-app 
fix other minor issue like setting $PID in shell script

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,12 +8,9 @@ prefix ?=
 sgx_mode ?= HW
 build_mode ?= debug
 MAKE_CMD = make
-# sgx_device can be /dev/sgx or /dev/isgx
-sgx_device ?= ${SGX_DEVICE}
 
-ifeq ($(sgx_device)x, x)
-	sgx_device ?= /dev/isgx
-endif
+# SGX_DEVICE can be /dev/sgx or /dev/isgx
+SGX_DEVICE ?= /dev/sgx
 
 ifeq ($(build_mode), release)
 	CARGO_BUILD_CMD = cargo build --release
@@ -89,8 +86,8 @@ IMAGE_RUST             = cryptocom/chain
 IMAGE_TENDERMINT       = tendermint/tendermint:v0.32.8
 DOCKER_FILE            = docker/Dockerfile
 DOCKER_FILE_RELEASE    = docker/Dockerfile.release
-ITEMS_START            = sgx-validation sgx-query chain-abci tendermint client-rpc
-ITEMS_STOP             = client-rpc tendermint chain-abci sgx-query sgx-validation
+ITEMS_START            = sgx-query chain-abci tendermint client-rpc
+ITEMS_STOP             = client-rpc tendermint chain-abci sgx-query
 
 create-path:
 	mkdir -p ${HOME}/.cargo/{git,registry}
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -201,20 +206,21 @@ rm-network:
...
-	--name $(prefix)sgx-validation \
-	-e SGX_MODE=$(SGX_MODE) \
-	-e NETWORK_ID=$(NETWORK_ID) \
-	-e RUST_LOG=$(RUST_LOG) \
-	--device $(sgx_device) \
-	-v  $(data_path)/enclave-storage:/crypto-chain/enclave-storage \
-	--workdir=/usr/local/bin \
-	$(IMAGE):$(TAG) \
-	bash ./run_tx_validation.sh
+# deprecated and replaced by abci
```
tomtau: I think this could just be removed?

docker/Dockerfile
```diff
@@ -19,12 +19,13 @@ COPY ./target/${BUILD_MODE}/dev-utils /usr/local/bin/
 # copy sgx binary file
 COPY ./target/${BUILD_MODE}/tx-query-app /usr/local/bin
 COPY ./target/${BUILD_MODE}/tx_query_enclave.signed.so /usr/local/bin
-COPY ./target/${BUILD_MODE}/tx-validation-app /usr/local/bin
+# COPY ./target/${BUILD_MODE}/tx-validation-app /usr/local/bin # deprecated and replaced by chain-abci
```
tomtau: could be deleted?

docker/Dockerfile
```diff
@@ -19,12 +19,13 @@ COPY ./target/${BUILD_MODE}/dev-utils /usr/local/bin/
...
 COPY ./target/${BUILD_MODE}/tx_validation_enclave.signed.so /usr/local/bin
 # copy scripts
 COPY ./docker/wait-for-it.sh /usr/local/bin
 COPY ci-scripts/run_tx_query.sh /usr/local/bin
-COPY ci-scripts/run_tx_validation.sh /usr/local/bin
+# COPY ci-scripts/run_tx_validation.sh /usr/local/bin # deprecated and replaced by chain-abci
+COPY ci-scripts/run_chain_abci.sh /usr/local/bin
 
 # install sgx enclave pws
 ARG SGX_ENCLAVE=https://download.01.org/intel-sgx/sgx-linux/2.7.1/distro/ubuntu18.04-server/libsgx-enclave-common_2.7.101.3-bionic1_amd64.deb
```
tomtau: perhaps a TODO comment to update PSW to 2.8? it's been split to more packages

Makefile
```diff
@@ -141,22 +142,27 @@ endif
...
-		bash -c '. /root/.docker_bashrc && \
-		echo "========  build dev-utils   =========" && \
-		${CARGO_BUILD_CMD} --bin dev-utils && \
-		echo "========  build chain-abci   =========" && \
-		cd chain-abci && $(CARGO_BUILD_CMD_ABCI) && \
-		echo "========  build client-cli   =========" && \
-		cd ../client-cli && $(CARGO_BUILD_CMD_CLI)&& \
-		echo "========  build client-rpc   =========" && \
-		cd ../client-rpc && $(CARGO_BUILD_CMD_CLI)'
+	@if [ -e "./target/${BUILD_MODE}/client-cli" ] && [ -e "./target/${BUILD_MODE}/chain-abci" ] && [ -e "./target/${BUILD_MODE}/client-rpc" ] && [ -e "./target/${BUILD_MODE}/dev-utils" ]; then \
```
yihuang: I think it's quite dangerous to use the same target directory for `make build` and `cargo build`. If the developer uses them both on different platforms, the resulting binary might be undefined?
But this issue is out of the scope of this PR.

tomtau: true; this may not be reproducible -- so even though one person follows the same instructions on one platform, it may not work.

Makefile
```diff
@@ -8,11 +8,12 @@ prefix ?=
...
-# sgx_device can be /dev/sgx or /dev/isgx
-sgx_device ?= ${SGX_DEVICE}
 
-ifeq ($(sgx_device)x, x)
-	sgx_device ?= /dev/isgx
+# sgx_device can be /dev/sgx or /dev/isgx
+ifeq (${SGX_DEVICE}x, x)
+	sgx_device = /dev/sgx
+else
+	sgx_device = ${SGX_DEVICE}
```
yihuang:
```suggestion
# sgx_device can be /dev/sgx or /dev/isgx
SGX_DEVICE ?= /dev/isgx
```
And use `SGX_DEVICE` directly?

ci-scripts/run_chain_abci.sh
```diff
@@ -8,10 +8,17 @@ if [ x"${SGX_MODE}" == "xHW" ]; then
   LD_LIBRARY_PATH=/opt/intel/libsgx-enclave-common/aesm /opt/intel/libsgx-enclave-common/aesm/aesm_service &
 
   echo "[aesm_service] Running in background ..."
-  # Wait for aesm_service to initialize
-  sleep 4
```
yihuang: Not sure is the original sleep added for a reason, @linfeng-crypto ?

docker/Dockerfile
```diff
@@ -19,12 +19,13 @@ COPY ./target/${BUILD_MODE}/dev-utils /usr/local/bin/
...
 COPY ./target/${BUILD_MODE}/tx-query-app /usr/local/bin
 COPY ./target/${BUILD_MODE}/tx_query_enclave.signed.so /usr/local/bin
-COPY ./target/${BUILD_MODE}/tx-validation-app /usr/local/bin
+# COPY ./target/${BUILD_MODE}/tx-validation-app /usr/local/bin # deprecated and replaced by chain-abci
 COPY ./target/${BUILD_MODE}/tx_validation_enclave.signed.so /usr/local/bin
 # copy scripts
 COPY ./docker/wait-for-it.sh /usr/local/bin
 COPY ci-scripts/run_tx_query.sh /usr/local/bin
-COPY ci-scripts/run_tx_validation.sh /usr/local/bin
+# COPY ci-scripts/run_tx_validation.sh /usr/local/bin # deprecated and replaced by chain-abci
```
yihuang:
```suggestion
```

Makefile
leejw51crypto: target folder is kindof mutex, simultaneous compiling is not performed.

Makefile
```diff
@@ -201,20 +206,21 @@ rm-network:
...
+# 	--restart=always \
+# 	--name $(prefix)sgx-validation \
+# 	-e SGX_MODE=$(SGX_MODE) \
+# 	-e NETWORK_ID=$(NETWORK_ID) \
+# 	-e RUST_LOG=$(RUST_LOG) \
+# 	--device $(sgx_device) \
+# 	-v $(data_path)/enclave-storage:/crypto-chain/enclave-storage \
+# 	--workdir=/usr/local/bin \
+# 	$(IMAGE):$(TAG) \
+# 	bash ./run_tx_validation.sh
```
leejw51crypto: if this is reference, you can make markdown file for document and future reference

Makefile
```diff
@@ -230,7 +236,8 @@ run-sgx-query:
 		-e NETWORK_ID=$(NETWORK_ID) \
 		-e SPID=${SPID} \
 		-e IAS_API_KEY=${IAS_API_KEY} \
-		-e TX_VALIDATION_CONN=tcp://$(prefix)sgx-validation:26650 \
+		-e TX_VALIDATION_CONN=ipc:///root/sockets/enclave.socket \
```
leejw51crypto: TX_VALIDATION_CON or
TX_VALIDATION_CONNECTION?

Makefile
calvinlauco: I think we can just delete it and use Git to recover it in the future if needed.

ci-scripts/run_chain_abci.sh
calvinlauco: This has existed quite a while but the sleep time has been increased since then. Perhaps we should use other ways to make sure aesm service is ready.
https://github.com/crypto-com/chain-tx-enclave/blob/8e1ecad29eed5cf3ac702d260bfdf67b0f186eb8/tx-validation/entrypoint.sh

yihuang: But in the drone integration tests pipeline, I never added any sleep, and have no problem so far.

PR #1157, opened 2020-03-02
Problem (Fix #1141): revised reward functionality and validator interactions not implemented

### LINKED ISSUE #1141: Problem: revised reward functionality not implemented
Ref:
https://github.com/crypto-com/chain-docs/issues/82
https://github.com/crypto-com/chain-docs/issues/76
https://github.com/crypto-com/chain-docs/pull/75

yihuang:
Solution:
- Update reward implementation

Marked the unit test `check_wallet_syncer_app_hash_on_multiple_tx` as ignored, recorded as #1168 .

### DIFF
--- a/chain-abci/src/app/app_init/validator_state.rs
+++ b/chain-abci/src/app/app_init/validator_state.rs
@@ -1,4 +1,3 @@
-use crate::app::rewards::RewardsDistribution;
 use crate::app::ChainNodeState;
 use crate::liveness::LivenessTracker;
 use crate::punishment::ValidatorPunishment;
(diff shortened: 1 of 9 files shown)


tomtau:
@yihuang wasn't @devashishdxt implementing it: https://github.com/crypto-com/chain/issues/1141 ?
it's ok, but the initial point here was that other person implements other person's spec, so it can be 1) code review can be done by the doc author, 2) checked that the doc is fully clear to / implementable by someone else than its author

yihuang:
> @yihuang wasn't @devashishdxt implementing it: #1141 ?
> it's ok, but the initial point here was that other person implements other person's spec, so it can be 1) code review can be done by the doc author, 2) checked that the doc is fully clear to / implementable by someone else than its author

I just find we don't actually have a language both precise and succinct to write the spec, rewarding logic straight forward, if you write down all the computation details in plain English precisely enough, it would be much longer than the actual rust code itself.

devashishdxt:
> > @yihuang wasn't @devashishdxt implementing it: #1141 ?
> > it's ok, but the initial point here was that other person implements other person's spec, so it can be 1) code review can be done by the doc author, 2) checked that the doc is fully clear to / implementable by someone else than its author
> 
> I just find we don't actually have a language both precise and succinct to write the spec, rewarding logic straight forward, if you write down all the computation details(mostly data structure conversions, overflow considerations) in plain English precisely enough, it would be longer than the actual rust code itself.

I don't think that the specs will get longer than the code (and it is ok even if they do). The main reason for having a specification is:
- Specification defines the behavior of code.
- Specification rarely changes.
- Code, in terms of design and implementation, may change but should always follow the specification.

tomtau:
will temporarily close PR this until https://github.com/crypto-com/chain-docs/pull/94 is finalized -- after that, @devashishdxt can take over and have this for reference

PR #1171, opened 2020-03-03
Problem: (cro-1133) no way to get progress from sync in `jsonrpc_call` C API

leejw51:
Solution: add get progress callback


leejw51crypto:
try again

PR #1172, opened 2020-03-03

leejw51crypto:
Solution:add get progress in json-rpc call


leejw51crypto:
try again

PR #1173, opened 2020-03-03

leejw51crypto:
Solution:add get progress callback to jsonrpc

### DIFF
--- a/client-core/src/wallet/syncer.rs
+++ b/client-core/src/wallet/syncer.rs
@@ -106,6 +106,12 @@ pub struct SyncerConfig<S: SecureStorage, C: Client> {
     block_height_ensure: u64,
 }
 
+#[derive(Clone)]
+pub struct SyncCallback {
+    pub user_data: u64,
+    pub user_callback: extern "C" fn(u64, u64, u64, u64) -> i32,
+}
+
 /// Wallet Syncer
 #[derive(Clone)]
 pub struct WalletSyncer<S: SecureStorage, C: Client, D: TxDecryptor> {
@@ -121,6 +127,7 @@ pub struct WalletSyncer<S: SecureStorage, C: Client, D: TxDecryptor> {
     decryptor: D,
     name: String,
     enckey: SecKey,
+    progress_callback: Option<SyncCallback>,
 }
 
 impl<S, C, D> WalletSyncer<S, C, D>
@@ -136,6 +143,7 @@ where
         progress_reporter: Option<Sender<ProgressReport>>,
         name: String,
         enckey: SecKey,
+        progress_callback: Option<SyncCallback>,
     ) -> WalletSyncer<S, C, D> {
         Self {
             storage: config.storage,
@@ -147,6 +155,7 @@ where
             enable_fast_forward: config.enable_fast_forward,
             batch_size: config.batch_size,
             block_height_ensure: config.block_height_ensure,
+            progress_callback,
         }
     }
 
@@ -187,6 +196,7 @@ where
         progress_reporter: Option<Sender<ProgressReport>>,
         name: String,
         enckey: SecKey,
+        progress_callback: Option<SyncCallback>,
     ) -> Result<WalletSyncer<S, C, TxObfuscationDecryptor<O>>>
     where
         O: TransactionObfuscation,
@@ -207,6 +217,7 @@ where
             progress_reporter,
             name,
             enckey,
+            progress_callback,
         ))
     }
 }
@@ -219,6 +230,7 @@ struct WalletSyncerImpl<'a, S: SecureStorage, C: Client, D: TxDecryptor> {
     wallet: Wallet,
     sync_state: SyncState,
     wallet_state: WalletState,
+    progress_callback: Option<SyncCallback>,
 }
 
 impl<'a, S: SecureStorage, C: Client, D: TxDecryptor> WalletSyncerImpl<'a, S, C, D> {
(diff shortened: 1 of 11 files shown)


yihuang:
I think you can use the existing `progress_reporter` thing, rather than adding a callback into the `WalletSyncer`.

leejw51crypto:
yes, that's possible,
but i don't need channel for progress, because it's c.

cro-clib/examples/rpc.c
```diff
@@ -0,0 +1,55 @@
...
+#include <inttypes.h>
+#include <stdlib.h>
+#include <string.h>
+#include <wchar.h>
+#include <assert.h>
+#include "../chain-core.h"
+#include "../chain.h"
+
+// rate: 0.0 ~ 100.0
+int32_t  progress(float rate)
```
yihuang: The callback probably needs some context to be useful.

client-core/src/wallet/syncer.rs
```diff
@@ -106,6 +106,8 @@ pub struct SyncerConfig<S: SecureStorage, C: Client> {
     block_height_ensure: u64,
 }
 
+pub type SyncCallback = extern "C" fn(f32) -> i32;
```
yihuang: The callback probably should be a closure to be useful, which can encapsulate the context for the c function.
Also if we choose callback to report the progress, ideally, should deprecate the `progress_reporter`, which can be implemented based on the callback feature.

leejw51crypto: yes, but this callback is "C" function

cro-clib/examples/rpc.c
leejw51crypto: some kind of struct would be fine?

client-core/src/wallet/syncer.rs
yihuang: I think a c function together with a context pointer can be wrapped in a rust closure?

leejw51crypto: that's possible

yihuang: > that's possible
> but i don't need additonal thread for this.
> what do you think?
> 
> if sync is persistent, current progress updater is fine.
> but for c-calling, it's just one thread for both of producer, consumer
> 
> thread context is very important in c, because other c-source code will be thread-unsafe.

No thread, just callbacks. 
By context, I mean the c callback function needs something to access the state, otherwise, all it have is an integer argument, what can it do other than printing it out. It needs something to connect back to the system.

leejw51crypto: got it,
callback is pure event info, just plain c primitive type would be fine.
don't need any opaque pointer

because it is just giving information from core to sdk-user.

cro-clib/examples/rpc.c
yihuang: I suggest something like this:
```
int progress(int current, int total, void *data);
```
- Change rate to current and total, user has more freedom to decide how to deal with the progress information.
- Add a data pointer, whatever user passes into the api, we pass it to the callback.
- Use the return value to support aborting?

leejw51crypto: ok

leejw51crypto:
i'm changing progress api

client-core/src/wallet/syncer.rs
yihuang: Use closure in logic code:
```
callback: F,
where F: Fn(int, int) -> bool
```
The ffi detail should be wrapped in the api library.

Ideally, we should implement progress_report with callback, but that can be done in another PR I think.

client-core/src/wallet/syncer.rs
```diff
@@ -106,6 +106,12 @@ pub struct SyncerConfig<S: SecureStorage, C: Client> {
     block_height_ensure: u64,
 }
 
+#[derive(Clone)]
+pub struct SyncCallback {
+    pub user_data: u64,
+    pub user_callback: extern "C" fn(u64, u64, u64, u64) -> i32,
+}
```
yihuang: You should use a rust closure here, wrap up the c ffi stuff in the c lib.

leejw51crypto: i tried, some strange issues, 
so i changed back to function again.

leejw51crypto:
with discussion with yi,
i'll do as below
1. change u64 -> void*
2. use closure

client-core/src/wallet/syncer.rs
tomtau: @leejw51crypto do you still have the code? you can possibly share on slack or via vscode live

leejw51crypto: i talked with yi,
modifying now
1. adding closure
2. use void* instead of u64
not to use ffi code in core

leejw51crypto:
make not to call ffi-c direclty.
rebase and will modify based on huang's modification

client-core/src/wallet/syncer.rs
```diff
@@ -384,6 +410,17 @@ impl<'a, S: SecureStorage, C: Client, D: TxDecryptor> WalletSyncerImpl<'a, S, C,
 
                 let block = FilteredBlock::from_block(&self.wallet, &block, &block_result)?;
                 self.update_progress(block.block_height);
+
+                if let Some(delegator) = &self.progress_callback {
```
leejw51crypto: this part will be changed,
using sync's callback

leejw51crypto:
resubmit

PR #1182, opened 2020-03-04
Problem: repeated db open/close and thread spawning `jsonrpc_call` C binding#1132

leejw51crypto:
Solution: add context, divide into 3 functions
add comments


leejw51crypto:
try again

Issue #1133, opened 2020-02-25
Problem: no way to get progress from sync in `jsonrpc_call` C API


tomtau:
could be some stream watch function? @yihuang https://github.com/crypto-com/chain/pull/1122#issuecomment-590678357

leejw51crypto:
started

leejw51crypto:
done,
now writing testing code

leejw51crypto:
pr submitted

leejw51crypto:
in code-review

PR #1183, opened 2020-03-04
Problem: repeated db open/close and thread spawning `jsonrpc_call` C …

leejw51crypto:
Solution: add context


leejw51crypto:
i'll re-modify after #1133

leejw51crypto:
will do it later

PR #1185, opened 2020-03-04
Problem(fix #914) Unable to restore wallet with the same mnemonic words

### LINKED ISSUE #914: Problem: Unable to restore wallet with the same seed (or mnemonic words) as existing wallet.

linfeng-crypto:
solution:
* store the (wallet_name, wallet's private_key) in `KeySerivce`
* move the (public_key, private_key) pair (used for transfer add and staking address) in `Wallet`
* extend the `root_hash` with the `wallet_name` as  storage_key to store the `multi_sig_address`
* add unit test

### DIFF
--- a/client-core/src/service/key_service.rs
+++ b/client-core/src/service/key_service.rs
@@ -1,11 +1,11 @@
 use zeroize::Zeroize;
 
 use client_common::Result;
-use client_common::{ErrorKind, PrivateKey, PublicKey, ResultExt, SecKey, SecureStorage, Storage};
+use client_common::{ErrorKind, PrivateKey, ResultExt, SecKey, SecureStorage, Storage};
 
 const KEYSPACE: &str = "core_key";
 
-/// Maintains mapping `public-key -> private-key`
+/// Maintains mapping `wallet-name -> private-key`
 #[derive(Debug, Default, Clone)]
 pub struct KeyService<T: SecureStorage> {
     storage: T,
@@ -21,32 +21,32 @@ where
         KeyService { storage }
     }
 
-    /// Adds a new public-private keypair to storage
-    pub fn add_keypair(
+    /// Adds a new wallet_name-private keypair to storage
+    pub fn add_wallet_private_key(
         &self,
+        wallet_name: &str,
         private_key: &PrivateKey,
-        public_key: &PublicKey,
         enckey: &SecKey,
     ) -> Result<()> {
         self.storage
             .set_secure(
                 KEYSPACE,
-                public_key.serialize(),
+                wallet_name.as_bytes(),
                 private_key.serialize(),
                 enckey,
             )
             .map(|_| ())
     }
 
-    /// Retrieves private key corresponding to given public key
-    pub fn private_key(
+    /// Retrieves private key corresponding to given wallet name
+    pub fn wallet_private_key(
         &self,
-        public_key: &PublicKey,
+        wallet_name: &str,
         enckey: &SecKey,
     ) -> Result<Option<PrivateKey>> {
         let private_key_bytes =
             self.storage
-                .get_secure(KEYSPACE, public_key.serialize(), enckey)?;
+                .get_secure(KEYSPACE, wallet_name.as_bytes(), enckey)?;
 
         private_key_bytes
             .map(|mut private_key_bytes| {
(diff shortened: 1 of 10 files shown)


client-core/src/service/key_service.rs
```diff
@@ -21,32 +21,32 @@ where
         KeyService { storage }
     }
 
-    /// Adds a new public-private keypair to storage
-    pub fn add_keypair(
+    /// Adds a new wallet_name-private keypair to storage
```
yihuang: I think it needs to store multiple key pairs for one wallet.

client-core/src/service/key_service.rs
```diff
@@ -21,32 +21,32 @@ where
...
+    pub fn add_wallet_private_key(
```
leejw51crypto: some explanation?
what's this privatekey?

client-core/src/service/key_service.rs
```diff
@@ -21,32 +21,32 @@ where
...
+                wallet_name.as_bytes(),
                 private_key.serialize(),
                 enckey,
             )
             .map(|_| ())
     }
 
-    /// Retrieves private key corresponding to given public key
-    pub fn private_key(
+    /// Retrieves private key corresponding to given wallet name
```
leejw51crypto: one wallet, one privatekey,
1:1 right?

client-core/src/service/key_service.rs
```diff
@@ -21,32 +21,32 @@ where
...
+        wallet_name: &str,
         private_key: &PrivateKey,
-        public_key: &PublicKey,
         enckey: &SecKey,
     ) -> Result<()> {
         self.storage
             .set_secure(
                 KEYSPACE,
-                public_key.serialize(),
+                wallet_name.as_bytes(),
```
leejw51crypto: why public-key is not necessary?

client-core/src/service/key_service.rs
```diff
@@ -21,32 +21,32 @@ where
...
             .map(|_| ())
     }
 
-    /// Retrieves private key corresponding to given public key
-    pub fn private_key(
+    /// Retrieves private key corresponding to given wallet name
+    pub fn wallet_private_key(
         &self,
-        public_key: &PublicKey,
+        wallet_name: &str,
```
leejw51crypto: also publickey

leejw51crypto: consider wallet_name + public_key

client-core/src/service/key_service.rs
linfeng-crypto: the private key is the one when a wallet created and the view key comes from this private key

client-core/src/service/key_service.rs
linfeng-crypto: because the public-key is the view key, already stored in `Wallet`

client-core/src/service/key_service.rs
linfeng-crypto: yes, the private key is the one when the wallet create, so there is only one private key.

client-core/src/service/key_service.rs
linfeng-crypto: mulitiple key pairs (for transfer addresses and staking address ) are stored in Wallet.

client-core/src/service/key_service.rs
leejw51crypto: if it's just one wallet, set_wallet_viewkey_private?

linfeng-crypto:
> from a quick look, it seems all right.
> since this will make it incompatible with old client data, could you add an entry to breaking changes (of unreleased 0.4.0) in CHANGELOG.md ?

Ok

client-core/src/service/key_service.rs
yihuang: So I think `KeyService` can be removed now since key pairs are in `Wallet`?

linfeng-crypto: but the private key is not stored in Wallet, how do you think we move all the associated items (like key pairs, private key, multisign_address) into the `Wallet`, because all the `items`  can be accessed with `enckey`(+ wallet_name) just like `Wallet`, we currently stored them in different place, if we want to get them, we first get the `Wallet`, then use the key in the wallet to get the value, it's not convenient.

PR #1191, opened 2020-03-05

leejw51crypto:
## Solution: add get progress callback
1. u64: for handle of json rpc context , * const c_void is not allowed in rust
    using handle as integer is recommended rather than putting pointer directly
2. callback in same thread, so can use ui in callback thread 
3. used Arc, Mutex for thread safety in calling c-callbacks 
## sync example 
```
   95.04  progress current=89117  start= 89002 ~ end=89123   user= 2020
   95.87  progress current=89118  start= 89002 ~ end=89123   user= 2020
   96.69  progress current=89119  start= 89002 ~ end=89123   user= 2020
   97.52  progress current=89120  start= 89002 ~ end=89123   user= 2020
   98.35  progress current=89121  start= 89002 ~ end=89123   user= 2020
   99.17  progress current=89122  start= 89002 ~ end=89123   user= 2020
  100.00  progress current=89123  start= 89002 ~ end=89123   user= 2020
  100.00  progress current=89123  start= 89002 ~ end=89123   user= 2020
response: {"jsonrpc":"2.0","result":null,"id":1}
```

### DIFF
--- a/client-rpc/src/rpc/sync_rpc.rs
+++ b/client-rpc/src/rpc/sync_rpc.rs
@@ -5,9 +5,23 @@ use crate::server::to_rpc_error;
 use client_common::tendermint::Client;
 use client_common::Storage;
 use client_core::synchronizer::PollingSynchronizer;
+use client_core::wallet::syncer::ProgressReport;
 use client_core::wallet::syncer::{ObfuscationSyncerConfig, WalletSyncer};
 use client_core::wallet::WalletRequest;
 use client_core::TransactionObfuscation;
+use std::sync::Arc;
+use std::sync::Mutex;
+
+pub trait CBindingCallback: Send + Sync {
+    fn progress(&self, current: u64, start: u64, end: u64) -> i32;
+    fn set_user(&mut self, user: u64);
+    fn get_user(&self) -> u64;
+}
+
+#[derive(Clone)]
+pub struct CBindingCore {
+    pub data: Arc<Mutex<dyn CBindingCallback>>,
+}
 
 #[rpc]
 pub trait SyncRpc: Send + Sync {
@@ -32,6 +46,7 @@ where
 {
     config: ObfuscationSyncerConfig<S, C, O>,
     polling_synchronizer: PollingSynchronizer,
+    progress_callback: Option<CBindingCore>,
 }
 
 impl<S, C, O> SyncRpc for SyncRpcImpl<S, C, O>
@@ -70,13 +85,17 @@ where
     C: Client + 'static,
     O: TransactionObfuscation + 'static,
 {
-    pub fn new(config: ObfuscationSyncerConfig<S, C, O>) -> Self {
+    pub fn new(
+        config: ObfuscationSyncerConfig<S, C, O>,
+        progress_callback: Option<CBindingCore>,
+    ) -> Self {
         let mut polling_synchronizer = PollingSynchronizer::default();
         polling_synchronizer.spawn(config.clone());
 
         SyncRpcImpl {
             config,
             polling_synchronizer,
+            progress_callback,
         }
     }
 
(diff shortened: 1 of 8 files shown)


leejw51crypto:
thanks @yihuang  for modifying callback api of sync

client-rpc/src/rpc/sync_rpc.rs
```diff
@@ -5,9 +5,26 @@ use crate::server::to_rpc_error;
...
 use client_common::Storage;
 use client_core::synchronizer::PollingSynchronizer;
+use client_core::wallet::syncer::ProgressReport;
 use client_core::wallet::syncer::{ObfuscationSyncerConfig, WalletSyncer};
 use client_core::wallet::WalletRequest;
 use client_core::TransactionObfuscation;
+use std::sync::Arc;
+use std::sync::Mutex;
+
+pub trait CBidingCallback: Send + Sync {
```
yihuang:
```suggestion
pub trait CBindingCallback: Send + Sync {
```

leejw51crypto: fixed

leejw51crypto:
callback is protected by Arc, Mutex
also designed to work in shared circumstances.

code works fine in testing

also for polling, polling is in-effective but safe, but not necessary for this.

leejw51crypto:
i tried,
rust doesn't allow using const* c_void

tomtau:
Can you post the code snippet with std::ffi::void on Slack?

If polling is simpler and works fine on different platforms, it's preferred -- especially with regards to other later goals, like reusing the DB etc. created context

leejw51crypto:
user_data: u64, which caller makes to distinguish request

leejw51crypto:
currently, in c, there is no thread,

to make polling,
calls should be asynchronous as well.
that means that jsoncall will have thread manager inside.

polling itself i think good feature,
so how about as another PR?

leejw51crypto:
also current approach, if multiple calls in multi-threaded,
user_data is the handle,

it will diverge corresponding instance in one callback

cro-clib/src/jsonrpc.rs
```diff
@@ -108,11 +110,41 @@ fn make_syncer_config(
...
+    user_data: u64,
+}
+
+impl CBindingCallback for CBindingData {
+    fn set_user(&mut self, user: u64) {
+        self.user_data = user;
+    }
+    fn get_user(&self) -> u64 {
+        self.user_data
+    }
```
tomtau: is this used?

client-rpc/src/rpc/sync_rpc.rs
```diff
@@ -5,9 +5,23 @@ use crate::server::to_rpc_error;
...
 use client_core::wallet::syncer::{ObfuscationSyncerConfig, WalletSyncer};
 use client_core::wallet::WalletRequest;
 use client_core::TransactionObfuscation;
+use std::sync::Arc;
+use std::sync::Mutex;
+
+pub trait CBindingCallback: Send + Sync {
+    fn progress(&self, current: u64, start: u64, end: u64) -> i32;
+    fn set_user(&mut self, user: u64);
+    fn get_user(&self) -> u64;
```
tomtau: are these needed?

cro-clib/src/jsonrpc.rs
```diff
@@ -108,11 +110,41 @@ fn make_syncer_config(
     ))
 }
 
+use std::sync::Arc;
+use std::sync::Mutex;
+
+#[derive(Clone)]
+struct CBindingData {
+    progress_callback: ProgressCallback,
+    user_data: u64,
```
tomtau: probably add some safety warning about the unsafe cast (*void -> u64) that this should never be accessed from within rust and should be only be passed back to the progress callback

cro-clib/src/jsonrpc.rs
leejw51: in example,
it's used to tell which context it's on
```
  CroJsonRpcPtr rpc= NULL;
    cro_create_jsonrpc(&rpc, ".storage", "ws://localhost:26657/websocket", 0xab, &progress);
    cro_run_jsonrpc(rpc, wallet_req, buf, sizeof(buf), user);
    printf("response: %s\n", buf);    

    cro_run_jsonrpc(rpc, req, buf, sizeof(buf),user);
    printf("response: %s\n", buf);
```

client-rpc/src/rpc/sync_rpc.rs
leejw51: because all calls are synchronous,
technically, not necessary,
but it's handy to give context info in callback.

cro-clib/src/jsonrpc.rs
leejw51crypto: ok

cro-clib/src/jsonrpc.rs
leejw51crypto: in example,
it's used to tell which context it's on
```
  CroJsonRpcPtr rpc= NULL;
    cro_create_jsonrpc(&rpc, ".storage", "ws://localhost:26657/websocket", 0xab, &progress);
    cro_run_jsonrpc(rpc, wallet_req, buf, sizeof(buf), user);
    printf("response: %s\n", buf);    

    cro_run_jsonrpc(rpc, req, buf, sizeof(buf),user);
    printf("response: %s\n", buf);
```

client-rpc/src/rpc/sync_rpc.rs
leejw51crypto: because all calls are synchronous,
technically, not necessary,
but it's handy to give context info in callback.

tomtau: I mean `set_user` and `get_user` aren't called anywhere?

leejw51crypto: set_user is called in jsonrpc.rs of cro_clib,
get_user is not called in rust, can be removed

tomtau: can you point to the line where `set_user` is called?

leejw51crypto: https://github.com/leejw51crypto/chain/blob/cro-1132g/cro-clib/src/jsonrpc.rs
line 328
it's another pr

tomtau: I see in this pr https://github.com/crypto-com/chain/pull/1196/files#diff-cb49bbc10a53700d54f20f45422fcc13R328

leejw51crypto:
ok, i'll modify now

leejw51crypto:
modified and fixed some clippy issue

tomtau:
bors retry

leejw51crypto:
i'll check

leejw51crypto:
bors

leejw51:
bors

tomtau:
bors try

PR #1196, opened 2020-03-06
Problem:(cro-1132) repeated db open/close and thread spawning `jsonrpc_call` C binding#1132

leejw51crypto:
## Solution ##
1.  add context
2. for different request, you can give different user_data as void*
```
actual code is  called in-directly via `handle_request_sync`, so, callback is stored at trait,
and called afterwards

this is for simplicity of reusing rpc-server.
also json run code is all synchronous.
```
## flow ##
rpc call1 end
rpc call2 end

## to do ##
for autosync, call is sync,
but actual syncing is endless. 
so, other thread checking via polling  
it's will be added soon.

### DIFF
--- a/cro-clib/examples/rpc.c
+++ b/cro-clib/examples/rpc.c
@@ -54,11 +54,40 @@ void sync()
         printf("error: %s\n", buf);
     }
 }
+void context_sync()
+{
+    const int BUFSIZE=1000;
+    char buf[BUFSIZE];
+    char* name= getenv("CRO_NAME");
+    char* passphrase=getenv("CRO_PASSPHRASE");
+    char* enckey=getenv("CRO_ENCKEY");
+    const char* req_template = "{\"jsonrpc\": \"2.0\", \"method\": \"sync\", \"params\": [{\"name\":\"%s\", \"passphrase\":\"%s\",\"enckey\":\"%s\"}], \"id\": 1}";
+    char req[BUFSIZE];
+    sprintf(req, req_template, name, passphrase, enckey);
+
+    const char* user="i'm user";
+    const char* wallet_req = "{\"jsonrpc\": \"2.0\", \"method\": \"wallet_list\", \"params\": [], \"id\": 1}";
+
+    printf("sync with context\n");
+    CroJsonRpcPtr rpc= NULL;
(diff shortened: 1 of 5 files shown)


client-rpc/src/rpc/sync_rpc.rs
```diff
@@ -5,9 +5,23 @@ use crate::server::to_rpc_error;
...
+
+pub trait CBindingCallback: Send + Sync {
+    fn progress(&self, current: u64, start: u64, end: u64) -> i32;
+    fn set_user(&mut self, user: u64);
+    fn get_user(&self) -> u64;
+}
+
+#[derive(Clone)]
+pub struct CBindingCore {
+    pub data: Arc<Mutex<dyn CBindingCallback>>,
```
leejw51crypto: Arc is necessary, because RpcSync has "Send+Sync" trait.

cro-clib/examples/rpc.c
```diff
@@ -0,0 +1,93 @@
...
+
+    printf("sync with context\n");
+    CroJsonRpcPtr rpc= NULL;
+    cro_create_jsonrpc(&rpc, ".storage", "ws://localhost:26657/websocket", 0xab, &progress);
+    cro_run_jsonrpc(rpc, wallet_req, buf, sizeof(buf), user);
+    printf("response: %s\n", buf);    
+
+    cro_run_jsonrpc(rpc, req, buf, sizeof(buf),user);
+    printf("response: %s\n", buf);
+
```
leejw51crypto: this is called one by one

yihuang: Is it able to run concurrently?

yihuang: Can you pass different user data in the test case, and show that callback indeed receives different userdatas.

leejw51crypto:
yes, i'll modify it

cro-clib/examples/rpc.c
leejw51crypto: it's for synchronous call,
different user data will return

for multi-threaded,
context should be different

calvinaco:
Sorry for a dumb question. Is it the context design made sled DB connection to be reused during the lifetime of the context and thus reduce continuous open-close of DB?

leejw51crypto:
yes, sled storage is reused
also other http io handlers 
rebased

PR #1216, opened 2020-03-11
Problem:(fix #961) client-* assumes all key pairs / wallets to be in a local storage

### LINKED ISSUE #961: Problem: client-* assumes all key pairs / wallets to be in a local storage
For HW wallet support, this won't be the case -- client will be exposed public keys (upon authorisation on the device) / addresses, but signing will be external

linfeng-crypto:
* add wallet_kind in `Wallet`
* add `PrivatekeyTrait`, any object `impl`ed this trait can be a private key.
* add `PrivatekeyKind` enum to collect all kinds of private key
* hardware wallet is not implemented.


client-common/src/key/private_key.rs
```diff
@@ -6,10 +6,81 @@ use zeroize::Zeroize;
...
 
+/// types of private key
+#[derive(Clone)]
+pub enum PrivateKeyKind {
+    /// normal private key
+    Normal(PrivateKey),
+    // TODO: Add Hardware wallet as Privatekey
+}
+
+impl PrivateKeyTrait for PrivateKeyKind {
```
yihuang: If we just use the extensible `PrivateKeyKind` enum in logic code, no need for the trait `PrivateKeyTrait`, just normal method is enough?

client-core/src/cipher/default.rs
```diff
@@ -143,14 +141,11 @@ impl TransactionObfuscation for DefaultTransactionObfuscation {
...
-                DecryptionRequest::create(
-                    &secp,
-                    transaction_ids.to_owned(),
-                    ch,
-                    &private_key.into(),
-                )
-            });
+            let pk = match private_key_kind {
+                PrivateKeyKind::Normal(p) => p.into(),
+            };
```
yihuang: What happens for the hardware kind here, which I think not able to return private key?

client-core/src/cipher/mock_abci_transaction_obfuscation.rs
```diff
@@ -64,18 +64,16 @@ where
...
         transaction_ids: &[TxId],
-        private_key: &PrivateKey,
+        private_key_kind: &PrivateKeyKind,
     ) -> Result<Vec<Transaction>> {
         if transaction_ids.is_empty() {
             return Ok(vec![]);
         }
+        let pk = match private_key_kind {
+            PrivateKeyKind::Normal(p) => p.into(),
+        };
```
yihuang: This pattern matching seems to repeat, encapsulate it with a method?

client-common/src/key/private_key.rs
```diff
@@ -6,10 +6,81 @@ use zeroize::Zeroize;
 
 use crate::{ErrorKind, PublicKey, Result, ResultExt, SECP};
 
+/// types of private key
+#[derive(Clone)]
+pub enum PrivateKeyKind {
+    /// normal private key
+    Normal(PrivateKey),
+    // TODO: Add Hardware wallet as Privatekey
+}
```
tomtau: maybe:
```
/// temporary place holder
pub type HWalletConfig = ();
/// location
pub enum PrivateKeyLocation {
/// private key in local filesystem
Local(PrivateKey),
External(HWalletConfig),
}

File: client-common/src/key/private_key.rs (hunk 1 above)
Comment (tomtau) [reply client-common/src/key/private_key.rs]: this would be useful for tests with some `TestOnlyKey{ connection_str: String}`

File: client-common/src/key/private_key.rs (hunk 4 above)
Comment (tomtau) [reply client-common/src/key/private_key.rs]: in match statements, it could do `todo!()` for the ops at the moment.

one can still test the key management -- e.g. if one can create a wallet with these external keys (and say display / return this information)

File: client-core/src/cipher/default.rs (hunk 2 above)
Comment (tomtau) [reply client-core/src/cipher/default.rs]: this is for the view key -- I guess if this is implementing the "mixed" mode (where this will always be local), then it doesn't make sense to change `private_key: &PrivateKey` to that wrapper type?

File: client-common/src/key/private_key.rs [hunk 5]
```diff
@@ -6,10 +6,81 @@ use zeroize::Zeroize;
...
+    fn sign<T: AsRef<[u8]>>(&self, bytes: T) -> Result<RecoverableSignature> {
+        match self {
+            Self::Normal(pk) => pk.sign(bytes),
+        }
+    }
+    fn schnorr_sign<T: AsRef<[u8]>>(&self, bytes: T) -> Result<SchnorrSignature> {
+        match self {
+            Self::Normal(pk) => pk.schnorr_sign(bytes),
+        }
+    }
```
tomtau: this won't work with HW wallets:

1. this interface assumes it's being passed the message digest ("TXID") -- just like in https://github.com/crypto-com/chain/blob/master/client-network/src/network_ops/default_network_ops_client.rs#L317 -- for HW wallet, it won't be passed TXID, it'll be passed the *transaction data*, it'll parse this data, show some confirmation screen to the user (e.g. "confirm to sign this transaction" + details about the transaction) -- if the user presses a button on the device, inside the device, it will compute the TXID (from the transaction data e.g. sent over USB) and sign that

client-common/src/key/private_key.rs
```diff
@@ -6,10 +6,81 @@ use zeroize::Zeroize;
...
+            Self::Normal(pk) => pk.public_key(),
+        }
+    }
+}
+
+/// a object acts like a private key should impl the trait
+pub trait PrivateKeyTrait: Sync + Send {
+    /// Signs a message with current private key
+    fn sign<T: AsRef<[u8]>>(&self, bytes: T) -> Result<RecoverableSignature>;
+
```
leejw51crypto: what's differnt with below?

client-common/src/key/private_key.rs
leejw51crypto: PrivateKeyTrait 
it's alreay trait, choose something else,
refer to other coding paradigm

client-common/src/key/private_key.rs
leejw51crypto: yes, just add dummy enum
HW_LEDGER(ledger),
HW_TREASURE(treasure)

client-common/src/key/private_key.rs
```diff
@@ -1,15 +1,73 @@
...
+/// a object acts like a private key should impl the trait
+pub trait PrivateKeyAction: Sync + Send {
+    /// Signs a message with current private key
+    fn sign(&self, tx: &Transaction) -> Result<RecoverableSignature>;
+
+    /// Signs a message with current private key (uses schnorr signature algorithm)
+    fn schnorr_sign(&self, tx: &Transaction) -> Result<SchnorrSignature>;
+
+    /// Signs a message with current private key
```
tomtau: I guess this comments are out of date?

client-common/src/storage/hardware_storage.rs
```diff
@@ -0,0 +1,44 @@
...
+use secp256k1::recovery::RecoverableSignature;
+use secp256k1::schnorrsig::SchnorrSignature;
+
+/// TODO: the ledger hardware wallet
+#[derive(Clone)]
+pub struct HardwareLedger();
+
+/// TODO: the treasure hardware wallet
+#[derive(Clone)]
+pub struct HardwareTrezo();
```
tomtau: Trezor?

client-core/src/wallet/default_wallet_client.rs
```diff
@@ -293,10 +295,12 @@ where
                 self.key_service
                     .add_wallet_private_key(name, &private_key, &enckey)?;
 
-                self.wallet_service.create(name, &enckey, public_key)?;
+                self.wallet_service
+                    .create(name, &enckey, public_key, wallet_kind)?;
 
                 Ok((enckey, Some(mnemonic)))
             }
+            WalletKind::HW => todo!("find hardware wallet as private key, and create a new wallet"),
```
tomtau: you can still create the "HW wallet" -- the only thing will be that the operations requiring `sign` or `public_key` will fail?

client-core/src/transaction_builder/raw_transfer_transaction_builder.rs
```diff
@@ -1171,14 +1177,14 @@ mod raw_transfer_transaction_builder_tests {
...
     ) -> TxInWitness {
         let signing_addr = create_transfer_addr(public_key.clone());
 
         let signer =
             KeyPairSigner::new(private_key, public_key).expect("should create KeyPairSigner");
         signer
-            .schnorr_sign(tx_id, &signing_addr)
+            .schnorr_sign(tx, &signing_addr)
             .expect("should sign transaction id")
     }
```
tomtau: this function is used in transaction creation and it won't work with HW wallets, as it requires PrivateKey

client-core/src/cipher/default.rs
```diff
@@ -93,7 +91,7 @@ impl TransactionObfuscation for DefaultTransactionObfuscation {
     fn decrypt(
         &self,
         transaction_ids: &[TxId],
-        private_key: &PrivateKey,
+        private_key_location: &PrivateKeyLocation,
```
tomtau: Does `decrypt` need to take a `PrivateKeyLocation`? if you're implementing the "mixed" mode, then this could just be `PrivateKey`?

client-common/src/storage/hardware_storage.rs
```diff
@@ -0,0 +1,44 @@
...
+    fn sign(&self, _tx: &Transaction) -> Result<RecoverableSignature> {
+        todo!()
+    }
+    fn schnorr_sign(&self, _tx: &Transaction) -> Result<SchnorrSignature> {
+        todo!()
+    }
+    fn public_key(&self) -> PublicKey {
+        todo!()
+    }
+}
```
tomtau: this `todo!` could just be on `HardwareWallet`?

client-common/src/key/private_key.rs
```diff
@@ -1,15 +1,73 @@
...
+#[derive(Clone)]
+pub enum PrivateKeyLocation {
+    /// normal private key
+    Local(PrivateKey),
+    /// external hardware wallet as a private key
+    External(HardwareWallet),
+}
+
+/// a object acts like a private key should impl the trait
+pub trait PrivateKeyAction: Sync + Send {
```
tomtau: there's not much of a point for trait if all methods take directly `PrivateKeyLocation` instead of generic `T: PrivateKeyAction`. Either add a test and make it generic -- e.g. something like
```
pub enum PrivateKeyLocation<T: PrivateKeyAction> {
     /// normal private key
     Local(PrivateKey),
     /// external hardware wallet as a private key
      External(T),
  }
```

or just add plain methods directly on `PrivateKeyLocation`, so that the client code that takes `PrivateKeyLocation` can just do `.sign(...)` instead of always pattern matching on its variants

client-common/src/storage/hardware_storage.rs
leejw51crypto: yes, it's trezor

client-core/src/wallet/default_wallet_client.rs
linfeng-crypto: Yes I can, it will give an unreachable warning when compile if I use the `hw_wallet.public_key()`, because the `hw_wallet` comes from a `todo!()`.

client-network/src/network_ops/default_network_ops_client.rs
```diff
@@ -492,9 +493,7 @@ where
                 )
             })?;
 
-        let signature = private_key
-            .sign(transaction.id())
-            .map(StakedStateOpWitness::new)?;
+        let signature = private_key.sign(&tx).map(StakedStateOpWitness::new)?;
```
tomtau: `private_key` is obtained on L486 with `self.wallet_client.private_key(...)` which returns the `PrivateKey`, not `PrivateKeyLocation`

client-common/src/key/private_key.rs
```diff
@@ -1,15 +1,98 @@
...
+
+    /// Signs a message with current private key
+    fn public_key(&self) -> PublicKey;
+}
+
 /// Private key used in Crypto.com Chain
 #[derive(Debug, PartialEq, Clone)]
 pub struct PrivateKey(SecretKey);
 
+impl PrivateKeyAction for PrivateKey {
```
tomtau: implement it for PrivateKeyLocation ?

client-network/src/network_ops/default_network_ops_client.rs
tomtau: it may be worth adding a test with external "mock" HW wallet over TCP to check that the core transaction building functionality doesn't assume the a local private key

client-common/src/key/private_key.rs
tomtau: @linfeng-crypto

tomtau:
The "mock" testing stuff can possibly be feature-guarded, as it's normally not needed. Plus a few minor issues like https://github.com/crypto-com/chain/pull/1216#discussion_r391436986

other than that, if it works (the HW APIs do not assume local spending keypairs now), it can be possibly be merged (improvements to it can be submitted in a different PR) -- @yihuang @leejw51crypto @devashishdxt see if you have any other comments

leejw51crypto:
yes, i think hw api support can be another pr.
just interface here

tomtau:
bors try

PR #1217, opened 2020-03-11
Problem:(cro-1160) no ios example for c-binding#1160

leejw51crypto:
Solution: add ios example for wallet restore,sync


leejw51crypto:
close for tidy up

PR #1221, opened 2020-03-11
Problem: i64 doesn't represent the semantic of BlockHeight

yihuang:
Solution:
- Change `BlockHeight` to a newtype over u64
- Only support safe operations like `checked_sub/add`, `saturating_sub/add`.

### DIFF
--- a/chain-abci/src/app/query.rs
+++ b/chain-abci/src/app/query.rs
@@ -1,14 +1,17 @@
+use std::convert::{TryFrom, TryInto};
+
 use super::ChainNodeApp;
 use crate::enclave_bridge::{mock::handle_enc_dec, EnclaveProxy};
 use crate::storage::get_account;
 use abci::*;
 use chain_core::common::{MerkleTree, Proof as MerkleProof, H256, HASH_SIZE_256};
 use chain_core::state::account::{CouncilNodeMetadata, StakedStateAddress};
+use chain_core::state::tendermint::BlockHeight;
 use chain_core::state::ChainState;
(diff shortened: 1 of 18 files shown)


tomtau:
for the option 2, instead of `encode_var_vec`, it could just use scale codec encode/decode

chain-abci/src/app/query.rs
```diff
@@ -196,7 +199,9 @@ impl<T: EnclaveProxy> ChainNodeApp<T> {
                     resp.code = 1;
                     resp.log += "tx query address not set / state is not persisted";
                 } else {
-                    let value = self.storage.get_historical_state(_req.height);
+                    let value = self
+                        .storage
+                        .get_historical_state(_req.height.try_into().unwrap());
```
leejw51crypto: expect?

yihuang: Done

yihuang:
> since one needs be careful about under/overflows: either
> 
> 1. add warning comments
> 2. instead of `u64`, it could be an opaque struct `struct BlockHeight(u64)` that has enforces the checked operations

Changed BlockHeight to a newtype on u64.

yihuang:
> a small compile time error: https://travis-ci.org/github/crypto-com/chain/jobs/661453874#L811

This one seems strange, the line of code is incorrect, and I just checked `chain-abci/tests/abci_app.rs`, all the `key_from: ` is alright.

tomtau:
bors retry

Issue #817, opened 2020-01-06
Problem: enclave code directly contacts IAS


tomtau:
As there may be other enclave code, like https://github.com/crypto-com/chain-docs/blob/master/plan.md#transaction-data-bootstrapping-enclave-tdbe , it'll be desirable to have the IAS configuration in one place, so enclave code wouldn't directly call IAS, but use a proxy (that has IAS api key etc.)

devashishdxt:
@tomtau A few questions about IAS proxy:

- Should this be a thin proxy or a fat one?
  - Thin proxy: Receives HTTP(S) request, adds IAS api key in header, forwards that request to IAS and returns the response directly to caller.
  - Fat proxy: Runs a TCP server (TLS?), get binary encoded data as request, convert request data in format expected by IAS, parse the response and send binary encoded data back to client over TCP.
- Should this be over HTTPs (if thin)/TLS (if fat)?

tomtau:
> Should this be a thin proxy or a fat one?
it'll make sense to combine with "AESM proxy" (or whatever to call it) functionality for obtaining quotes. Whatever is simpler and makes sense.

> Should this be over HTTPs (if thin)/TLS (if fat)?
whatever is the simplest -- enclave will need to fully verify the responses anyway

PR #1246, opened 2020-03-16
Problem: Missing primitives for implementing remote attestation

devashishdxt:
Solution: Added crypto primitives in `enclave-utils`. Related to #817.


devashishdxt:
> tests?

I'll add them in future PRs. I'm breaking down one big PR in three PRs.
- Add primitives
- Implement `ra-sp`
- Add tests

chain-tx-enclave/enclave-utils/src/crypto/key_exchange.rs
```diff
@@ -0,0 +1,123 @@
+use std::io::Write;
```
tomtau: when is this custom key exchange used?

tomtau:
without tests or comments, it's hard to tell whether this solves the problem it claims to solve

chain-tx-enclave/enclave-utils/src/crypto/key_exchange.rs
devashishdxt: https://github.com/ndokmai/rust-sgx-remote-attestation/blob/7e2c26930d4a87aa040b3e1d5602c61fcd5145ee/ra-sp/src/context.rs#L142

devashishdxt:
> without tests or comments, it's hard to tell whether this solves the problem it claims to solve

Ok. I'll add tests in this PR.

chain-tx-enclave/enclave-utils/src/crypto/key_exchange.rs
tomtau: it's copying that example with a modified Sigma protocol: https://github.com/intel/sgx-ra-sample I guess?
Which component will need a communication over Sigma? Here, the end goal is to have an attested TLS client<->server channel, so one needs HTTPS to IAS, processing its payloads, AESM comms + verification of quotes (checking CMAC, SHA256(nonce||quote)... )

tomtau:
closing for now, as @devashishdxt will submit a new PR later

Issue #1222, opened 2020-03-11
Problem: non-enclave / public transactions don't have a common outer type


tomtau:
currently, they are individual variants of `TxAux`

PR #1249, opened 2020-03-16
Problem (Fix #1142): revised punishment mechanisms not implemented

### LINKED ISSUE #1142: Problem: revised punishment mechanisms not implemented
Ref:
https://github.com/crypto-com/chain-docs/pull/89
https://github.com/crypto-com/chain-docs/issues/84
#1096 
#1079
#856  
#987 
#988

yihuang:
Solution:
- Refactored validator state and public tx processing.

The punishment spec:
https://github.com/crypto-com/chain-docs/blob/master/docs/getting-started/punishments.md

The doc for staking state transition:
https://github.com/crypto-com/chain-docs/blob/bc4f5db86ab1bbb8cda06d67c05d27cbde8cde66/docs/modules/staking-state.md

The doc for tx validation logic:
https://github.com/crypto-com/chain-docs/blob/9779c11ccf647d1ec27538f264df79998312c1ca/docs/modules/transactions.md

The main logic is in the several newly added files:
- `chain-abci/src/staking_table.rs`, staking state transition, also the processing logic of public tx.
- `chain-abci/src/tx.rs`, entry point for tx processing.
- `chain-storage/src/buffer.rs`, buffered storage implementation.

TODO:
- Buffered storage not implemented for kv storage yet, after that, `check_tx` can call `process_tx` directly, currently `check_tx` is empty for now.
- `MAX_USED_VALIDATOR_ADDR` to be made a network parameter, currently a constant, modify network parameter involves update app_hash of unit tests, so I postpone that to later issue.
- `MAX_EVIDENCE_AGE` should be replaced by consensus parameter after upgraded to tendermint 0.33, currently a constant.
- The staking documentation about `MIN_EFFECTIVE_STAKE` need to be changed.

---
This change is [Reviewable](https://reviewable.io/reviews/crypto-com/chain/1249)


tomtau:
This is too complex to review, as it encompasses some other orthogonal changes and no-one may have time to review this massive PR in one go -- I used "reviewable" to go through it, as Github UI makes it infeasible on a small screen, and left a few comments. But overall, it'll be better in the interest of everyone to split this into several incremental PRs

yihuang:
> > Reviewed 53 of 53 files at r1.
> > _[Reviewable](https://reviewable.io/reviews/crypto-com/chain/1249)_ status: all files reviewed, 25 unresolved discussions (waiting on @calvinaco, @devashishdxt, @leejw51crypto, @linfeng-crypto, and @yihuang)
> 
> _[chain-abci/src/liveness.rs, line 37 at r1](https://reviewable.io/reviews/crypto-com/chain/1249#-M2agyYcDrnpPuW7Bg0m:-M2agyYcDrnpPuW7Bg0n:bk0obp8) ([raw file](https://github.com/crypto-com/chain/blob/2bfa4c672672ce8589681696e5647211290bd334/chain-abci/src/liveness.rs#L37)):_
> 
> > ```rust
> >     /// Count false
> >     pub fn count_false(&self) -> usize {
> >         self.liveness.iter().filter(|b| !b).count()
> > ```
> 
> should this be a "public" function?
> this seems more of an internal thing in tracking -- the outside logic just needs to know who to punish -- how that's determined is an implementation detail. Given the new docs keep validators out of "max" around, iterating through all of them on every block isn't going to cut it -- so later, I assume this will get optimised (keeping cached result for inactive validators; doing the count with a few POPCOUNT instructions instead of iterating over bits etc.)

It's not self-complete to decide "who to punish", because the network parameter is belong to chain state, and it is possible to change, what it can do is to tell how many non-lives blocks have recorded.
It might be better to rename to `count_inactives` or something like that.

> 
> _[chain-abci/src/liveness.rs, line 42 at r1](https://reviewable.io/reviews/crypto-com/chain/1249#-M2agT3o1bs87GWMADh2:-M2agT3o1bs87GWMADh3:b-bkr5s8) ([raw file](https://github.com/crypto-com/chain/blob/2bfa4c672672ce8589681696e5647211290bd334/chain-abci/src/liveness.rs#L42)):_
> 
> > ```rust
> >     /// grow BitVec to size.
> >     /// FIXME don't support thrink yet
> >     pub fn resize(&mut self, size: usize) {
> > ```
> 
> why is this needed? I don't see it in https://github.com/crypto-com/chain-docs/blob/master/docs/getting-started/punishments.md#livenesstracker

The size should always be consistent with the network parameter `block_signing_window`, so the size is actually dynamic because the network parameter can be changed.

> 
> _[chain-abci/src/staking_table.rs, line 1 at r1](https://reviewable.io/reviews/crypto-com/chain/1249#-M2ak4JT17lNxJloh72o:-M2ak4JT17lNxJloh72p:b-e691b5) ([raw file](https://github.com/crypto-com/chain/blob/2bfa4c672672ce8589681696e5647211290bd334/chain-abci/src/staking_table.rs#L1)):_
> 
> > ```rust
> > //! `ValidatorTable` combined with merkle trie is complete state of staking and validator,
> > ```
> 
> This file has over 1000 lines 🥺

It's ok to separate it into several files, I thought it might be easier to check correctness against specifications when putting together.
This file is the mostly self-contained(with the passed in staking store), it's the core logic of all staking/validator state related operations.

> 
> _[chain-abci/src/staking_table.rs, line 32 at r1](https://reviewable.io/reviews/crypto-com/chain/1249#-M2aiMKyBGujFDQuGsc3:-M2aiMKyBGujFDQuGsc4:b-9k807w) ([raw file](https://github.com/crypto-com/chain/blob/2bfa4c672672ce8589681696e5647211290bd334/chain-abci/src/staking_table.rs#L32)):_
> 
> > ```rust
> > const MAX_EVIDENCE_AGE: u64 = 10000;
> > // FIXME add a network parameter
> > const MAX_USED_VALIDATOR_ADDR: usize = 100;
> > ```
> 
> 100 may be too much -- generally, there'll only be 1 or 2 in normal circumstances. So 10 would be generous.
> this can be moved to chain-core and enforce this in decoding

I think it should be put into the network parameter?

> 
> _[chain-abci/src/staking_table.rs, line 35 at r1](https://reviewable.io/reviews/crypto-com/chain/1249#-M2ajiil6ptH6Bi-yAx8:-M2ajiil6ptH6Bi-yAx9:b-u0pn0t) ([raw file](https://github.com/crypto-com/chain/blob/2bfa4c672672ce8589681696e5647211290bd334/chain-abci/src/staking_table.rs#L35)):_
> 
> > ```rust
> > #[derive(Error, Debug)]
> > pub enum UnjailError {
> > ```
> 
> seems all these errors are orthogonal -- perhaps this is one thing that could be put in a separate PR (preceding this one)?

Maybe, but partially refactoring the existing code structure and make it correct is much extra work.

> 
> _[chain-abci/src/staking_table.rs, line 94 at r1](https://reviewable.io/reviews/crypto-com/chain/1249#-M2akq4tEeJr-hL_84vf:-M2akq4tEeJr-hL_84vg:bfi424d) ([raw file](https://github.com/crypto-com/chain/blob/2bfa4c672672ce8589681696e5647211290bd334/chain-abci/src/staking_table.rs#L94)):_
> 
> > ```rust
> > /// Invarient 2.2:
> > ///   All the secondary indexes should be partial index with condition `validator is not null`
> > ///   So there is no nothing-at-stake risk here, because there is a cost to become validator.
> > ```
> 
> how is that relevant? I'm not quite sure what this comment is trying to say

It basically states that all the staking items contained in the indexing structures should satisfy:
```
validator.is_some()
```
It means:
- Validator field in `StakedState` and items in the indexing structure are created/removed at the same time.
- Since there is a cost(min_required_staking) to join, the size of the indexing structure won't be too many.

> 
> _[chain-abci/src/staking_table.rs, line 108 at r1](https://reviewable.io/reviews/crypto-com/chain/1249#-M2alGBEDf5SD1qSDfLn:-M2alGBEDf5SD1qSDfLo:b-eafe1b) ([raw file](https://github.com/crypto-com/chain/blob/2bfa4c672672ce8589681696e5647211290bd334/chain-abci/src/staking_table.rs#L108)):_
> 
> > ```rust
> > pub struct StakingTable {
> >     // Selected validator voting powers of last executed end block
> >     pub validator_snapshot: BTreeMap<StakedStateAddress, TendermintVotePower>,
> > ```
> 
> does it need to be pub?

It's used by unit tests to check things, currently only used by `chain-abci`, so maybe `pub(crate)` is enough.

> 
> _[chain-abci/src/staking_table.rs, line 112 at r1](https://reviewable.io/reviews/crypto-com/chain/1249#-M2alfakAFumrnKxh8c4:-M2alfakAFumrnKxh8c5:b-ewnwiv) ([raw file](https://github.com/crypto-com/chain/blob/2bfa4c672672ce8589681696e5647211290bd334/chain-abci/src/staking_table.rs#L112)):_
> 
> > ```rust
> >     proposer_stats: BTreeMap<StakedStateAddress, u64>,
> > 
> >     // Call `initialize` to populate the indexes after deserialized.
> > ```
> 
> this is a bit error prone -- can this be called in a custom `impl Decode`?

No, because the initialization is executed against external staking storage, in this way the content of the indexing structure is automatically consistent with staking storage when loading up.
Of course, we can just serialize all the fields, but we might risk inconsistencies between chain state and staking state.
This function is only needed to be called in one place: `new_with_storage`, so I think the risk of misuse is minimal.

> 
> _[chain-abci/src/staking_table.rs, line 164 at r1](https://reviewable.io/reviews/crypto-com/chain/1249#-M2ambl9-hZOhRnKvZQX:-M2ambl9-hZOhRnKvZQY:b-5dnsef) ([raw file](https://github.com/crypto-com/chain/blob/2bfa4c672672ce8589681696e5647211290bd334/chain-abci/src/staking_table.rs#L164)):_
> 
> > ```rust
> >     /// Handle abci begin_block event
> >     /// no error other than internal invariants broken
> >     /// - cleanup validator records
> > ```
> 
> break up by sub-routines?
> 
> _[chain-abci/src/staking_table.rs, line 293 at r1](https://reviewable.io/reviews/crypto-com/chain/1249#-M2anC_G2a2BlUVgZsqv:-M2anC_G2a2BlUVgZsqw:bm4q9sw) ([raw file](https://github.com/crypto-com/chain/blob/2bfa4c672672ce8589681696e5647211290bd334/chain-abci/src/staking_table.rs#L293)):_
> 
> > ```rust
> >     /// Handle `NodeJoinTx`
> >     pub fn node_join(
> > ```
> 
> a bit orthogonal change? could be separated out in a PR;

The logic of `node_logic` and other public transactions is very closely related to staking/validator state logic, that's why it's impossible to only change staking/validator state without change public tx processing.
And I think the implementation also should be put together, it's easier to verify all the invariants, or the implications of state mutations.

> 
> just like with #1215 it'll be good to separate "mutation-less" validation from the update (if the transaction is valid)
> 
> _[chain-abci/src/staking_table.rs, line 361 at r1](https://reviewable.io/reviews/crypto-com/chain/1249#-M2anNezAwvGEdP57pdy:-M2anNezAwvGEdP57pdz:buzqflj) ([raw file](https://github.com/crypto-com/chain/blob/2bfa4c672672ce8589681696e5647211290bd334/chain-abci/src/staking_table.rs#L361)):_
> 
> > ```rust
> >     /// Handle `UnjailTx`
> >     pub fn unjail(
> > ```
> 
> this again seems like an orthogonal change

The same logic as `node-join`.

> 
> _[chain-abci/src/staking_table.rs, line 417 at r1](https://reviewable.io/reviews/crypto-com/chain/1249#-M2aniW9DOwn_4Q7Ucxu:-M2aniW9DOwn_4Q7Ucxv:b-1gdw11) ([raw file](https://github.com/crypto-com/chain/blob/2bfa4c672672ce8589681696e5647211290bd334/chain-abci/src/staking_table.rs#L417)):_
> 
> > ```rust
> >     /// The heap should not use the uncommited buffer.
> >     pub fn reward_total_staking(&self, heap: &StakingGetter) -> Coin {
> > ```
> 
> not sure how often this is to be called, but could be just a single value that gets updated with other updates (assuming they are well encapsulated)

This is only called once per reward period, bringing in an intermediate state would bringing extra complexity to maintain consistency between the mutable states.

> 
> _[chain-abci/src/staking_table.rs, line 460 at r1](https://reviewable.io/reviews/crypto-com/chain/1249#-M2ao2mu0W0HllbyZPFO:-M2ao2mu0W0HllbyZPFP:buzqflj) ([raw file](https://github.com/crypto-com/chain/blob/2bfa4c672672ce8589681696e5647211290bd334/chain-abci/src/staking_table.rs#L460)):_
> 
> > ```rust
> >     /// Handle deposit tx
> >     /// Enclave validation is done in enclave, only incomplete check here.
> >     pub fn deposit(
> > ```
> 
> this again seems like an orthogonal change

Same logic as `node_join`, to keep the mutation logic together for easier verification of the correctness.

> 
> _[chain-abci/src/staking_table.rs, line 481 at r1](https://reviewable.io/reviews/crypto-com/chain/1249#-M2ao6Df3tUbgiDhKfxN:-M2ao6Df3tUbgiDhKfxO:buzqflj) ([raw file](https://github.com/crypto-com/chain/blob/2bfa4c672672ce8589681696e5647211290bd334/chain-abci/src/staking_table.rs#L481)):_
> 
> > ```rust
> >     /// Handle unbond tx
> >     pub fn unbond(
> > ```
> 
> this again seems like an orthogonal change

Same logic as `node_join`

> 
> _[chain-abci/src/staking_table.rs, line 513 at r1](https://reviewable.io/reviews/crypto-com/chain/1249#-M2ao7381IErXtIGknHX:-M2ao7381IErXtIGknHY:buzqflj) ([raw file](https://github.com/crypto-com/chain/blob/2bfa4c672672ce8589681696e5647211290bd334/chain-abci/src/staking_table.rs#L513)):_
> 
> > ```rust
> >     /// Handle withdraw tx
> >     /// Enclave validation is done in enclave, only incomplete check here.
> >     pub fn withdraw(
> > ```
> 
> this again seems like an orthogonal change

Same logic as above.

> 
> _[chain-abci/src/staking_table.rs, line 676 at r1](https://reviewable.io/reviews/crypto-com/chain/1249#-M2aoabc-33SXWXr8fIx:-M2aoabc-33SXWXr8fIy:b9k7r5z) ([raw file](https://github.com/crypto-com/chain/blob/2bfa4c672672ce8589681696e5647211290bd334/chain-abci/src/staking_table.rs#L676)):_
> 
> > ```rust
> >     /// - Remove the validator record from heap
> >     /// - Remove from index structure
> >     /// Complexity: O(N), Do we need to make it O(log(N))?
> > ```
> 
> N is all validator records (even those out of max?)? how often is this executed?

It's executed once per block, I think O(N) might be good enough even for hundreds of items. We can do some profiling sometimes.
We can also add a time-ordered `BTreeMap` to index, so we can iterate by the range of time, that would bring down the computation complexity at the cost of code complexity.

> 
> _[chain-abci/src/staking_table.rs, line 856 at r1](https://reviewable.io/reviews/crypto-com/chain/1249#-M2apvLP1iEENFhR_FQp:-M2apvLQCgvEYJfLlLH3:b1aictw) ([raw file](https://github.com/crypto-com/chain/blob/2bfa4c672672ce8589681696e5647211290bd334/chain-abci/src/staking_table.rs#L856)):_
> 
> > ```rust
> > /// generate validator updates
> > fn diff_validators(
> >     old: &BTreeMap<StakedStateAddress, TendermintVotePower>,
> > ```
> 
> both are bounded by max validators, right?

Yes.

> 
> _[chain-abci/src/tx.rs, line 17 at r1](https://reviewable.io/reviews/crypto-com/chain/1249#-M2aqBHmCrHKIshMC4qB:-M2aqBHn9JpXCTzac9lU:b-bg60ti) ([raw file](https://github.com/crypto-com/chain/blob/2bfa4c672672ce8589681696e5647211290bd334/chain-abci/src/tx.rs#L17)):_
> 
> > ```rust
> > use crate::storage::{execute_enclave_tx, process_public_tx, verify_enclave_tx};
> > 
> > #[derive(thiserror::Error, Debug)]
> > ```
> 
> seems again orthogonal?

It's hard to only change staking logic without change public tx processing code, so I just separated public tx processing out of the enclave tx processing.

> 
> _[chain-abci/src/app/mod.rs, line 72 at r1](https://reviewable.io/reviews/crypto-com/chain/1249#-M2aqc3a5sKUDwb-nQ6a:-M2aqc3a5sKUDwb-nQ6b:b-m79hlw) ([raw file](https://github.com/crypto-com/chain/blob/2bfa4c672672ce8589681696e5647211290bd334/chain-abci/src/app/mod.rs#L72)):_
> 
> > ```rust
> >         info!("received checktx request");
> >         // FIXME implement check_tx as deliver_tx with different overlay buffer.
> >         ResponseCheckTx::new()
> > ```
> 
> checkstx accepts everything? probably also a "warning" log

I plan to fix this ASAP after current PR gets merged, by letting `check_tx` execute the same code as `deliver_tx`.

> 
> _[chain-abci/src/storage/mod.rs, line 281 at r1](https://reviewable.io/reviews/crypto-com/chain/1249#-M2arD817PUFGI632BNo:-M2arD817PUFGI632BNp:b70lpz5) ([raw file](https://github.com/crypto-com/chain/blob/2bfa4c672672ce8589681696e5647211290bd334/chain-abci/src/storage/mod.rs#L281)):_
> 
> > ```rust
> > }
> > 
> > fn check_attributes(
> > ```
> 
> orthogonal?

The result of refactoring public tx processing.

> 
> _[chain-core/src/state/account.rs, line 269 at r1](https://reviewable.io/reviews/crypto-com/chain/1249#-M2arzdt6IxkuO2eiMpB:-M2arzdt6IxkuO2eiMpC:b44wjxl) ([raw file](https://github.com/crypto-com/chain/blob/2bfa4c672672ce8589681696e5647211290bd334/chain-core/src/state/account.rs#L269)):_
> 
> > ```rust
> > #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Encode, Decode)]
> > #[cfg_attr(not(feature = "mesalock_sgx"), derive(Serialize, Deserialize))]
> > pub struct Validator {
> > ```
> 
> As some of these new data types are a superset or a slight modification of the existing ones, perhaps they could be in a different PR (just that logic operating on them would then be in a follow up PR)?

This is closely related to the new staking state logic, which is explained in the staking state transition specification.

> 
> _[chain-core/src/state/account.rs, line 476 at r1](https://reviewable.io/reviews/crypto-com/chain/1249#-M2asOJk6W5W8RFbtm6f:-M2asOJl5yhC7oEyIveB:b-t1b491) ([raw file](https://github.com/crypto-com/chain/blob/2bfa4c672672ce8589681696e5647211290bd334/chain-core/src/state/account.rs#L476)):_
> 
> > ```rust
> > #[derive(Clone, Debug, PartialEq, Eq, Encode, Decode)]
> > #[cfg_attr(not(feature = "mesalock_sgx"), derive(Serialize, Deserialize))]
> > pub struct ValidatorSortKey {
> > ```
> 
> does this need to be in chain-core?

I don't know, what're the general rules of putting things into chain-core?
I put it here because I think it's a property of `StakedState`, although it's only used by `staking_stable.rs` to choose validators.

> 
> _[chain-core/src/tx/mod.rs, line 249 at r1](https://reviewable.io/reviews/crypto-com/chain/1249#-M2asYLC-GoseovK9vER:-M2asYLC-GoseovK9vES:b-6lkl2v) ([raw file](https://github.com/crypto-com/chain/blob/2bfa4c672672ce8589681696e5647211290bd334/chain-core/src/tx/mod.rs#L249)):_
> 
> > ```rust
> >     /// Get chain hex id, only works on public tx.
> >     pub fn chain_hex_id(&self) -> u8 {
> > ```
> 
> maybe this PR would be better if this is resolved first: #1222 ?

The #1222 is actually implemented after this one, but of course, we can switch them, #1222 is relatively a small change.

> 
> _[chain-storage/src/buffer.rs, line 1 at r1](https://reviewable.io/reviews/crypto-com/chain/1249#-M2assct462qO22kkHhC:-M2assct462qO22kkHhD:b-b3fu4x) ([raw file](https://github.com/crypto-com/chain/blob/2bfa4c672672ce8589681696e5647211290bd334/chain-storage/src/buffer.rs#L1)):_
> 
> > ```rust
> > use std::collections::{hash_map::RandomState, HashMap};
> > ```
> 
> this one seems all right, but could be added earlier in a separate PR with more comments or tests?

I think it's ok to add this in a separate PR without modifying existing staking logic code, but if we want to change existing staking logic code to use this idea, that would be too much extra effort.
I'll add more comments.

> 
> _[integration-tests/run_multinode.sh, line 81 at r1](https://reviewable.io/reviews/crypto-com/chain/1249#-M2atDUw8sJavPnVBhsD:-M2atDUw8sJavPnVBhsE:blkb5le) ([raw file](https://github.com/crypto-com/chain/blob/2bfa4c672672ce8589681696e5647211290bd334/integration-tests/run_multinode.sh#L81)):_
> 
> > ```shell
> > fi
> > 
> > runtest "join" # non-live fault slash, re-join, unbond, re-join
> > ```
> 
> so does what jail did?

The non-live fault doesn't jail anymore, so I moved the non-live fault punishment test into the join test, save some testing environment preparation time.
Maybe we can should more cases into the byzantine tests: unjail/re-join

yihuang:
> just like with #1215 it'll be good to separate "mutation-less" validation from the update (if the transaction is valid)

I think there's no good reason to separate validation and execution for public tx.
- extra effort to keep them correct and consistent with each other.
- check_tx can(also should I think) execute the same code as deliver_tx
- let check_tx execute tx result on a shared buffer, allows sequence tx(e.g. tx2 spend output of tx1)
- For performance consideration, loading a staked state from merkle trie probably far slower than some checks and mutations in ram, and I think we need to do that loading anyway for any meaningful check_tx implementations. Of course, we'll need profiling to confirms these performance estimations.

tomtau:
> The logic of node_logic and other public transactions is very closely related to staking/validator state logic, that's why it's impossible to only change staking/validator state without change public tx processing.

I know some of these are related, but the point about orthogonal changes was that this is doing several other changes -- for the sanity of everyone, it'll be better in several smaller PRs that change this functionality. Some PRs itself may have a sequence / order of dependencies.

> I think there's no good reason to separate validation and execution for public tx...

there are several good reasons for separating them out -- e.g. raw tx offline checking without a full node state. The points mentioned can still be achieved with having these two cleanly separated rather than intermingled

yihuang:
> > The logic of node_logic and other public transactions is very closely related to staking/validator state logic, that's why it's impossible to only change staking/validator state without change public tx processing.
> 
> I know some of these are related, but the point about orthogonal changes was that this is doing several other changes -- for the sanity of everyone, it'll be better in several smaller PRs that change this functionality. Some PRs itself may have a sequence / order of dependencies.
> 
> > I think there's no good reason to separate validation and execution for public tx...
> 
> there are several good reasons for separating them out -- e.g. raw tx offline checking without a full node state. The points mentioned can still be achieved with having these two cleanly separated rather than intermingled

> there are several good reasons for separating them out -- e.g. raw tx offline checking without a full node state. The points mentioned can still be achieved with having these two cleanly separated rather than intermingled

That's dependent on how much information at hand, and how complete the check it need, maybe for `unbond` and `unjail`, it only takes the `StakedState` and some network parameters, but for `NodeJoin`, you need all the validator states to the check. 
For the raw tx offline checking, it probably doesn't even have `StakedState` at hand, the situation is different for different tx types and different usage scenarios.
We can wait for those specific scenarios to appear, then to decide what and how the code should be shared. 
For ABCI, what it needs is full and complete checking, it's easier to check the completeness when it's put together. If we separate them, should execution code panic or check again when some invariants are broken?

For example:
```rust
if let Some(val) = staking.validator {
  Err(ValidationFailed)
} else {
  do something with val.
}
```

If we seperate validation from execution, it becomes:
```rust
// validation
if staking.validator.is_some() {
  return Err(ValidationFailed);
}

// execution
let val = staking.validator.unwrap() {
  do something with val.
}
```

You'll have a "unwrap" in execution code, and to prove it actually won't panic, you need to refer to the validation code.

tomtau:
based on that example you gave, it's a misunderstanding -- it's for separating out stateless validation (and having explicit inputs after lookup or against a trait that does the lookup; + docs what's assumed to be checked) from stateful validation and state updates.

yihuang:
> based on that example you gave, it's a misunderstanding -- I meant for separating out stateless validation (and having explicit inputs after lookup or against a trait that does the lookup; + docs what's assumed to be checked) from stateful validation and state updates.

I guess what you mean is the first three points of public transaction validation listed here(https://github.com/crypto-com/chain-docs/blob/master/docs/modules/transactions.md#public-transactions)? which is:
- Check tx attributes (chain_hex_id, app_version)
- Verify staking witness
- Check the witness match the staking address

Do these checks in a standalone function?

tomtau:
possibly, it'll be better to check these in some PR that does the tx processing changes

PR #1258, opened 2020-03-17
Problem (Fix #1222): non-enclave / public transactions don't have a common outer type

### LINKED ISSUE #1222: Problem: non-enclave / public transactions don't have a common outer type

yihuang:
Solution:
- Change the structure of TxAux


yihuang:
> looks all right -- would be good to add an entry to breaking changes in changelog

Done

PR #1269, opened 2020-03-19
Problem: check_tx logic is not complete

yihuang:
- Use the buffer storage abstraction to share tx process code between
  check_tx/deliver_tx

> still laying the foundation for the revised staking pr.

---
This change is [Reviewable](https://reviewable.io/reviews/crypto-com/chain/1269)


chain-abci/src/app/app_init.rs
```diff
@@ -138,20 +141,29 @@ pub struct ChainNodeApp<T: EnclaveProxy> {
...
     pub tx_query_address: Option<String>,
+
+    /// consensus buffer of staking merkle trie storage
+    pub staking_buffer: HashMap<StakedStateAddress, StakedState>,
+    /// mempool buffer of staking merkle trie storage
+    pub mempool_staking_buffer: HashMap<StakedStateAddress, StakedState>,
+    /// consensus buffer of key-value storage
+    pub kv_buffer: HashMap<(u32, Vec<u8>), Option<Vec<u8>>>,
+    /// mempool buffer of key-value storage
+    pub mempool_kv_buffer: HashMap<(u32, Vec<u8>), Option<Vec<u8>>>,
```
tomtau: perhaps create at least type aliases (in chain-storage) for these field types (e.g. `StakingBuffer` and `KVBuffer`).
that way, it'll be a bit clearer in function parameters that it's expecting the chain-storage stuff (instead of potentially some other map) + a bit easier if this is to be switched to something else in the future (e.g. `CHashMap`)

chain-abci/src/app/app_init.rs
```diff
@@ -466,20 +490,68 @@ impl<T: EnclaveProxy> ChainNodeApp<T> {
...
+            .map(|state| state.top_level.account_root);
+        staking_getter!(self, root, deliver)
+    }
+
+    pub fn kv_store(&mut self, deliver: bool) -> impl StoreKV + '_ {
+        kv_store!(self, deliver)
+    }
+
+    pub fn flush_buffers(&mut self) {
+        let state = self.last_state.as_mut().unwrap();
```
tomtau: I'm not sure if panics/errors improved in "unwraps" in the latest rust -- it may still be better to use `expect` with some human-readable message that would help diagnosing what went wrong ("getting latest app state", "flushing staking storage", "flushing kv storage"...)

chain-abci/src/app/app_init.rs
```diff
@@ -466,20 +490,68 @@ impl<T: EnclaveProxy> ChainNodeApp<T> {
...
+        state.top_level.account_root = flush_staking_storage(
+            &mut self.accounts,
+            Some(state.top_level.account_root),
+            std::mem::take(&mut self.staking_buffer),
```
tomtau: very minor thing (no need to resolve / ok to merge without it) -- generally, functions are imported directly or their immediately surrounding module (`mem::take(...)`)

chain-abci/src/app/app_init.rs
tomtau: (also for consistency, as `tx_extra_info` uses `expect`)

chain-abci/src/app/macros.rs
```diff
@@ -0,0 +1,48 @@
+// Fight borrow checker
```
tomtau: not that useful comment -- I assume these are for buffer constructor boilerplate? Maybe better to put them in chain-storage + more explanatory comment?

chain-abci/src/app/app_init.rs
```diff
@@ -466,20 +490,68 @@ impl<T: EnclaveProxy> ChainNodeApp<T> {
...
+
+    pub fn staking_getter(&self, deliver: bool) -> impl GetStaking + '_ {
+        let root = self
+            .last_state
+            .as_ref()
+            .map(|state| state.top_level.account_root);
+        staking_getter!(self, root, deliver)
+    }
+
+    pub fn kv_store(&mut self, deliver: bool) -> impl StoreKV + '_ {
```
tomtau: it's unclear what this `deliver: bool` param is meant to be set when. As these are public functions, it'll be good to document them -- either with top-level comments, or using a more descriptive `enum` type instead of bool

chain-abci/src/app/mod.rs
```diff
@@ -1,3 +1,6 @@
+#[macro_use]
```
tomtau: is this still needed in Rust 2018: https://doc.rust-lang.org/edition-guide/rust-2018/macros/macro-changes.html ? It's not needed on crates, not sure about modules

chain-storage/src/api.rs
```diff
@@ -0,0 +1,161 @@
...
+    TxId,
+};
+
+use super::buffer::{GetKV, StoreKV};
+use super::{
+    LookupItem, StoredChainState, CHAIN_ID_KEY, COL_APP_HASHS, COL_APP_STATES, COL_EXTRA,
+    COL_NODE_INFO, GENESIS_APP_HASH_KEY, LAST_STATE_KEY,
+};
+
+pub fn get_last_app_state(db: &impl GetKV) -> Option<Vec<u8>> {
```
tomtau: do all these need to be `pub` or are they expected only within the crate (`pub(super)` or `pub(crate)`)?

chain-storage/src/buffer.rs
```diff
@@ -318,6 +319,22 @@ pub fn flush_kvdb<S: KeyValueDB + ?Sized, H: BuildHasher>(
     storage.write(tx)
 }
 
+/// Flush buffer to storage
+pub fn flush_storage<H: BuildHasher>(
+    storage: &mut Storage,
+    buffer: HashMap<(u32, Vec<u8>), Option<Vec<u8>>, H>,
+) -> std::io::Result<()> {
```
tomtau: `use std::io::Result`

chain-storage/src/lib.rs
```diff
@@ -79,8 +80,15 @@ pub struct Storage {
...
-    /// uncommitted values (for lookups during current block execution processing)
-    temp_tx_db: BTreeMap<(TxId, LookupItem), Vec<u8>>,
+}
+
+impl Get for Storage {
+    type Key = (u32, Vec<u8>);
+    type Value = Vec<u8>;
+    fn get(&self, key: &Self::Key) -> Option<Self::Value> {
+        let (col, key) = key;
+        self.db.get(*col, &key).unwrap()
```
tomtau: the unwrap is for IO errors? again, `expect` or a comment would be helpful

chain-storage/src/lib.rs
```diff
@@ -294,89 +223,9 @@ impl Storage {
...
-            self.temp_tx_db.clear();
             self.db.write(dbtx)
         } else {
             Ok(())
         }
     }
 }
-
-#[cfg(test)]
-mod test {
```
tomtau: could there be similar tests in buffer? currently, the tests in buffer only use `kvdb_memorydb` which has different semantics "on drop" (or potentially sigterm)

chain-abci/src/app/app_init/validator_state.rs
```diff
@@ -643,7 +636,7 @@ impl ValidatorStateHelper {
     }
 
     /// creates a fresh helper from storage and last known state
-    pub fn restore(accounts: &AccountStorage, last_app_state: &ChainNodeState) -> Self {
```
tomtau: opened this https://github.com/crypto-com/chain/issues/1272 for later

chain-abci/src/app/macros.rs
yihuang: They are closely relate to implementation detail of `ChainNodeApp` (the field names), ideally it should be methods of `ChainNodeApp`, but borrow checker won't allow that, because it will retain the whole reference to `self`, although it only need several fields.

chain-abci/src/app/mod.rs
yihuang: https://doc.rust-lang.org/edition-guide/rust-2018/macros/macro-changes.html#more-details
Yes, it's still needed for locally defined macros.

tomtau:
bors try

chain-abci/src/app/app_init.rs
yihuang: I think for diagnoses it don't make much difference, but it's better for human reading the code. I'll add them.

chain-storage/src/api.rs
yihuang: These are the public api directly used by outside, because they can operate on different buffers.
Currently, the methods left in `Storage` are:
- read only, skipped the buffer, implies read committed semantic.
- genesis store methods, they are used before `ChainNodeApp` get initiated.

chain-storage/src/buffer.rs
yihuang: That would conflict with prelude `Result`, then you need to rename into `IoResult` or something, I think the fully qualified name is short enough to use directly?

chain-abci/src/app/app_init.rs
yihuang: Done, Changed to `BufferType`.

chain-storage/src/lib.rs
yihuang: Done

chain-storage/src/lib.rs
yihuang: > could there be similar tests in buffer? currently, the tests in buffer only use `kvdb_memorydb` which has different semantics "on drop" (or potentially sigterm?)

Do you mean different semantics when dropped without calling `persist_write`?

chain-abci/src/app/macros.rs
yihuang: Added some comments there.

chain-abci/src/app/app_init.rs
yihuang: Done

chain-abci/src/app/app_init.rs
yihuang: Actually this method is not used, so removed.

chain-abci/src/app/app_init.rs
yihuang: Done

yihuang:
The integration tests failure didn't reproduce, record is as #1273

chain-storage/src/lib.rs
tomtau: probably not relevant, as the top-level apis changed

chain-storage/src/api.rs
tomtau: it's more that similar undocumented public APIs are exposed both on "Storage" and here -- later when it's more cleaned up, it'll be good to consolidate it and document

PR #1278, opened 2020-03-21
Problem: infeasible to create a large chunk of transfer addresses in …

leejw51crypto:
Solution: make wallet to use storage directly
1. Root cause of the problem:
wallet is encoded, decoded, to one key, value.
a few address are ok, but addresses are about 1,000,000,

also if wallet is very big, it's also encoded, decoded, that time linearly increases with address increases.

it's O(n) , became very slow
now it's O(log(n))

also, for finding roothash, publickey,
it was O(n), now, it utilize keyvalue get ,
became, O(log(n))

2.What solutions did you consider to solve the root cause?
now, it's O(log(n)), and wallet is written to db directly.

3. What is the chosen solution?
publickey, stakingkey, roothases are flattened to key,value with different keyspace.

4.New benchmarks for generaton 1,0000,000 utxo addresses
999994
999995
999996
999997
999998
999999
1000000

real	82m2.542s
user	51m53.714s
sys	1m47.806s

key generation and write to db time is constant)
time = 334.744µs
time = 323.544µs
time = 321.788µs
costs are from http restful api processing cost,
so, i'll add batch generation,too.
if then, time can be done with 10 minutes for generating 1,000,000 utxo addresses

with batch) for generation 1,000,000 addresses by 50000 batch,
as original issue stated , batch made even faster for generating addresses

result 750000 50000
result 800000 50000
result 850000 50000
result 900000 50000
result 950000 50000
result 1000000 50000

real	3m43.747s
user	0m0.172s
sys	0m0.01

size benchmark) for 1,000,000 addresses
about 30 GB -> 1.9 GB for .storage

### DIFF
--- a/client-core/src/service/root_hash_service.rs
+++ b/client-core/src/service/root_hash_service.rs
@@ -1,10 +1,10 @@
 use parity_scale_codec::{Decode, Encode};
 
+use super::wallet_service::get_multisig_keyspace;
 use chain_core::common::{Proof, H256};
 use chain_core::tx::witness::tree::RawXOnlyPubkey;
 use client_common::MultiSigAddress;
 use client_common::{ErrorKind, PublicKey, Result, ResultExt, SecKey, SecureStorage, Storage};
-
 const KEYSPACE: &str = "core_root_hash";
 
 /// Maintains mapping `multi-sig-public-key -> multi-sig address`
@@ -36,21 +36,25 @@ where
             MultiSigAddress::new(public_keys, self_public_key, required_signers)?;
 
         let root_hash = multi_sig_address.root_hash();
-        let mut storage_key = root_hash.clone().to_vec();
-        storage_key.extend(name.as_bytes().iter());
 
-        self.storage
-            .set_secure(KEYSPACE, storage_key, multi_sig_address.encode(), enckey)?;
+        // key: roothash
+        // value: multisig address info
+        let multisigaddress_keyspace = get_multisig_keyspace(name);
+        self.storage.set_secure(
+            multisigaddress_keyspace,
+            hex::encode(&root_hash),
+            multi_sig_address.encode(),
+            enckey,
+        )?;
 
         Ok((root_hash, multi_sig_address))
     }
 
     /// delete root hash
-    pub fn delete_root_hash(&self, name: &str, root_hash: &H256, enckey: &SecKey) -> Result<()> {
-        let mut storage_key = root_hash.to_vec();
-        storage_key.extend(name.as_bytes().iter());
-        self.storage.get_secure(KEYSPACE, &storage_key, enckey)?;
-        self.storage.delete(KEYSPACE, &storage_key)?;
+    pub fn delete_root_hash(&self, name: &str, root_hash: &H256, _enckey: &SecKey) -> Result<()> {
+        let multisigaddress_keyspace = get_multisig_keyspace(name);
+        self.storage
+            .delete(multisigaddress_keyspace, hex::encode(&root_hash))?;
         Ok(())
     }
 
@@ -91,11 +95,10 @@ where
         root_hash: &H256,
         enckey: &SecKey,
     ) -> Result<MultiSigAddress> {
-        let mut storage_key = root_hash.to_vec();
-        storage_key.extend(name.as_bytes().iter());
+        let multisigaddress_keyspace = get_multisig_keyspace(name);
         let address_bytes = self
             .storage
-            .get_secure(KEYSPACE, storage_key, enckey)?
+            .get_secure(multisigaddress_keyspace, hex::encode(root_hash), enckey)?
             .chain(|| (ErrorKind::InvalidInput, "Address not found"))?;
 
         MultiSigAddress::decode(&mut address_bytes.as_slice()).chain(|| {
--- a/client-core/src/service/wallet_service.rs
+++ b/client-core/src/service/wallet_service.rs
@@ -2,6 +2,7 @@ use indexmap::IndexSet;
 use parity_scale_codec::{Decode, Encode, Input, Output};
 
 use crate::service::{load_wallet_state, WalletState};
+use crate::types::WalletKind;
 use chain_core::common::H256;
 use chain_core::init::address::RedeemAddress;
 use chain_core::state::account::StakedStateAddress;
@@ -9,17 +10,50 @@ use chain_core::tx::data::address::ExtendedAddr;
 use client_common::{
     Error, ErrorKind, PrivateKey, PublicKey, Result, ResultExt, SecKey, SecureStorage, Storage,
 };
-use std::fmt;
-//use serde::ser::{Serialize, SerializeStruct, Serializer};
-use crate::types::WalletKind;
-use parity_scale_codec::alloc::collections::BTreeMap;
 use secstr::SecUtf8;
 use serde::de::{self, Visitor};
 use serde::export::PhantomData;
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
-
+use std::fmt;
+use std::str;
 /// Key space of wallet
 const KEYSPACE: &str = "core_wallet";
+
+fn get_public_keyspace(name: &str) -> String {
+    format!("{}_{}_publickey", KEYSPACE, name)
+}
+
+fn get_stakingkey_keyspace(name: &str) -> String {
+    format!("{}_{}_stakingkey", KEYSPACE, name)
+}
+
+fn get_stakingkeyset_keyspace(name: &str) -> String {
+    format!("{}_{}_stakingkeyset", KEYSPACE, name)
+}
+
+fn get_private_keyspace(name: &str) -> String {
+    format!("{}_{}_privatekey", KEYSPACE, name)
+}
+
+fn get_roothash_keyspace(name: &str) -> String {
+    format!("{}_{}_roothash", KEYSPACE, name)
+}
+
+fn get_roothashset_keyspace(name: &str) -> String {
+    format!("{}_{}_roothashset", KEYSPACE, name)
+}
+
+pub fn get_multisig_keyspace(name: &str) -> String {
+    format!("{}_{}_multisigaddress", KEYSPACE, name)
+}
+
+fn get_info_keyspace(name: &str) -> String {
+    format!("{}_{}_info", KEYSPACE, name)
+}
+
+fn get_wallet_keyspace() -> String {
+    format!("{}_walletname", KEYSPACE)
+}
 
 fn serde_to_str<T, S>(value: &T, serializer: S) -> std::result::Result<S::Ok, S::Error>
 where
@@ -75,17 +109,14 @@ pub struct WalletInfo {
 }
 
 /// Wallet meta data
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub struct Wallet {
     /// view key to decrypt enclave transactions
     pub view_key: PublicKey,
-    /// public key and private key pair
-    pub key_pairs: BTreeMap<PublicKey, PrivateKey>,
-    /// public keys to construct transfer addresses
-    pub public_keys: IndexSet<PublicKey>,
     /// public keys of staking addresses
     pub staking_keys: IndexSet<PublicKey>,
     /// root hashes of multi-sig transfer addresses
+    // this is transfer address
     pub root_hashes: IndexSet<H256>,
     /// wallet type
     pub wallet_kind: WalletKind,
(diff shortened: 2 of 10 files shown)


client-core/src/wallet/default_wallet_client.rs
```diff
@@ -360,10 +360,6 @@ where
         }
         self.key_service.delete_wallet_private_key(name, &enckey)?;
 
-        for root_hash in wallet.root_hashes.iter() {
-            self.root_hash_service
-                .delete_root_hash(name, root_hash, &enckey)?;
-        }
```
leejw51crypto: this is removed when keyspace is removed

client-core/src/service/wallet_service.rs
```diff
@@ -248,28 +290,73 @@ where
...
 
     /// Get the wallet state from storage
+    // storage -> wallet
     pub fn get_wallet_state(&self, name: &str, enckey: &SecKey) -> Result<WalletState> {
         load_wallet_state(&self.storage, name, enckey)?.err_kind(ErrorKind::InvalidInput, || {
             format!("WalletState with name ({}) not found", name)
         })
     }
 
+    /// Save wallet to storage
```
leejw51crypto: this is used , when wallet is imported

leejw51crypto:
ok,i'll add it

devashishdxt:
> Solution: make wallet to use storage directly
> 
>     1. Root cause of the problem:
>        wallet is encoded, decoded, to one key, value.
>        a few address are ok, but addresses are about 1,000,000,
> 
> 
> also if wallet is very big, it's also encoded, decoded, that time linearly increases with address increases.
> 
> it's O(n) , became very slow
> now it's O(log(n))
> 
> also, for finding roothash, publickey,
> it was O(n), now, it utilize keyvalue get ,
> became, O(log(n))
> 
> 2.What solutions did you consider to solve the root cause?
> now, it's O(log(n)), and wallet is written to db directly.
> 
>     1. What is the chosen solution?
>        publickey, stakingkey, roothases are flattened to key,value with different keyspace.
> 
> 
> 4.New benchmarks for generaton 1,0000,000 utxo addresses
> 999994
> 999995
> 999996
> 999997
> 999998
> 999999
> 1000000
> 
> real 82m2.542s
> user 51m53.714s
> sys 1m47.806s
> 
> key generation and write to db time is constant)
> time = 334.744µs
> time = 323.544µs
> time = 321.788µs
> costs are from http restful api processing cost,
> so, i'll add batch generation,too.
> if then, time can be done with 10 minutes for generating 1,000,000 utxo addresses
> 
> with batch) for generation 1,000,000 addresses by 50000 batch,
> as original issue stated , batch made even faster for generating addresses
> 
> result 750000 50000
> result 800000 50000
> result 850000 50000
> result 900000 50000
> result 950000 50000
> result 1000000 50000
> 
> real 3m43.747s
> user 0m0.172s
> sys 0m0.01
> 
> size benchmark) for 1,000,000 addresses
> about 30 GB -> 1.9 GB for .storage

I thought the issue was with `Sled` database engine. Is using `Sled` file @leejw51crypto?

leejw51crypto:
yes, sled is not the cause of the problem.
i think it's not bad

client-core/src/service/wallet_service.rs
```diff
@@ -212,17 +208,63 @@ pub fn load_wallet<S: SecureStorage>(
...
+            if let Ok(value) =
+                storage.get_secure(private_keyspace.clone(), pubkey.serialize(), enckey)
+            {
+                if let Some(raw_value) = value {
+                    let privatekey = PrivateKey::deserialize_from(&raw_value)
+                        .expect("get privatekey in load_wallet");
+                    new_wallet.key_pairs.insert(pubkey, privatekey);
+                }
+            }
+        }
```
devashishdxt: Is there any reason to manualy fetch all these values from different keyspaces instead of encoding at a single place (like it was earlier)?

client-core/src/service/wallet_service.rs
```diff
@@ -248,28 +290,73 @@ where
...
         load_wallet_state(&self.storage, name, enckey)?.err_kind(ErrorKind::InvalidInput, || {
             format!("WalletState with name ({}) not found", name)
         })
     }
 
+    /// Save wallet to storage
+    pub fn save_wallet(&self, name: &str, enckey: &SecKey, wallet: &Wallet) -> Result<()> {
+        self.storage
+            .save_secure(KEYSPACE, name, enckey, wallet)
+            .expect("save walet-name in save_wallet");
```
devashishdxt: There is a lot of places where `expect` is used. Proper error handling will be better.

client-core/src/service/root_hash_service.rs
```diff
@@ -36,21 +36,25 @@ where
...
 
         let root_hash = multi_sig_address.root_hash();
-        let mut storage_key = root_hash.clone().to_vec();
-        storage_key.extend(name.as_bytes().iter());
 
-        self.storage
-            .set_secure(KEYSPACE, storage_key, multi_sig_address.encode(), enckey)?;
+        // key: roothash
+        // value: multisig address info
+        let multisigaddress_keyspace = get_multisig_keyspace(name);
```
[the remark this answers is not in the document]
leejw51crypto: good point,
i'll fix that.

client-core/src/service/wallet_service.rs
```diff
@@ -91,53 +119,16 @@ pub struct Wallet {
...
-            key.encode_to(dest);
-        }
-        (self.staking_keys.len() as u64).encode_to(dest);
-        for key in self.staking_keys.iter() {
-            key.encode_to(dest);
-        }
-        (self.root_hashes.len() as u64).encode_to(dest);
-        for hash in self.root_hashes.iter() {
-            hash.encode_to(dest);
-        }
```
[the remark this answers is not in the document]
leejw51crypto: that make wallet very big, also encoded and decoded in inserting one tx address.
it's written directly in wallet service.

client-core/src/service/wallet_service.rs
```diff
@@ -91,53 +119,16 @@ pub struct Wallet {
...
-        let mut staking_keys = IndexSet::with_capacity(len as usize);
-        for _ in 0..len {
-            staking_keys.insert(PublicKey::decode(input)?);
-        }
-
-        let len = u64::decode(input)?;
-        let mut root_hashes = IndexSet::with_capacity(len as usize);
-        for _ in 0..len {
-            root_hashes.insert(H256::decode(input)?);
-        }
```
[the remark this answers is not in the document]
leejw51crypto: these moved to wallet service.

client-core/src/service/wallet_service.rs
leejw51crypto: because that make inserting speed O(n), when addresses are 1,0000,000

client-core/src/service/wallet_service.rs
leejw51crypto: good point, i'll fix it

client-core/src/service/wallet_service.rs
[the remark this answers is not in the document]
devashishdxt: But how is changing the encoding/decoding is a fix. This seems more like a hack. Sholdn't we change the data model instead of chaning encoding/decoding?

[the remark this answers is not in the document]
leejw51crypto: no, this is not a hack.
it's common style of using embedded db

client-core/src/service/wallet_service.rs
devashishdxt: Maybe I'm missing something. But, how does this improve time complexity. This is doing exactly the same thing as earlier (the only different is that now we're storing values in different keyspaces).

leejw51crypto: no, it's very fast
see the benchmark

client-core/src/service/wallet_service.rs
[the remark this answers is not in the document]
devashishdxt: How is this not a hack? If I have some data in `Wallet` struct, after encoding and decoding back to `Wallet`, initial and final objects won't me equal. This is the basic test of checking encoding and decoding (and current code will fail that check).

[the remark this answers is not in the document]
leejw51crypto: i wan to remove wallet encoding, decoding,
but other code relies on it.
it is not necessary.

client-core/src/service/wallet_service.rs
devashishdxt: Ok. Maybe I'm not able to see the whole picture. I'll let others comment on this.

leejw51crypto: thanks,
previous code cannot produce 1,000,000 addresses within 5 minutes.
i search many ways to do it.

leejw51crypto:
i look at the previous code,
but multi-sig address are all in wallet-name space.
```
 let mut storage_key = root_hash.clone().to_vec();
 storage_key.extend(name.as_bytes().iter());
```

storage-key is per wallet.

client-core/src/service/root_hash_service.rs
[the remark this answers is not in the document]
leejw51crypto: it's the same with previous code, could you check it again?
```
    let root_hash = multi_sig_address.root_hash();
        let mut storage_key = root_hash.clone().to_vec();
        storage_key.extend(name.as_bytes().iter());

```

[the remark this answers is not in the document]
leejw51crypto: i checked
```
   let mut storage_key = root_hash.clone().to_vec();
        storage_key.extend(name.as_bytes().iter());

```
it's previous code.
it's the same, only extending name is moved to keyspace.

tomtau:
bors try

client-core/src/service/wallet_service.rs
tomtau: > (the only different is that now we're storing values in different keyspaces).

my guess is that this makes a difference, my understanding is:
- before there was just one key-pair, something like: "wallet-name" -> "wallet-data"
- each address creation, added bytes to wallet-data -- each time this whole thing had to be fetched, decrypted, deserialized etc.
- besides that, it keeps updating this single key -- generally, inserts are faster than updates + may create less storage bloat in internal sled accounting etc.

client-core/src/service/wallet_service.rs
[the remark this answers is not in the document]
tomtau: Which code relies on wallet encode/decode?
Are those fields like "key_pairs", "public_keys", "staking_keys" etc. still necessary, if this initialises as "empty"... and many of those:
```
.get_wallet(name, enckey)?		
              .<SOME_LOOKUP_METHOD>(...)
```
which did use these fields were replaced with directly looking it up in the storage?

leejw51crypto:
i'll check

leejw51crypto:
done

tomtau:
bors retry

tomtau:
the unit tests fail https://travis-ci.org/github/crypto-com/chain/jobs/665733283#L1146

leejw51crypto:
i'm checking

leejw51crypto:
fix the unitest and batch rpc returns u32

tomtau:
bors retry

leejw51crypto:
i'm checking drone.io , it's caused by different error messages.

tomtau:
bors retry

leejw51crypto:
there is error in multi-node test,
I'm checking it

leejw51crypto:
integration test cause is
address generation order is ignored.

i'll fix it

thanks @yihuang

tomtau:
bors retry

tomtau:
bors retry

leejw51crypto:
fix travis issue

tomtau:
bors retry

client-core/src/service/wallet_service.rs
```diff
@@ -91,53 +127,16 @@ pub struct Wallet {
...
+        let key_pairs = BTreeMap::new();
+        let public_keys = IndexSet::new();
+        let staking_keys = IndexSet::new();
+        let root_hashes = IndexSet::new();
```
tomtau: Are these fields still necessary if they are read from storage directly?

client-core/src/service/wallet_service.rs
```diff
@@ -177,52 +176,159 @@ impl Wallet {
...
+    storage: &S,
+    keyspace: &str,
+    key: &str,
+    defaut_value: Option<u64>,
+) -> Result<u64> {
+    let value = storage.get(keyspace, key.as_bytes())?;
+    if let Some(raw_value) = value {
+        let mut v: [u8; 8] = [0; 8];
+        v.copy_from_slice(&raw_value);
+        let index_value: u64 = u64::from_be_bytes(v);
```
tomtau: any special reason why big endian (note: almost everywhere else in the code, little endian is used)?

client-core/src/service/wallet_service.rs
```diff
@@ -248,28 +354,76 @@ where
...
-        save_wallet(&self.storage, name, enckey, &wallet)
+        self.save_wallet(name, enckey, &wallet)
     }
 
     /// Finds staking key corresponding to given redeem address
     pub fn find_staking_key(
         &self,
         name: &str,
-        enckey: &SecKey,
+        _enckey: &SecKey,
```
tomtau: perhaps TODO comment to change this API?

client-core/src/service/wallet_service.rs
```diff
@@ -307,31 +485,72 @@ where
...
+            .set(wallet_keyspace, name, name.as_bytes().to_vec())?;
+
+        Ok(())
     }
 
     /// Returns view key of wallet
     pub fn view_key(&self, name: &str, enckey: &SecKey) -> Result<PublicKey> {
-        let wallet = self.get_wallet(name, enckey)?;
-        Ok(wallet.view_key)
+        let _wallet_found = self.get_wallet(name, enckey)?;
```
tomtau: this is just to check if the wallet exists?

leejw51crypto: yes, because integration needs exact the same error message

client-core/src/service/wallet_service.rs
leejw51crypto: ok

client-core/src/service/wallet_service.rs
leejw51crypto: no special reason, little endian is also good

client-core/src/service/wallet_service.rs
leejw51crypto: other sync code use it, so to be compatible with other sources

client-core/src/service/wallet_service.rs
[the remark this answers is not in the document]
leejw51crypto: get public keys, get roothash, get staking keys is now directly referring to storage.
i mean all apis of wallet.
actually wallet is just temp holder

client-core/src/service/wallet_service.rs
tomtau: @leejw51crypto do you have some example where these fields are used directly? 
In sync, I only see it uses `wallet.staking_addresses()`, for example, and this function was changed to load from storage instead of that field?

tomtau: If they are never used, they can perhaps be removed from the `struct Wallet` ?

client-core/src/service/wallet_service.rs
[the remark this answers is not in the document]
tomtau: https://github.com/crypto-com/chain/pull/1278/files#r397585069

client-core/src/service/wallet_service.rs
leejw51crypto: agree, i'll remove it after check

tomtau: but that's also taken directly from the storage: https://github.com/crypto-com/chain/pull/1278/files#diff-fcaa70a81e2fb374ccdf66db46aeae86R565 ?

leejw51crypto: yes, that's possible, 
but need to change syncer_logic, sync.

leejw51crypto: also memory stoage should use BTreeMap, need to apply after this pr

tomtau: @leejw51crypto no, my question / observation was that syncer_logic / sync seem to only call methods on `Wallet` -- and these methods were changed in this PR -- so it seems those fields in `Wallet` struct are never used, even in syncer_logic / sync ?

tomtau:
bors try

leejw51crypto:
i'll check

client-core/src/service/wallet_service.rs
yihuang: Wallet syncer uses `wallet.view_key`/`wallet.staking_addresses()`/`wallet.transfer_addresses()`.

leejw51crypto:
rebased, fix conflics
writing document for new database approach

tomtau:
bors try

client-core/src/service/wallet_service.rs
tomtau: will merge this PR; some improvements (removing these redundant fields; safer concurrency etc. can be in other PRs)

leejw51crypto: ok

Issue #1223, opened 2020-03-11
Problem: TxoIndex is misused to refer to output size 


tomtau:
it'll be good to have an explicit name for it -- e.g. TxoSize

yihuang:
We want to have `TxoSize` used both for size and index right, just like `usize`?

tomtau:
yeah, maybe that's fine and still meaningful, as with `usize`

leejw51crypto:
started

leejw51crypto:
i'll change
TxoIndex -> TxoSize

like usize,
just TxoSize seems OK

leejw51crypto:
in code-review

PR #1285, opened 2020-03-23
Problem:(cro-1223) TxoIndex is misused to refer to output size #1223

leejw51crypto:
Solution: rename TxoIndex to TxoSize
this is just renaming type "TxoIndex", that is u16

other code is the same.


leejw51crypto:
ok , i'll check

tomtau:
bors try

PR #1292, opened 2020-03-24
Problem (Fix #1142): revised punishment mechanisms not implemented

### LINKED ISSUE #1142: Problem: revised punishment mechanisms not implemented

yihuang:
Solution:
- revised staking and punishment implementation

---
This change is [Reviewable](https://reviewable.io/reviews/crypto-com/chain/1292)


tomtau:
bors try

yihuang:
- Added `chain-abci/src/tx_error.rs`
- Splited `staking_table.rs` into `staking/mod.rs`, `staking/tx.rs`, `staking/table.rs`.

yihuang:
> 
> _[chain-abci/src/staking_table.rs, line 451 at r1](https://reviewable.io/reviews/crypto-com/chain/1292#-M3B4NSX5ZjS_nWrIbJa:-M3FR-D10RQpV5rmLTBO:b1ccpc7) ([raw file](https://github.com/crypto-com/chain/blob/4618ee0d0244b4587491577b573b64bfb1054495/chain-abci/src/staking_table.rs#L451)):_
> 
> Previously, yihuang (yihuang) wrote…
> > I don't think they should be together, because nonce increment should stick to operation, an operation might modify bonded multiple times, nonce should increase only once.
> 
> it'd increase multiple times, as the resulting amount may be the same (but other things may change), so would require double-checking the tx

For example, unbond tx will mutate staking at several places (unbounded, bonded, unbonded_from), the nonce should increase once or multiple times?

yihuang:
For the integration test failure, I added a temporary fix:
```diff
unbonded_from = rpc.staking.state(addr)['unbonded_from']
 print('Wait until unbonded_from', unbonded_from)
-wait_for_blocktime(rpc, unbonded_from)
+# FIXME add extra 5 seconds to prevent nonce contension, remove after #1293 fixed
+wait_for_blocktime(rpc, unbonded_from + 5)
```

yihuang:
Modified nonce logic, local test has passed.

tomtau:
bors try

yihuang:
The current failure of multitx test, also happen occasionally, I think it's not related to this PR, recorded as  #1316

yihuang:
I've tracked down the cause for the integration test failure, I think it's a bug in wallet sync, will be solved in #1316

PR #1298, opened 2020-03-25
Problem (Fix #1296): sgx-cargo-1804-hw1 is not easy to run locally

### LINKED ISSUE #1296: Problem: sgx-cargo-1804-hw1 is not easy to run locally
Maybe move it into default pipeline, or change pipeline type to default.
Easier to test locally with drone-cli.

yihuang:
Solution:
    - Move it into default pipeline

```
drone exec --trusted --include sgx-test
```


tomtau:
bors try

PR #1309, opened 2020-03-26
Problem: enclave code directly contacts IAS

### LINKED ISSUE #817: Problem: enclave code directly contacts IAS

devashishdxt:
Solution: Added `ra-sp-server` (proxy to IAS and AESM) and `ra-sp-client` (utility to connect to `ra-sp-server`). Fixes #817.

---
This change is [Reviewable](https://reviewable.io/reviews/crypto-com/chain/1309)


chain-tx-enclave/enclave-ra/ra-common/src/quote/quote_body.rs
```diff
@@ -0,0 +1,73 @@
...
+#[derive(Debug)]
+pub struct QuoteBody {
+    /// Version of the quote structure
+    pub version: u16,
+    /// Signature type of quote
+    pub sig_type: u16,
+    /// ID of the Intel EPID group of the platform belongs to
+    pub gid: u32,
+    /// Security version number of Quoting Enclave
+    pub qe_svn: u16,
```
leejw51crypto: how full name?
quoting_enclave_version?

chain-tx-enclave/enclave-ra/ra-sp-server/src/ias_client.rs
```diff
@@ -0,0 +1,142 @@
...
+            ias_key,
+            http_client,
+            ias_base_uri,
+            ias_sig_rl_path,
+            ias_report_path,
+        }
+    }
+
+    /// Gets SigRL (Signature revocation list) from IAS
+    pub fn get_sig_rl(&self, gid: [u8; 4]) -> Result<Option<Vec<u8>>, IasClientError> {
```
leejw51crypto: how about
get_signature_revocation_list

devashishdxt: All the names corresponds to Intel IAS API Specifications. Intel calls these "SigRL". So, I gess, it is best to stay true to specs.

PR #1310, opened 2020-03-26
Problem: incorrect ordering of keys() in MemoreyStorage#1294

leejw51crypto:
Solution: replace hashmap with btreemap
memory stroages used hashmap, which keys ordering is different.

so in some integration test(multinode), address list order is very important, 
hashmap can break it sometimes.

solution is to replace hashmap with btreemap, which is the same with sled or rocksdb.

### DIFF
--- a/client-common/src/storage/memory_storage.rs
+++ b/client-common/src/storage/memory_storage.rs
@@ -1,4 +1,4 @@
-use std::collections::HashMap;
+use std::collections::BTreeMap;
 use std::sync::{Arc, RwLock};
 
 use crate::{Error, ErrorKind, Result, Storage};
@@ -6,7 +6,7 @@ use crate::{Error, ErrorKind, Result, Storage};
 /// Storage backed by `HashMap`
 #[allow(clippy::type_complexity)]
 #[derive(Debug, Default, Clone)]
-pub struct MemoryStorage(Arc<RwLock<HashMap<Vec<u8>, HashMap<Vec<u8>, Vec<u8>>>>>);
+pub struct MemoryStorage(Arc<RwLock<BTreeMap<Vec<u8>, BTreeMap<Vec<u8>, Vec<u8>>>>>);
 
 impl Storage for MemoryStorage {
     fn clear<S: AsRef<[u8]>>(&self, keyspace: S) -> Result<()> {
@@ -18,7 +18,7 @@ impl Storage for MemoryStorage {
         })?;
 
         if let Some(ref mut space) = memory.get_mut(keyspace.as_ref()) {
-            space.drain();
+            space.clear();
         }
 
         Ok(())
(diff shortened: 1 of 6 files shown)


leejw51crypto:
yes

devashishdxt:
> Solution: replace hashmap with btreemap
> memory stroages used hashmap, which keys ordering is different.
> 
> so in some integration test(multinode), address list order is very important,
> hashmap can break it sometimes.
> 
> solution is to replace hashmap with btreemap, which is the same with sled or rocksdb.

Is there any guaruntee that `sled` and `rocksdb` will always store values in an order. I don't think our code/tests should depend on the ordering of keys in database because databases will not provide these guaruntees.

leejw51crypto:
in multinode.py, we use ordering in address list which is done by IndexSet previously.

devashishdxt:
> in multinode.py, we use ordering in address list which is done by IndexSet previously.

If it is already handled by `IndexSet`, why are we changing it?

leejw51crypto:
yes, IndexSet is removed
it's fetched directly from storage

tomtau:
> > Solution: replace hashmap with btreemap
> > memory stroages used hashmap, which keys ordering is different.
> > so in some integration test(multinode), address list order is very important,
> > hashmap can break it sometimes.
> > solution is to replace hashmap with btreemap, which is the same with sled or rocksdb.
> 
> Is there any guaruntee that `sled` and `rocksdb` will always store values in an order. I don't think our code/tests should depend on the ordering of keys in database because databases will not provide these guaruntees.

in RocksDB, it's according to the user function/comparator, so it can be guaranteed.
in sled, I don't know -- there's this note in README: https://github.com/spacejam/sled#a-note-on-lexicographic-ordering-and-endianness
I guess sled will try to preserve/guarantee lexicographic ordering? As that previous PR used `le` (to address my comment whether `be` needs to be there), it may be messed up with more than 256 items

devashishdxt:
> yes, IndexSet is removed
> it's fetched directly from storage

I'm a bit confused. Are you saying that if I iterate over all the keys in a specific keyspace, our code expects those keys in a specific order? If yes, we shouldn't depend on order of keys in database and if there is not other way, we should handle ordering ourselves (by using something like `IndexSet`) instead of depending on storage engine.

leejw51crypto:
yes, i think sled will be the same with rocksdb.

but for sled, sled supports in-memory style,
so we can wrap it up with in-memory sled.

leejw51crypto:
> > yes, IndexSet is removed
> > it's fetched directly from storage
> 
> I'm a bit confused. Are you saying that if I iterate over all the keys in a specific keyspace, our code expects those keys in a specific order? If yes, we shouldn't depend on order of keys in database and if there is not other way, we should handle ordering ourselves (by using something like `IndexSet`) instead of depending on storage engine.

keys() should behave the same with real db
for real db,keys are alphabet order

leejw51crypto:
current iterating contains bugs 
i found by checking  @devashishdxt  idea
thanks!

i'm modifying now

leejw51crypto:
i'll add unit-test

client-common/src/storage/memory_storage.rs
```diff
@@ -151,3 +151,43 @@ impl Storage for MemoryStorage {
...
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use blake2::Blake2s;
+    use chain_core::common::hash256;
+
+    #[test]
+    fn check_memorystorage_ordering() {
```
tomtau: this doesn't check the order against sled or what the problem was?

leejw51crypto: it was in memory storage only, sled has no problem.
we didn't use keys() for ordering previously, there was no issue before.

leejw51crypto:
ok, i will add for sled stroage ,too.

yihuang:
Since we don't remove the `IndexSet`, is there other reason to rely on key ordering of storage?

leejw51crypto:
currently, we don't use keys ordering,
but behaviours are different between MemoryStorage and SledStorage

if this bug is fixed, we can use keys() for iteration.
source can be simpler

yihuang:
> currently, we don't use keys ordering,
> but behaviours are different between MemoryStorage and SledStorage
> 
> if this bug is fixed, we can use keys() for iteration.
> source can be simpler

I mean, rely on key ordering might be an issue in the future, when upgrade or switch to other db.

leejw51crypto:
yes,
because users will expect the same behaviour from memorystorage, sledstorage.

yihuang:
> yes,
> because users will expect the same behaviour from memorystorage, sledstorage.
> 
> also for db, hashkey is un-common,
> because physical db page is too sparse.
> 
> switching to other db is beyond scope of this issue.

I mean keep the key ordering different between memory storage and sled storage maybe a good thing, if we don't intent to rely on the ordering of the storage, it can help us to identify problems.

leejw51crypto:
ok, i'll add to comments

tomtau:
> keep the key ordering different between memory storage and sled storage maybe a good thing

I'll close this for now, as it's unclear whether the stated problem is really a problem

Issue #802, opened 2020-01-06
Problem: non-stable staked state merkle trie storage


tomtau:
Currently, states are persisted and their integrity is checked using `starling` which is maintained by a single person and seems more of a hobby project.

Solution options:

1. vendor-in code, verify and maintain it ourselves

2. migrate to some more robust solutions (iAVL+, jellyfish merkle tree: https://github.com/libra/libra/blob/master/storage/jellyfish-merkle/src/lib.rs#L6 ...) -- write ADR for that first

yihuang:
The `jellyfish-merkle` seems nicely designed, also storage agnostic, there should be no headaches to switch to that.

tomtau:
There's also https://github.com/nomic-io/merk but it's +- similar situation as with `starling`.

`jellyfish-merkle` is nice -- we'd need to adapt the code though, as otherwise it'd bring loads of (potentially incompatible) libra-only dependencies

tomtau:
After the examination of code by @devashishdxt, the best tradeoff may be the option 1 to vendor in the code -- currently forked to here: https://github.com/crypto-com/merkle_bit

tomtau:
(we may maintain it in that repository, or if it diverges too much, incorporate it directly in chain repo)

PR #1330, opened 2020-03-31
Problem: non-stable merkle trie library #802

yihuang:
Solution:
- Write an adr for migrate to jellyfish-merle-tree

### DIFF
--- a/architecture-docs/adr-003.md
+++ b/architecture-docs/adr-003.md
@@ -1,0 +1,169 @@
+# ADR 003: Migrate to jellyfish merkle trie
+
+## Changelog
+* 02-04-2020: Initial Draft
+
+## Context
+
+Need a merkle trie library to store staking state, current one has several disadvantages:
+
+- Written and maintained by single person.
+- Custom data structure and algorithm.
+- Not enough documentation.
+
+## Decision
+
+Choose [jellyfish merkle trie](https://github.com/libra/libra/blob/master/storage/jellyfish-merkle/src/lib.rs) as the
+merkle trie solution.
+
+#### Advantages
+
+- Review from Facebook security engineers
+- Small code base, well documented.
+- Well known data structure (simplified version of ethereum MPT).
+- Easy to integrate.
+- Optimized for pbft: Use u64 version number instead of hash.
+- License is apache 2.0.
+- Support inclusion proof, exclusion proof and range proof.
+
+Maintain a custom fork of libra repository, make nessesary changes to keep jellyfish-merkle separated from other part
+of the project.
+
+> Initial work has done here: https://github.com/yihuang/libra
+
+### Integration
+
+- Add a column `COL_TRIE_NODE` to store trie nodes in the same key-value store as other states.
+
+- Implement `jellyfish_merkle::TreeReader` for `GetKV`.
+
+  ```rust
+  struct KVReader<'a, S: GetKV>(&'a S);
+  impl<'a, S: GetKV> TreeReader for KVReader<'a, S> {
+      fn get_node_option(&self, node_key: &NodeKey) -> Result<Option<Node>> {
+          self.0
+              .get(&(COL_TRIE_NODE, node_key.encode()?))
+              .map(|bytes| Node::decode(&bytes))
+              .transpose()
+      }
+  
+      fn get_rightmost_leaf(&self) -> Result<Option<(NodeKey, LeafNode)>> {
+          unimplemented!("this feature is only used in merkle tree restore which we don't need yet");
+      }
+  }
+  ```
+
+- Implement `StakingGetter` with `JellyfishMerkleTree` against any `GetKV`,
+
+  Use `BlockHeight` as the version number.
+  
+  ```rust
+  pub struct StakingGetter<'a, S: GetKV> {
+      storage: &'a S,
+      block_height: BlockHeight,
+  }
+  
+  impl<'a, S: GetKV> Get for StakingGetter<'a, S> {
+      type Key = StakedStateAddress;
+      type Value = StakedState;
+      fn get(&self, key: &Self::Key) -> Option<Self::Value> {
+          JellyfishMerkleTree::new(&KVReader::new(self.storage))
+              .get_with_proof(HashValue::new(to_stake_key(key)), self.block_height.into())
+              .expect("merkle trie internal error")
+              .0
+              .map(|blob| {
+                  StakedState::decode(&mut blob.as_ref()).expect("merkle trie storage corrupted")
+              })
+      }
+  }
+  ```
+  
+- Use the same buffer abstraction as before:
+  
+  ```rust
+  pub type StakingBufferStore<'a, S, H> = BufferSimpleStore<'a, StakingGetter<'a, S>, H>;
+  pub type StakingBufferGetter<'a, S, H> = BufferGetter<'a, StakingGetter<'a, S>, H>;
+  ```
+  
+- When commit, flush the staking buffer into the Merkle trie, write the nodes into the `StoreKV`, which is the buffered
+  key-value storage shared with other states.
+  
+  ```rust
+  pub fn flush_stakings<S: StoreKV>(
+      storage: &mut S,
+      block_height: BlockHeight,
+      buffer: StakingBuffer,
+  ) -> Result<(usize, usize)> {
+      let reader = KVReader::new(storage);
+      let tree = JellyfishMerkleTree::new(&reader);
+      let (_, batch) = tree.put_blob_sets(
+          vec![buffer
+              .values()
+              .map(|staking| (HashValue::new(staking.key()), staking.encode().into()))
+              .collect::<Vec<_>>()],
+          block_height.into(),
+      )?;
+      for (key, node) in batch.node_batch.iter() {
+          storage.set((COL_TRIE_NODE, key.encode()?), node.encode()?);
+      }
+      storage.set(
+          (COL_STALED_NODE, block_height.encode()), 
+          batch.stale_node_index_batch.encode()
+      );
+      Ok((batch.num_new_leaves, batch.num_stale_leaves))
+  }
+  ```
+  
+- A complete commit would be like this:
+  
+  ```rust
+  flush_staking(&mut kv_store!(self), block_height, staking_buffer);
+  flush_kv(&mut self.kvdb, kv_buffer);  // Write to the disk atomically.
+  ```
+
+### Chain node state changes
+
+- Remove `AccountStorage` from `ChainNodeApp`.
+
+- In `ChainState`, change `account_root` to `last_account_block_height`, which records the last block height there's any
+  staking state modified.
+  
+  Use the block height to query history state directly.
+
+### Staled nodes cleanup
+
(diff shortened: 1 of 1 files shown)


architecture-docs/adr-003.md
```diff
@@ -0,0 +1,54 @@
...
+- Maintain a custom repository with code separated from libra repo.
+
+  Initial work has done here: https://github.com/yihuang/libra-jellyfish-merkle
+
+  Sync the code periodically with script.
+
+- Use version number to replace root hash, version number is current block height.
+
+- Integrate with buffer storage abstraction, so underlying key value store is possible persist all states atomically in
+  single transaction.
```
tomtau: > This section explains all of the details of the proposed solution, including implementation details. It should also describe affects / corollary items that may need to be changed as a part of this. If the proposed change will be large, please also indicate a way to do the change to maximize ease of review. (e.g. the optimal split of things to do between separate PR's)

https://github.com/crypto-com/chain/blob/master/architecture-docs/adr-template.md#decision

architecture-docs/adr-003.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Other candidates
+
+TODO
+
+## Status
```
tomtau: missing consequences + references sections

architecture-docs/adr-003.md
```diff
@@ -0,0 +1,54 @@
...
+- Support inclusion proof and range proof.
+
+#### Disadvantages
+
+- Tied with a lot of custom Libra crates, need to maintain a separated library.
+- No extension node compared to ethereum MPT, deeper tree in theory, a trade off made in favor of simplicity.
+
+### Integration
+
+- Maintain a custom repository with code separated from libra repo.
```
tomtau: more details needed -- e.g. are all the other crates like https://github.com/yihuang/libra-jellyfish-merkle/tree/master/crypto and all its transitive dependencies needed?
If this is a custom version updated with a script (instead of git merging), why not to include it in the main repo + have the rsync script?

calvinaco: One of the concern would be whether it is manageable to keep in sync with the master from time-to-time. I have a quick look on the commit history, it seems it is relatively stable since Dec.

Do you know what is their plan? Is the implementation freezed?

tomtau: not sure if there's a public code roadmap, but seems fairly stable
As for signatures -- they shouldn't be needed for the jellyfish-merkle anyway?

calvinaco: Oh sorry, for signature I am referring to the method signature. But yes  it seems to be relatively stable.

yihuang: From what I see, there is actually no clear winner that is good enough for us to use as a black box. Even if it's written by Facebook, it's still pre-alpha software in theory. And the code syncing and patching process is also error prone.
I think no matter we choose jellyfish or starling (or other solution), we need to write down the algorithm as a specification, to make sure we all understand the implications.

calvinaco: I am not sure if this will be overkilled. But can we implement an extra abstraction layer between our code and whatever lib we decided to use? That way we are free from changes from the underlying library.

When there are breaking changes we will still be affected, but we lock down the potion of code we will have to change.

tomtau: > I am not sure if this will be overkilled. But can we implement an extra abstraction layer between our code and whatever lib we decided to use? That way we are less dependent with changes from the underlying library.
> 
> When there are breaking changes we will still be affected, but at least we lock down the portion of code we will have to change.

That's already kind of done -- "persistent storage" is abstracted out in `kvdb`; "updates and lookups from merkle trie" are in buffer. 
but it should be documented in ADR how exactly this other crate is planned to be integrated in -- https://github.com/crypto-com/chain/pull/1330#discussion_r400715258

yihuang: > more details needed -- e.g. are all the other crates like https://github.com/yihuang/libra-jellyfish-merkle/tree/master/crypto and all its transitive dependencies needed?
> If this is a custom version updated with a script (instead of git merging), why not to include it in the main repo + have the rsync script?

We can remove these dependencies like crypto/crypto-derive/lcs by extracting needed code into the merkle-trie crate, the result is easier to maintain by us, but harder to sync code with upstream automatically.

tomtau: > the result is easier to maintain by us, but harder to sync code with upstream automatically.

that's fine -- but should be mentioned in the impact; even with a script, it may become hard with some external repo restructuring (as opposed to maintaining a fork and using git merges/patches)

yihuang: Updated

- Use a fork of the whole libra repository, more convenient to sync the code in the future with git.
- Describe more detailed integration work.

architecture-docs/adr-003.md
```diff
@@ -0,0 +1,144 @@
+# ADR 003: Migrate to jellyfish merkle trie
+
+## Changelog
+* 31-03-2020: Initial Draft
```
tomtau, line 4: there were changes from the initial write up?

architecture-docs/adr-003.md
```diff
@@ -0,0 +1,144 @@
...
+- License is apache 2.0.
+- Support inclusion proof and range proof.
+
+#### Disadvantages
+
+- Tied with a lot of custom Libra crates, need to extract an independent library.
+
+  Initial work has done here: https://github.com/yihuang/libra
+
+- No extension node compared to ethereum MPT, deeper tree in theory, a trade off made in favor of simplicity.
```
tomtau, line 36: this is not a "decision" -- this is like a context mixed with consequences ?

architecture-docs/adr-003.md
yihuang: The integration code:
https://github.com/yihuang/chain/blob/issue802/chain-storage/src/jellyfish.rs

architecture-docs/adr-003.md
```diff
@@ -0,0 +1,144 @@
...
+  ```rust
+  pub struct StakingGetter<'a, S: GetKV> {
+      storage: &'a S,
+      block_height: BlockHeight,
+  }
+  
+  impl<'a, S: GetKV> Get for StakingGetter<'a, S> {
+      type Key = StakedStateAddress;
+      type Value = StakedState;
+      fn get(&self, key: &Self::Key) -> Option<Self::Value> {
```
tomtau, line 78: what about getting/returning the proof?

architecture-docs/adr-003.md
```diff
@@ -0,0 +1,144 @@
...
+  ```
+
+## Status
+
+Proposed
+
+## Consequences
+
+- Need to merge the code from upstream with git manually periodically.
+
```
tomtau, line 140: > This section describes the consequences, after applying the decision. All consequences should be summarized here, not just the "positive" ones.

architecture-docs/adr-003.md
```diff
@@ -0,0 +1,144 @@
...
+#### Disadvantages
+
+- Tied with a lot of custom Libra crates, need to extract an independent library.
+
+  Initial work has done here: https://github.com/yihuang/libra
+
+- No extension node compared to ethereum MPT, deeper tree in theory, a trade off made in favor of simplicity.
+
+### Integration
```
tomtau, line 38: how will this be done? one giant PR or PR for adding the crate and then PR for implementing readers etc.? or ?

architecture-docs/adr-003.md
yihuang: I think no need to track changed not yet committed?

architecture-docs/adr-003.md
yihuang: The whole section of "Disadvantages" can be moved to consequences then.

architecture-docs/adr-003.md
yihuang: > what about getting/returning the proof?

Just use `JellyfishMerkleTree` api directly then.

architecture-docs/adr-003.md
yihuang: I estimates that one PR should be good enough, won't be too big.

architecture-docs/adr-003.md
yihuang: Done

architecture-docs/adr-003.md
yihuang: Done

architecture-docs/adr-003.md
tomtau: should be mentioned.

Also "ChainState" integrations -- will/can this support historical querying? 
My assumption is that this can support historical querying, but instead of using the root hash, one can use the "u64" version number?

yihuang: > should be mentioned.
> 
> Also "ChainState" integrations -- will/can this support historical querying?
> My assumption is that this can support historical querying, but instead of using the root hash, one can use the "u64" version number?

Yes, use the block height as the version number.

tomtau: so I guess that's a positive impact that there's no need to store that `APP_HASH -> block height` mapping for historical querying?

yihuang: > so I guess that's a positive impact that there's no need to store that `APP_HASH -> block height` mapping for historical querying?

Yes, I'll add that.

yihuang: > should be mentioned.
Done.
Added: "Implementation plan", "Staled nodes cleanup strategy", "Chain node state changes",

yihuang: Changed the implementation plans to use two PRs.

architecture-docs/adr-003.md
```diff
@@ -0,0 +1,168 @@
...
+
+- In `ChainState`, change `account_root` to `last_account_block_height`, which records the last block height there's any
+  staking state modified.
+  
+  Use the block height to query history state directly.
+
+### Staled nodes cleanup strategy
+
+- The keys of staled nodes are stored in column `COL_STALED_NODE`, indexed by block height.
+- Periodically cleanup the nodes for block heights too old by iterate keys in colume `COL_STALED_NODE`.
```
tomtau, line 136: What are "staled nodes"?

yihuang: > What are "staled nodes"?

The nodes not used in newer block heights of the tree, only referenced by old block heights.
It's possible to clean them up, if we don't want to keep all the old histories.

tomtau: it'll be good to explain this in the doc -- with starling, I think it's also possible to prune unused nodes, but it wasn't done.
It'll make sense to do this in the "validator" setup (where one just cares about the latest state in order to validate new transactions), but for "full node" setup, one will want to keep them, as one may need to query a historical state with a proof.

Not sure how much extra complexity it is -- one particular thing here would be whether "integration test' would be possible (to test that the cleanup mechanism doesn't accidentally cause problems)... perhaps it may be easier to leave it out first? The setting for cleanup can be in a different PR?

yihuang: I'll mark it as a potencial feature for the future, we keep all the old states for now.

tomtau:
ok, will merge -- I'll create a follow up PR with a few polishes for this ADR later + any final comments

PR #1331, opened 2020-03-31
Problem: inclusion proof does not verify that hash(leaf_value) == path.leaf_hash

### LINKED ISSUE #1329: Problem: inclusion proof does not verify that hash(leaf_value) == path.leaf_hash
One can manually construct and encode a proof where value does not match the leaf hash, and it'd still verify.

devashishdxt:
Solution: Added a check to verify that `hash(leaf_value) == path.leaf_hash`. Fixes #1329

### DIFF
--- a/chain-core/src/common/merkle_tree.rs
+++ b/chain-core/src/common/merkle_tree.rs
@@ -209,8 +209,9 @@ impl<T> Proof<T> {
     where
         T: AsRef<[u8]>,
     {
-        // FIXME: https://github.com/crypto-com/chain/issues/1329
-        root_hash == self.path.hash() && self.path.verify()
+        root_hash == self.path.hash()
+            && self.path.verify()
+            && hash_leaf(&self.value) == self.path.leaf_hash
     }
 
     /// Returns a borrow of value contained in this proof
--- a/chain-core/src/common/merkle_tree.rs
+++ b/chain-core/src/common/merkle_tree.rs
@@ -809,4 +809,17 @@ mod tests {
             .unwrap()
             .verify(&new_tree.root_hash()));
     }
+
+    #[test]
+    fn check_wrong_leaf_value() {
+        let values = vec!["one", "two", "three", "four"];
+        let tree = MerkleTree::new(values);
+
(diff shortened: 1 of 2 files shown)


devashishdxt:
>     * isn't encoding/decoding "leaf_hash" redundant if it needs to be computed anyway -- could `leaf_hash` be removed and this is passed to self.path.verify(hash_leaf(&self.value)) directly?

This'll be a big change as other methods (`Path::hash()`) also depend on `leaf_hash` value stored in `Path`. We can create a new issue for this if we want to change the API design of `MerkleTree`.

devashishdxt:
>     * for the internal checkings / comparisons, it should ideally use [Hash](https://docs.rs/blake3/0.3.0/blake3/struct.Hash.html) instead of `[u8; 32]`, because Hash has constant-time equality checking -- this is important for https://github.com/crypto-com/chain/blob/master/chain-tx-validation/src/witness.rs#L22 where comparisons on `[u8; 32]` may time-leak the address

Created a new issue for this: https://github.com/crypto-com/chain/issues/1332

chain-core/src/common/merkle_tree.rs
```diff
@@ -808,4 +809,17 @@ mod tests {
...
+    #[test]
+    fn check_wrong_leaf_value() {
+        let values = vec!["one", "two", "three", "four"];
+        let tree = MerkleTree::new(values);
+
+        let mut proof = tree.generate_proof("one").unwrap();
+        assert!(proof.verify(&tree.root_hash()));
+
+        // Intentionally change the value in proof
+        proof.value = "two";
```
devashishdxt: Note that it is not possible write this line of code from outside of this module because value is a private member of `Proof`. So, there is no way to change the value of a proof or create an incorrect proof in current public API.

tomtau: it's possible with "encode" and directly manipulating the bytes

devashishdxt: Yeah. Anyway, we have this check in place now.

Issue #1335, opened 2020-04-01
Problem: no guide for cbindings android compile on mac


leejw51crypto:
Solution: add guide

leejw51crypto:
started

PR #1336, opened 2020-04-01
Problem: (cro-1335) no guide for cbindings android compile on mac #1335

leejw51crypto:
Solution: add document

### DIFF
--- a/cro-clib/android.md
+++ b/cro-clib/android.md
@@ -18,6 +18,7 @@ export NDK_HOME=$ANDROID_HOME/ndk/21.0.6113669
 export RUST_ANDROID=$HOME/rust_android/NDK
 ```
 
+
 ## for simple compile
 ```
 make android
--- a/cro-clib/android_mac.md
+++ b/cro-clib/android_mac.md
@@ -1,0 +1,86 @@
+## install android for mac
+1. install android studio
+2. install ndk, sdk(api 26)
+3. locations as below  
+```
+sdk:  $HOME/Library/Android/sdk
+ndK:  $HOME/Library/Android/sdk/ndk/21.0.6113669
+```
+
+## modify path 
+to make it compatible with linux script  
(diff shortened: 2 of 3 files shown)


cro-clib/android.md
```diff
@@ -18,6 +18,15 @@ export NDK_HOME=$ANDROID_HOME/ndk/21.0.6113669
...
 ```
 
+## check crate-type
+1. vi Cargo.toml  in cro-clib folder
+2. modify crate-type as staticlib only  
+```
+[lib]
+name = "cro_clib"
+crate-type =["staticlib"]
+```
```
tomtau, line 28: Can this change be made directly in Cargo.toml? Most platforms need `staticlib`, so I guess it'll make sense to make it a default crate-type, so it's not necessary for someone to modify Cargo.toml (if someone really needs `cdylib` for some reason, they can tweak it... but for 99% cases, `staticlib` should be fine?)

tomtau: iOS also needs staticlib?

cro-clib/android_mac.md
```diff
@@ -0,0 +1,95 @@
...
+
+
+## check crate-type
+1. vi Cargo.toml  in cro-clib folder
+2. modify crate-type as staticlib only  
+```
+[lib]
+name = "cro_clib"
+crate-type =["staticlib"]
+```
```
tomtau, line 50: can it be without this step if Cargo.toml is changed?

cro-clib/android.md
leejw51crypto: yes, both staticlib
in ios, all libs are static

i'll modify

staticlib is find & stable

cro-clib/android_mac.md
leejw51crypto: i'll remove it

PR #1340, opened 2020-04-01
Problem: no Java example for C bindings#1236

leejw51crypto:
Solution: add java example


tomtau:
to be consistent with iOS etc., it'll be better in an external repository -- https://github.com/crypto-com/sample-chain-java-example

PR #1368, opened 2020-04-03
Problem (WIP #802): non-stable staked state merkle trie storage

yihuang:
Solution:
- Add jellyfish module into `chain-storage`
  step1 of adr-003

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -133,6 +133,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "anyhow"
+version = "1.0.28"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d9a60d744a80c30fcb657dfe2c1b22bcb3e814c1a1e3674f32bf5820b570fbff"
+
+[[package]]
 name = "arrayref"
 version = "0.3.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


Cargo.lock
```diff
@@ -2830,6 +3060,47 @@ dependencies = [
...
+
+[[package]]
+name = "prost-types"
+version = "0.6.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1834f67c0697c001304b75be76f67add9c89742eda3a085ad8ee0bb38c3417aa"
+dependencies = [
+ "bytes 0.5.4",
+ "prost",
+]
```
tomtau, line 3102: is this necessary?

Cargo.lock
```diff
@@ -2183,6 +2257,80 @@ dependencies = [
...
+ "libra-crypto-derive",
+ "libra-nibble",
+ "mirai-annotations",
+ "once_cell",
+ "rand 0.6.5",
+ "serde",
+ "sha2",
+ "thiserror",
+ "tiny-keccak",
+]
```
tomtau, line 2289: is this necessary?

yihuang:
> this PR should come after the "jellyfish-merkle" is cleaned up / properly isolated and doesn't bring unnecessary dependency bloat -- we'll want to keep using the same binary codec (scale) + cryptography (blake3)

Ok

PR #1391, opened 2020-04-07
Problem: Tendermint RPC client does not support websocket over TLS

### LINKED ISSUE #1346: Problem: Tendermint websocket client does not support connecting over TLS

devashishdxt:
Solution: Moved to `tokio-tungstenite` with support for websockets over TLS. Fixes #1346.


devashishdxt:
bors try

client-common/src/tendermint/rpc_client/async_rpc_client.rs
```diff
@@ -0,0 +1,262 @@
...
+        Ok(responses)
+    }
+
+    /// Sends a JSON-RPC request and returns `request_id` and `response_channel`
+    async fn send_request(
+        &self,
+        method: &str,
+        params: &[Value],
+    ) -> Result<(String, Receiver<JsonRpcResponse>)> {
+        let id = self.unique_id.fetch_add(1, Ordering::Relaxed).to_string();
```
tomtau, line 178: I guess relaxed is fine here? Or would random 128-bit UUID make it more robust?

client-common/src/tendermint/rpc_client/async_rpc_client.rs
```diff
@@ -0,0 +1,262 @@
...
+    ) -> Result<Value> {
+        let response = timeout(RESPONSE_TIMEOUT, receiver)
+            .await
+            .context("Tendermint RPC request timed out")?;
+
+        let response = match response.context("Unable to receive message from channel receiver") {
+            Ok(response) => response,
+            Err(err) => {
+                self.channel_map.lock().await.remove(id);
+                bail!(err)
```
tomtau, line 220: those existing error chaining helpers in client-* don't work in async?

client-common/src/tendermint/rpc_client/sync_rpc_client.rs
```diff
@@ -0,0 +1,287 @@
...
+    /// Makes an RPC call and deserializes response
+    pub fn call<T>(&self, method: &'static str, params: Vec<Value>) -> Result<T>
+    where
+        T: Send + 'static,
+        for<'de> T: Deserialize<'de>,
+    {
+        let (sender, receiver) = sync_channel(1);
+        let async_rpc_client = self.async_rpc_client.clone();
+
+        self.runtime.spawn(async move {
```
tomtau, line 61: should this use `spawn` and not `block_on` or something like that? I think for `spawn`, if some mid-step panics, it may keep hanging?

client-common/src/tendermint/rpc_client/sync_rpc_client.rs
```diff
@@ -0,0 +1,287 @@
...
+    /// Makes RPC call in batch and deserializes responses
+    pub fn call_batch<T>(&self, params: Vec<(&'static str, Vec<Value>)>) -> Result<Vec<T>>
+    where
+        T: Send + 'static,
+        for<'de> T: Deserialize<'de>,
+    {
+        let (sender, receiver) = sync_channel(1);
+        let async_rpc_client = self.async_rpc_client.clone();
+
+        self.runtime.spawn(async move {
```
tomtau, line 88: as before

client-common/src/tendermint/rpc_client/async_rpc_client.rs
devashishdxt: Yes. `client-common::Error` is not `Send + Sync`. We can slowly start moving everything to `anyhow` (in `client-*`) or `thiserror` (in `chain-*`).

client-common/src/tendermint/rpc_client/sync_rpc_client.rs
devashishdxt: Ideally we should use something like `block_on` here, but not `block_on`, because `block_on` takes mutable reference of `runtime` that means that we cannot `block_on` two different things at the same time (in our case, that'll mean that we cannot make two requests in parallel). The current implementation is mimicking `block_on` with non-mutable reference of `runtime`.

client-common/src/tendermint/rpc_client/sync_rpc_client.rs
devashishdxt: https://github.com/crypto-com/chain/pull/1391/files#r404559510

client-common/src/tendermint/rpc_client/async_rpc_client.rs
devashishdxt: I think this is fine. Didn't cause any problem until now.

client-common/src/tendermint/rpc_client/sync_rpc_client.rs
tomtau: what about `block_on_all`?

devashishdxt: I don't know of any `block_on_all` function. Can you post docs link here?

tomtau: oh, it's 0.2

tomtau: if you do say `panic!("test")` after response, the whole process will fail, right?

devashishdxt: I've added a `recv_timeout` when reading from response channel. That should be enough?

tomtau: did you try with that `panic!`?

devashishdxt: Where do you want to put `panic!`?

devashishdxt: I'm a bit confused. Are you worried about `async` code panicking or are you worried about it getting stuck when it panics. After `recv_timeout` change, if the `async` code panics, user will see a `"Request timed out"` error.

tomtau: e.g. between lines 64 and 65

devashishdxt: As I mentioned earlier, if `async` code panics, user will see `"Request timed out"` error.

tomtau: so the process will keep hanging even though it's not recoverable?
that's not ideal, but if there's no way around it

devashishdxt: I'm still not able to understand how the async code can panic? There is not a single `unwrap` or `expect` in async code. If you're worried about someone making changes in future and putting an `unwrap` in async code (which ideally shouldn't be done and we should use proper error handling), then yes, user will keep getting this error when internal code panics.

tomtau: it's not only the direct code here, but all the dependencies -- the websocket library, tendermint-rs, parsers... 
if the process exited, it'd be faster to find out it went wrong because of this (rather than diagnosing tendermint startup etc.)

devashishdxt: Yeah. This should ideally be done with panic-hooks. I'll investigate about this and create a new PR with improvements (if there are some).

tomtau:
seems the websocket failing is still there?

tomtau:
```

thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Tendermint RPC error: Unable to connect to tendermint RPC websocket at: ws://127.0.0.1:26657/websocket
--
547 | => Unable to connect to tendermint websocket server at: ws://127.0.0.1:26657/websocket', client-rpc/src/program.rs:78:5
548 | stack backtrace:


```

devashishdxt:
> was it that the old websocket library had a longer timeout or for the first connection, it waited indefinitely for Tendermint RPC?

Timeout and behaviour is unchanged. We try to reconnect with tendermint RPC if the connection fails. I'm checking the logs to get the root cause of test failure.

tomtau:
> > was it that the old websocket library had a longer timeout or for the first connection, it waited indefinitely for Tendermint RPC?
> 
> Timeout and behaviour is unchanged. We try to reconnect with tendermint RPC if the connection fails. I'm checking the logs to get the root cause of test failure.

what about the first connection, not the reconnection? I don't mean the code in client-*, but the underlying library's behaviour

devashishdxt:
bors try

devashishdxt:
> > > was it that the old websocket library had a longer timeout or for the first connection, it waited indefinitely for Tendermint RPC?
> > 
> > 
> > Timeout and behaviour is unchanged. We try to reconnect with tendermint RPC if the connection fails. I'm checking the logs to get the root cause of test failure.
> 
> what about the first connection, not the reconnection? I don't mean the code in client-*, but the underlying library's behaviour

Underlying library does not provide any retry mechanism. We have a custom retry mechanism in `client-rpc` for first connection (later reconnections are handled by `websocket_rpc_loop`). [Here](https://github.com/crypto-com/chain/blob/49dee51dc388063102fdf821b040b21ad2602354/client-rpc/src/server.rs#L187) is the retry mechanism in `client-rpc`. My guess is that the `ErrorKind` returned by new websocket implementation is different. I changed it. Hope it works fine now.

devashishdxt:
@yihuang Any idea about this error:
```
[2020-04-07T09:41:29Z ERROR client_rpc::server] Validation error: Staking state is not yet unbonded
```

You can also check logs in drone.

yihuang:
> @yihuang Any idea about this error:
> 
> ```
> [2020-04-07T09:41:29Z ERROR client_rpc::server] Validation error: Staking state is not yet unbonded
> ```
> 
> You can also check logs in drone.

It seems logged by `client-network/src/network_ops/default_network_ops_client.rs#277`: 
```
        if staked_state.unbonded_from > last_block_time {
            return Err(Error::new(
                ErrorKind::ValidationError,
                "Staking state is not yet unbonded",
            ));
        }
```

yihuang:
> ```
> 
> thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Tendermint RPC error: Unable to connect to tendermint RPC websocket at: ws://127.0.0.1:26657/websocket
> --
> 547 | => Unable to connect to tendermint websocket server at: ws://127.0.0.1:26657/websocket', client-rpc/src/program.rs:78:5
> 548 | stack backtrace:
> ```

supervisors will restart process when panicked, so it should be no problem.

tomtau:
bors retry

devashishdxt:
Thanks @yihuang!

tomtau:
bors retry

PR #1401, opened 2020-04-07
Problem (Fix #1400) `get_last_block_time` returns zero before first block is generated

### LINKED ISSUE #1400: Problem: `get_last_block_time` returns zero before first block is generated
Solution:
- Fetch genesis_time in this case.

yihuang:
Solution:
- Fetch genesis_time in this case


devashishdxt:
This is the first occurance of this error on our CI: http://13.82.172.109/crypto-com/chain/317/1/5

tomtau:
that commit only updates docs -- before that SDK was updated to 2.9 + there were those old hyper-related fixes

yihuang:
> seems that failure still persists:
> 
> ```
> [2020-04-08T03:29:23Z ERROR client_rpc::server] Validation error: Staking state is not yet unbonded
> ...
> ```
> 
> does that multi-node integration test need some extra waiting time?

The problem is `Height::default` is 1 not 0 ;(

tomtau:
bors retry

PR #1419, opened 2020-04-10
Problem: jellyfish merkle tree don't allow empty version

yihuang:
Solution:
- Change adr-003 to use dedicated version number rather than block height.

The reason for this is the library doesn't allow empty versions, but we have empty blocks, so we can't use block height as the version number.
So we need to use a dedicated version number, it's also the number of blocks which have at least a staking modifications.

### DIFF
--- a/architecture-docs/adr-003.md
+++ b/architecture-docs/adr-003.md
@@ -4,6 +4,7 @@
 * 02-04-2020: Initial Draft
 * 02-04-2020: Minor Edits + Decision Updated
 * 03-04-2020: Root Hash Kept in Chain Node State (for light client verification)
+* 11-04-2020: Use dedicated version number rather than block height
 
 ## Context
 For the authenticated integrity-checking storage of staking states, the prototype code has been using the [starling](https://crates.io/crates/starling) crate that implements "Merkle Binary Indexed Tree". This was choice, however, temporary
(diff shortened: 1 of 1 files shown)


yihuang:
An alternative solution is just to modify the jellyfish Merkle tree, to allow empty versions. For empty versions, it'll just create a new root node which is the same as the previous one, only version number increased.

architecture-docs/adr-003.md
```diff
@@ -118,19 +119,21 @@ its nodes will be written into the `StoreKV` buffer, which is the buffered key-v
...
   ```
 
 ### Chain Node State Changes
 
 - `AccountStorage` will be removed from `ChainNodeApp` (as the node storage will be in the dedicated `COL_TRIE_NODE` column).
 
-- In ABCIQuery, the block height from the request will be used to query historical states directly.
+- The block height to root hash mapping in storage will be changed to block height to version number.
+
+  ABCIQuery will convert block height to version number first to query history state.
```
tomtau, line 132: Or rather, in that tx-query setting, it'll store the historical `ChainState`s, so it'll lookup chain state at that height and get its version?

architecture-docs/adr-003.md
```diff
@@ -164,6 +166,7 @@ Accepted
 (or equivalent deterministic-finality consensus algorithms), but for probabilistic-finality consensus algorithms
 - One more repository
 - One "rocksdb" instance (as the nodes will be stored in one column rather than in a separate database instance)
+- The block height to root hash mapping is changed to block height to version number for ABCI querying
```
tomtau, line 169: this one may be neutral, as other storage may need this, or I guess this is compared with the alternative of changing jellyfish implementation?

architecture-docs/adr-003.md
yihuang: Yes, but currently the state is only stored optionally.

architecture-docs/adr-003.md
yihuang: I changed it into:
```
When query history staking state, convert the request block height to version number through history ChainNodeState first
```
It is in neutral.

architecture-docs/adr-003.md
tomtau: yep -- this abciquery path will be for full nodes serving light clients

PR #1421, opened 2020-04-13
Problem: clippy/fmt didn't checked in drone

yihuang:
- Add the check before build


tomtau:
it's checked in Travis -- the purpose of drone pipelines is integration testing in HW, so no need to duplicate (there's other things beyond clippy/fmt to check)

PR #1428, opened 2020-04-14
Problem (Fix #1308): unnecessary item wrappers for merkle trie storage

### LINKED ISSUE #1308: Problem: unnecessary item wrappers for merkle trie storage
currently, there's https://github.com/crypto-com/chain/blob/master/chain-storage/src/account/mod.rs

after resolving https://app.zenhub.com/workspaces/chain-core-5dfc2dec79454cc329bf161c/issues/crypto-com/chain/1307 
it may be possibly to get this removed.
It may still be "valuable" to change `StakedState` encode/decode to prefix/tag it with byte "0" (enum-like) for future-proofing of the trie storage -- e.g. if in the future, there's some other state data (prefixed say with "1") that will be inserted/coexist in the trie (or just use a different trie for such case?)

yihuang:
Solution:
- Do the switch to jellyfish merkle library.


tomtau:
bors try

yihuang:
> for historical reasons, chain-abci aggregated some unused dependencies -- it still contains starling: https://github.com/crypto-com/chain/blob/master/chain-abci/Cargo.toml#L28 so can be removed ?

Done

Issue #1440, opened 2020-04-15
Problem: no progess api in sync rpc


leejw51crypto:
Solution: add asynchronous "SYNC" api and progress api

leejw51crypto:
done
submitted for code-review

PR #1443, opened 2020-04-15
Problem: no progess api in sync rpc #1440

leejw51crypto:
Solution: add asynchronous SYNC and  get_progress api
currently, when we sync with testnet, it took a lot of time.
so "run_sync" is added, sync is now can be done in multi-threaded,
for each wallet, one thread can be activated.

by run_sync_progress, we can get current progress.

moved current sync code by using spawning thread,
to get information, i used Arc and Mutex

React, or Angular now can poll via "run_sync_progress", update UI
when it's done, can close the progress UI.

"sync" api will behave exactly the same as before.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ## v0.4.0
 
 ### Breaking changes
+* *client-rpc* [1443](https://github.com/crypto-com/chain/pull/1443): add progress, and multi-thread to sync api
 * *chain-abci* [1239](https://github.com/crypto-com/chain/pull/1239): basic versioning
 * *chain-abci* [1090](https://github.com/crypto-com/chain/pull/1090): upper bounds of reward parameters changed
 * *chain-abci* [1100](https://github.com/crypto-com/chain/pull/1100): account nonce increased after reward distribution
--- a/client-rpc/src/rpc/sync_rpc.rs
+++ b/client-rpc/src/rpc/sync_rpc.rs
@@ -1,6 +1,5 @@
-use jsonrpc_core::Result;
-use jsonrpc_derive::rpc;
-
+use super::sync_worker::SyncWorker;
+use super::sync_worker::WorkerShared;
 use crate::server::to_rpc_error;
 use client_common::tendermint::Client;
 use client_common::Storage;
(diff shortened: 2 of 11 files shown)


tomtau:
bors try

yihuang:
Can you write a simple doc to explain how to use the new APIs?
And the old API like `sync_unlock_wallet` can be removed?

leejw51crypto:
sure,
sync_unlock_wallet is automatic one, it's necessary.

leejw51crypto:
sync_unlock_wallet is single-threaded,
different design

my approach, one thread for each wallet.
suitable for current web wallet

client-rpc/src/rpc/sync_rpc.rs
```diff
@@ -47,6 +76,126 @@ where
...
+        .map_err(to_rpc_error)
+}
+
+impl<S, C, O> SyncRpcImpl<S, C, O>
+where
+    S: Storage + 'static,
+    C: Client + 'static,
+    O: TransactionObfuscation + 'static,
+{
+    #[inline]
```
yihuang, line 154:
```suggestion
```

yihuang:
I think there is not much practical use case for `unlockWallet` anyway, so it might be better to focus on a set of more consistently designed APIs.

- `sync`, synchronized sync, sync to latest block then return.
- `sync_all`, change it to `sync` with a parameter.
- `run_sync`, synchronized version of sync.
- `run_sync_all`, can be changed to `run_sync` with a parameter.
- `run_sync_progress`, get the progress of `run_sync`.
- `unlockWallet`, it can be changed to `run_sync` with a parameter.

Also, I think we can rename `run_sync` to something more clear, `async_sync` maybe a bit weird, how about `spawn_sync`?

leejw51crypto:
spawn_sync seems better

unlockWallet : syncing in client-rpc side, i'm thinking multi-plexing as upgrade
receive block: update all info of wallets
user don't have to "spawn sync"

once sync is made, all sync is done via event emitted block. it's only one-way, so fast

yihuang:
How about something like this:
```
sync(wallet: WalletRequest, blocking: bool, reset: bool, loop: bool)
sync_progress(wallet: WalletRequest)
sync_stop(wallet: WalletRequest)
```

It can cover all the scenarios.

leejw51crypto:
it look OK
what do you think? @tomtau

tomtau:
3  api methods would be simpler and more consistent -- just update the changelog

leejw51crypto:
ok,
i'll modify to 
```
sync(wallet: WalletRequest, blocking: bool, reset: bool, loop: bool)
sync_progress(wallet: WalletRequest)
sync_stop(wallet: WalletRequest)
```

leejw51crypto:
done,
fixing integration test

tomtau:
bors try

leejw51crypto:
checking integration test

leejw51crypto:
rebased

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ## v0.4.0
 
 ### Breaking changes
+* *client-rpc* [1440](https://github.com/crypto-com/chain/pull/1440): add progress, and multi-thread to sync api
```
yihuang, line 8:
```suggestion
* *client-rpc* [1443](https://github.com/crypto-com/chain/pull/1443): add progress, and multi-thread to sync api
```

client-rpc/src/rpc/sync_worker.rs
```diff
@@ -0,0 +1,145 @@
...
+        } else {
+            1 // cotninue
+        }
+    }
+}
+pub type NodeShared = Arc<Mutex<SyncWorkerNode>>;
+
+#[derive(Default)]
+pub struct SyncWorker {
+    //  works: Vec<JoinHandle<()>>,
```
yihuang, line 82:
```suggestion
```

tomtau:
bors try

leejw51crypto:
ok

PR #1447, opened 2020-04-16
Problem: (fix #1242)client-rpc may include duplicate view keys

### LINKED ISSUE #1242: Problem: client-rpc may include duplicate view keys
1. unbond
2. withdraw
inside WithdrawUnbondedStakeTransaction,
TxAttributes-> allowed_view->TxAccessPolicy
two same policies, modify with just one

linfeng-crypto:
solution: use BTreeSet instread of Vector

### DIFF
--- a/client-rpc/src/rpc/transaction_rpc.rs
+++ b/client-rpc/src/rpc/transaction_rpc.rs
@@ -9,6 +9,7 @@ use chain_core::tx::data::output::TxOut;
 use chain_core::tx::data::{Tx, TxId};
 use chain_core::tx::TransactionId;
 use client_common::PublicKey;
+use std::collections::BTreeSet;
 
 #[derive(Debug, Serialize, Deserialize)]
 pub struct RawTransaction {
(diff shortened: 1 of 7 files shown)


chain-abci/tests/abci_app.rs
```diff
@@ -49,6 +49,7 @@ use chain_tx_filter::BlockFilter;
 use hex::decode;
 use kvdb::KeyValueDB;
 use kvdb_memorydb::create;
+use parity_scale_codec::alloc::collections::BTreeSet;
```
devashishdxt, line 52: Is there such a type? Why not just use `std::collections::BTreeSet`?

linfeng-crypto: Oh, the IDE did that automatically for me, it doesn't matter, because BTreeSet does not match the requirement.

client-rpc/src/rpc/transaction_rpc.rs
```diff
@@ -9,6 +9,7 @@ use chain_core::tx::data::output::TxOut;
 use chain_core::tx::data::{Tx, TxId};
 use chain_core::tx::TransactionId;
 use client_common::PublicKey;
+use parity_scale_codec::alloc::collections::BTreeSet;
```
tomtau, line 12:
```suggestion
use std::collections::BTreeSet;
```

PR #1451, opened 2020-04-20
Problem (Fix #1436): Performance of merkle library is not clear

### LINKED ISSUE #1436: Problem: Performance of merkle library is not clear
Benchmark metrics:
- tps
- number of internal nodes
- disk usage

yihuang:
#### Internal node fill rate

10000 stakings per version, commit 1000 versions in memory.

Internal node fill rate: 0.22 - 0.25, when the number of stakings is under 10M.

```
$ cargo run --example jellyfish_nodes -p chain-storage 1000 10000
...
internal/leaf: 65557 / 170000 = 0.3856294117647059
internal avg fill rate: 0.22457174672422472
...
internal/leaf: 807343 / 2060000 = 0.39191407766990294
internal avg fill rate: 0.22197365308177566
...
internal/leaf: 1090517 / 2840000 = 0.3839848591549296
internal avg fill rate: 0.2252667771341483
...
internal/leaf: 3333197 / 10000000 = 0.3333197
internal avg fill rate: 0.2500076503128978
```

#### In memory operations benchmark

```
$ cargo bench -p chain-storage
jellyfish, insert 256   time:   [4.6825 ms 4.6957 ms 4.7107 ms]
jellyfish, insert       time:   [30.993 us 31.144 us 31.313 us]
jellyfish, get_with_proof
                        time:   [15.605 us 15.721 us 15.877 us]
```

#### Disk usage

10000 stakings per version, no pruning.

```
$ cargo run --example disk_storage -p chain-storage /tmp/db 10 10000
$ du -sh /tmp/db
 37M	/tmp/db
$ rm -r /tmp/db
$ cargo run --example disk_storage -p chain-storage /tmp/db 100 10000
$ du -sh /tmp/db
 625M	/tmp/db
```

#### Disk get benchmark

Get staking from db with existing address:

```
$ DBPATH=/tmp/db STAKING_ADDRESS=0000000000000000000000000000000000000000  cargo run --example bench_jellyfish_disk -p chain-storage -- --bench
jellyfish_disk, get_with_proof
                        time:   [505.86 us 511.42 us 517.79 us]
```

Not exist address:
```
$ DBPATH=/tmp/db STAKING_ADDRESS=ffffffffffffffffffffffffffffffffffffffff cargo run --example bench_jellyfish_disk -p chain-storage -- --bench
jellyfish_disk, get_with_proof
                        time:   [370.42 us 375.35 us 383.60 us]
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -382,7 +382,10 @@ version = "0.2.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2889e6d50f394968c8bf4240dc3f2a7eb4680844d27308f798229ac9d4725f41"
 dependencies = [
+ "lazy_static",
  "memchr",
+ "regex-automata",
+ "serde",
 ]
 
 [[package]]
@@ -425,6 +428,15 @@ name = "bytes"
 version = "0.5.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "130aac562c0dd69c56b3b1cc8ffd2e17be31d0b6c25b61c96b76231aa23e39e1"
+
+[[package]]
+name = "cast"
+version = "0.2.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4b9434b9a5aa1450faa3f9cb14ea0e8c53bb5d2b3c1bfd1ab4fc03e9f33fbfb0"
+dependencies = [
+ "rustc_version",
+]
 
 [[package]]
 name = "cbindgen"
@@ -538,6 +550,7 @@ dependencies = [
  "bit-vec 0.6.1",
  "blake3",
  "chain-core",
+ "criterion",
  "integer-encoding",
  "jellyfish-merkle",
  "kvdb",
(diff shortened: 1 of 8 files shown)


yihuang:
> looks ok; for the disk get benchmark, it'll be more interesting to try with different addresses (not only zero-address)

Done

leejw51crypto:
speed is not so fast , can i get speed with other ?

leejw51crypto:
additional benchmark with starling is necessary

leejw51crypto:
i think 
1 insert roughly 0.085 us

tomtau:
> > looks ok; for the disk get benchmark, it'll be more interesting to try with different addresses (not only zero-address)
> 
> Done

I rather meant overall comparisons that it'll be looking up several different addresses in that microbenchmark -- to double check that the time for one particular existing address isn't some sort of the best edge case.

It's ok to merge now -- some further tests will be required later.

> speed is not so fast
> 
> also, In benchmark
> add os, cpu, ram
> 
> i also need benchmark with other merkletrie to compare

@leejw51crypto it seems ok for the account / staking storage, as there updates aren't that frequent  and for querying, it seems fine and one can load-balance. 

For UTXO commitments, it'll need more benchmarking and examination to find out whether the same trie could be used for that purpose, or something different would be preferable for lower overhead.

tomtau:
(for UTXO commitment, one should experiment with a similar UTXO set size to Bitcoin)

yihuang:
> (for UTXO commitment, one should experiment with a similar UTXO set size to Bitcoin)

Bitcoin UTxO size: 70M.
Jellyfish don't support deletion, I double they will support this in the future, so we might need to do this ourselves if we want to support UTxO.

yihuang:
> > > looks ok; for the disk get benchmark, it'll be more interesting to try with different addresses (not only zero-address)
> > 
> > 
> > Done
> 
> I rather meant overall comparisons that it'll be looking up several different addresses in that microbenchmark -- to double check that the time for one particular existing address isn't some sort of the best edge case.
> 
> It's ok to merge now -- some further tests will be required later.

I checked with different addresses manually, the result doesn't change much, because the key is a further hashing of the address, so it will get randomized.

Issue #1464, opened 2020-04-22
Problem: Some abci_query returns binary format which is hard to parse


yihuang:
Currently, some paths in abci_query("account", "state") returns binary format.
I propose to change them into JSON, so a client can use them directly, rather than go through `client-rpc` or `client-cli`.

yihuang:
An alternative is to use https://github.com/polkascan/py-scale-codec to decode, so chain-abci only need to concern about scale-codec encoding.

tomtau:
one of reasons for choosing scale codec was that it has implementations in several languages:
https://github.com/polkadot-js/api/tree/master/packages/types/src/codec
https://github.com/ChainSafe/gossamer/tree/development/lib/scale
...

yihuang:
> one of reasons for choosing scale codec was that it has implementations in several languages:
> https://github.com/polkadot-js/api/tree/master/packages/types/src/codec
> https://github.com/ChainSafe/gossamer/tree/development/lib/scale
> ...

Ok, I'll close this issue, we keep chain-abci sticking to SCALE codec, the only use JSON at client side.

PR #1466, opened 2020-04-22
Problem (WIP #1313): light client doesn't verify the fetched staking state

### LINKED ISSUE #1313: Problem: light client doesn't verify the fetched staking state
1. abci query when "proof=true" should return the inclusion proof for the queried state: https://docs.tendermint.com/master/spec/abci/abci.html#query validators don't need to store this, but full nodes for (historical) querying may need to extend the storage of app_hash=>height, app_hash=>(height, trie root) unless it's already done
2. light client should keep track of the state trie root (and app hash?) from sync, so that it can verify the inclusion proof

yihuang:
Solution:
- Support query merkle inclusion proof
- Also support query historical staking state
- Query json encoded staking from abci_query directly #1464

This feature(query historical staked state) is also needed to make the reward integration tests more stable.
Added a new abci_query path "staking" which is like "account", but it returns JSON directly, also support the new features (historical query, proof), the next step will be update the "staking_state" API in client-rpc and client-cli to use the new one.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@
 *Unreleased*
 ## v0.5.0
 ### Breaking changes
+
+- *chain-storage* [1466](https://github.com/crypto-com/chain/pull/1466): add a colume to store historical staking versions
+
 ### Features
 ### Improvements
 ### Bug Fixes
--- a/chain-storage/src/api.rs
+++ b/chain-storage/src/api.rs
@@ -1,8 +1,9 @@
 use std::collections::BTreeMap;
 
 use bit_vec::BitVec;
-use parity_scale_codec::Encode;
+use parity_scale_codec::{Decode, Encode};
 
+use crate::jellyfish::Version;
 use chain_core::common::H256;
 use chain_core::state::tendermint::BlockHeight;
 use chain_core::tx::data::{
@@ -13,7 +14,7 @@ use chain_core::tx::data::{
 use super::buffer::{GetKV, StoreKV};
 use super::{
     LookupItem, StoredChainState, CHAIN_ID_KEY, COL_APP_HASHS, COL_APP_STATES, COL_EXTRA,
-    COL_NODE_INFO, GENESIS_APP_HASH_KEY, LAST_STATE_KEY,
+    COL_NODE_INFO, COL_STAKING_VERSIONS, GENESIS_APP_HASH_KEY, LAST_STATE_KEY,
 };
 
 pub fn get_last_app_state(db: &impl GetKV) -> Option<Vec<u8>> {
@@ -65,6 +66,11 @@ pub fn get_historical_app_hash(db: &impl GetKV, height: BlockHeight) -> Option<H
     Some(stored_ah)
 }
 
+pub fn get_historical_staking_version(db: &impl GetKV, height: BlockHeight) -> Option<Version> {
+    let sah = db.get(&(COL_STAKING_VERSIONS, height.encode()))?;
+    Version::decode(&mut sah.as_slice()).ok()
+}
+
 pub fn store_chain_state<T: StoredChainState>(
     db: &mut impl StoreKV,
     genesis_state: &T,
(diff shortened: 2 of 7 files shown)


CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ Note that releases after 0.4.0 will aim to preserve the transaction wire format.
 * *chain-abci* [1292](https://github.com/crypto-com/chain/pull/1292): implements new punishment and staking state transition specification and new nonce logic
 * *chain-storage* [1424](https://github.com/crypto-com/chain/pull/1424): add two columnes to the database
 * *chain-storage* [1428](https://github.com/crypto-com/chain/pull/1428): change merkle tree of staking states to jellyfish-merkle
+* *chain-storage* [1466](https://github.com/crypto-com/chain/pull/1466): add a colume to store historical staking versions
```
tomtau, line 29: not a part of 0.4.0 -- 0.5.0?

tomtau: but this PR isn't breaking -- 0.4.1? As "account" path is preserved and this JSON-encoded version is on a new path?

tomtau: oh, it's breaking because of the new column

yihuang:
> is the new column really required / does this need a breaking change? this is only for querying and if one provides the tx-query flag, it should be storing historical app states and each app state contains the staking version?

The current historical app states only stores `ChainState`, staking version is not in that, so an alternative solution is to put staking version into the `ChainState`, or store historical `ChainNodeState`, I'm not sure.

tomtau:
> > is the new column really required / does this need a breaking change? this is only for querying and if one provides the tx-query flag, it should be storing historical app states and each app state contains the staking version?
> 
> The current historical app states only stores `ChainState`, `staking_version` is not in that, so an alternative solution is to put `staking_version` into the `ChainState`, or store historical `ChainNodeState`, I'm not sure.

all of the options are breaking. a new column is probably the simplest. 

> > is the new column really required / does this need a breaking change? this is only for querying and if one provides the tx-query flag, it should be storing historical app states and each app state contains the staking version?
> 
> The current historical app states only stores `ChainState`, `staking_version` is not in that, so an alternative solution is to put `staking_version` into the `ChainState`, or store historical `ChainNodeState`, I'm not sure.

all the options are breaking -- a new column is probably the simplest.
I guess it's not possible to retroactively create this column, so the easiest migration will be to sync from scratch

yihuang:
> all the options are breaking -- a new column is probably the simplest.
> I guess it's not possible to retroactively create this column, so the easiest migration will be to sync from scratch?

Yes, the information was not there.

CHANGELOG.md
yihuang: Moved to 0.5.0.

chain-storage/src/api.rs
```diff
@@ -80,6 +86,10 @@ pub fn store_chain_state<T: StoredChainState>(
         (COL_APP_HASHS, encoded_height.clone()),
         genesis_state.get_last_app_hash().to_vec(),
     );
+    db.set(
+        (COL_STAKING_VERSIONS, encoded_height.clone()),
+        genesis_state.get_staking_version().encode(),
+    );
```
tomtau, line 92: should this be under `write_history_states`?

tomtau: or can it function properly without historical states? I assume the use case here would be that:
1. one gets the app hash (that one can verify its block was signed on by expected validators)
2. get components of app hash from historical app state and verify they match the app state
3. fetch state at that height and verify the proof leads to that expected root (from the components)

yihuang: Actually, if the light client doesn't trust the node, it can't just fetch state from the node, it needs to store the root hash it trusted during synchronization, then verify the proof against it.

yihuang:
> I guess the column is the simplest breaking change -- just one thing is whether this could function without historical states?

https://github.com/crypto-com/chain/pull/1466#discussion_r413606318

> Also I wouldn't mark this as "fix #1313", as client wasn't updated to verify fetched staking states in this PR?

Yes, I'll remove the "fix".

Issue #1456, opened 2020-04-21
Problem: tx-history not correctly restored after restoration


leejw51crypto:
after hdwallet restoration, tx-history is not fully recovered.
Outgoing tx is marked as Incoming.
tx is not recorded to wallet state in processing with memento pattern

solution would be , in doing memento pattern,
write "tx" right away to wallet state.
so txs which reference that tx as utxo input, will work correctly.

leejw51crypto:
this is found in processing of "address auto recovery" of bip44 
modified.
tidying up now

tomtau:
fixed in #1396

PR #1473, opened 2020-04-22
Problem: no address discovery after wallet recovery#1396

leejw51crypto:
Why i resolved this issue?
1. because in hd wallet restoration, i didn't see tx history. there is no way to do it except for manual importing, which is very hard to use.
2. this feature enables very handy managing of private info as well, because user only need to remember this hdwallet mnemonics.
3. also this is quite working well with "sync with loop", balance is fetched right away automatically.

Solution:
1. added dynamic address recovery in syncing
2. used 20 addresses sliding window to discover new addresses,
if addresses are beyond that limit, it's regarded as not owned ones
3. there was sync glitch for tx history recovery, this is fixed, too.

Difficulty in working this issue:
after working dynamic address recovery, some txs are displayed as "incoming", because in syncing, input TxOut was not written to Wallet State,
which is pre-fetched in beginning of sync, which is quite old data. which is not updated while in processing each TX. 
(this will be done as another PR)
so that TX's input UTXO isn't identified, so system consider it as "incoming" (from the staking).
obvisously it's not, it's "outgoing" operation to other wallet.

How to check:
1. made transfer
2. remove .storage of client-rpc
3. restore hdwallet
4. sync with loop
5. you can see txhistory
6. make new wallet, repeat above


client-cli/src/command.rs
```diff
@@ -343,6 +348,7 @@ impl Command {
                 let tendermint_client = WebsocketRpcClient::new(&tendermint_url())?;
                 let tx_obfuscation = get_tx_query(tendermint_client.clone())?;
                 let enckey = ask_seckey(None)?;
+                let _wallet_client = get_wallet_client()?;
```
yihuang, line 351: unused code?

client-core/src/wallet.rs
```diff
@@ -54,7 +54,10 @@ pub struct WalletRequest {
 }
 
 /// Interface for a generic wallet
-pub trait WalletClient: Send + Sync {
+pub trait WalletClient: Send + Sync + Clone {
+    /// recover address beforehand
+    /// ret: true== refetch transfer addresses
```
yihuang, line 59: What does this comment mean?

client-rpc/src/rpc/sync_rpc.rs
```diff
@@ -73,31 +73,60 @@ pub trait SyncRpc: Send + Sync {
     fn sync_stop(&self, request: WalletRequest) -> Result<()>;
 }
 
-pub struct SyncRpcImpl<S, C, O>
+pub struct SyncRpcImpl<S, C, O, T>
 where
     S: Storage,
     C: Client,
     O: TransactionObfuscation,
+    T: WalletClient,
```
yihuang, line 81: `WalletClient` is a big trait, normally `WalletClient` calls the syncer, rather than reverse.

client-core/src/wallet.rs
leejw51crypto: if return value is true, refetch transfer address
i'll change

client-rpc/src/rpc/sync_rpc.rs
leejw51crypto: it's for calling wall service, i'll find a way not to to use WalletClient

client-cli/src/command.rs
leejw51crypto: ok, i'll check

client-core/src/wallet.rs
leejw51crypto: because it's modifying addresses, changed name to
recover_addresses

client-rpc/src/rpc/sync_rpc.rs
leejw51crypto: that's because of WalletClient
fn recover_addresses
it need current addresses, and also generate,
it need to use WalletClient

leejw51crypto:
# WalletSyncer calls WalletClient trait seems not good?
wallet state will be recoded, i'll do it in another pr

# address gap(20) should also be ensured in client logic (not allow transaction which violates the max gap)
this is good point, i'll make another issue for this

# address discovery logic should be optional, and only work on HD wallet.
it's only activated in hdwallet

client-core/src/wallet/syncer.rs
```diff
@@ -277,14 +294,74 @@ impl<'a, S: SecureStorage, C: Client, D: TxDecryptor, F: FnMut(ProgressReport) -
...
         let enclave_txids = blocks
             .iter()
             .flat_map(|block| block.enclave_transaction_ids.iter().copied())
             .collect::<Vec<_>>();
         let enclave_txs = self.env.decryptor.decrypt_tx(&enclave_txids)?;
 
-        let memento = handle_blocks(&self.wallet, &self.wallet_state, &blocks, &enclave_txs)
+        if crate::types::WalletKind::HD == self.wallet.wallet_kind {
+            // only hdwallet
+            self.recover_address(&blocks)?;
```
yihuang, line 361: Even with HD wallet, should we make this feature optional with a command line option?

leejw51: ok, i'll add option for client-rpc

yihuang: > ok, i'll add option for client-rpc

Does client-cli support this feature?

leejw51crypto: client-cli,too.
I'll add option

leejw51crypto: yes, i'll add option

leejw51crypto:
added option,
rebased

client-cli/src/command.rs
```diff
@@ -193,6 +198,12 @@ pub enum Command {
             help = "Disable fast forward, which is not secure when connecting to outside nodes"
         )]
         disable_fast_forward: bool,
+        #[structopt(
+            name = "disable-address-recovery",
+            long,
+            help = "Disable address recovery, which is not necessary, if addresses already exist"
```
tomtau, line 204: should there be a default value?

leejw51crypto: i'll check with mock hardware wallet

leejw51crypto: default is using "address recovery".

leejw51crypto:
i'm checking travis issue

leejw51crypto:
fix travis & rebased

leejw51crypto:
checking integration test

client-cli/src/command.rs
tomtau: But "HW wallet" will be a "HD wallet" as well -- the difference is that it doesn't expose you private information -- I assume one would want the history recovery as well... open an issue for it?

tomtau:
bors try

client-core/src/transaction_builder.rs
```diff
@@ -23,7 +23,7 @@ use chain_core::tx::data::TxId;
 
 /// Interface for wallet transaction building from output addresses and amount.
 /// This trait is also responsible for UTXO selection.
-pub trait WalletTransactionBuilder: Send + Sync {
+pub trait WalletTransactionBuilder: Send + Sync + Clone {
```
tomtau, line 26: why is clone needed?

client-cli/src/command.rs
tomtau: "HW" wallet is generally also a "HD wallet" -- it's that it doesn't expose private keys and the current code relies on having private keys.
One can derive the public keys / addresses from the parent public key: https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki#public-parent-key--public-child-key

leejw51crypto:
checking..
codecov/patch

leejw51crypto:
yes, it's enough to check with pubkey + address index,
but main purpose for wallet client is to use "new transfer address" through trait.

tomtau:
> yes, it's enough to check with pubkey + address index,
> but main purpose for wallet client is to use "new transfer address" through trait.
> 
> ```
> WalletClient-> Syncer -> SynceLogic
>                                           check whether address exist in range of 20 window
>                                           make new transfer address
> ```

this "main purpose" (? the identified problem to solve was to allow address discovery... it doesn't need to use the walletclient trait) makes the solution more complex than it needs to be:

1. it creates a circular dependency WalletClient <-> Syncer
2. WalletClient trait is big and exposes a lot of unnecessary things to Syncer
3. the reliance on having access to (locally stored) private keys will make it impossible to be used with HW wallets

It is hence not a "minimal and accurate" solution, as per the contribution guidelines, and if it is merged, it will be need to be rewritten.

If the "purpose" was to fix https://github.com/crypto-com/chain/issues/1456 , it'll be better to fix that issue in a separate PR.

Closing this PR for now, feel free to reopen two separate PRs.

leejw51crypto:
adding get address without privatekey,
that will be another pr

Issue #1235, opened 2020-03-12
Problem: ported tx-query not tested


tomtau:
integration tests need to launch IAS proxy + new tx-query

1. multiple tx-query in multi-node -- needs port configuration workaround for EDP
2. extra flag for linker for building in Docker, but that needs to be reset back or tx-query excluded from unit tests
3. some decoding issues -- potentially different crate version

PR #1494, opened 2020-04-27
Problem: tx-query not yet ported to EDP

### LINKED ISSUE #1301: Problem: tx-query not yet ported to EDP

devashishdxt:
Solution:

- Implemented `tx-query` code using EDP in `tx-query-next`
- Added `"edp"` feature flag in `chain-core` (compiles `secp256k1` with
`"edp"` flag)
- Added `"edp-obfuscation"` and `"sgx-obfuscation"` feature flags in
client-core. `"edp-obfuscation"` compiles client with remote attestation
using EDP for connecting to `tx-query`. `"sgx-obfuscation"` compiles
client with remote attestation using SGX SDK for connecting to `tx-query`.
- `tx-query-next` internally uses a thread from a thread pool for each connection.
- Modified `ZmqHelper` in `app-runner`

Fixes #1301.

### DIFF
--- a/chain-tx-enclave/tx-query-next/app-runner/src/zmq_helper.rs
+++ b/chain-tx-enclave/tx-query-next/app-runner/src/zmq_helper.rs
@@ -1,0 +1,204 @@
+use std::{
+    convert::TryInto,
+    io::{self, Read, Write},
+};
+
+use tokio::{
+    io::{AsyncRead, AsyncWrite},
+    prelude::Async,
+    sync::lock::Lock,
+};
+use zmq::{Context, Message, Result, Socket, REQ};
+
+use enclave_protocol::FLAGS;
+
+macro_rules! lock {
+    ($e:expr) => {
+        match $e.poll_lock() {
+            Async::Ready(inner) => inner,
+            Async::NotReady => return Ok(Async::NotReady),
+        }
+    };
+}
+
+macro_rules! try_lock {
+    ($e:expr) => {
+        match $e.poll_lock() {
+            Async::Ready(inner) => inner,
+            Async::NotReady => return Err(io::ErrorKind::WouldBlock.into()),
+        }
+    };
+}
+
+macro_rules! try_poll {
+    ($e:expr) => {
+        match $e {
+            Async::Ready(inner) => inner,
+            Async::NotReady => return Err(io::ErrorKind::WouldBlock.into()),
+        }
+    };
+}
+
+/// Temporary bridging between the old code and EDP. In the long term, it may go away completely.
+///
+/// 1. Decryption requests may include the sealed payload in them (so no need to request them via zmq)
+/// 2. Encryption requests may be done via a direct attested secure channel (instead of this passing of sealed requests)
+pub struct ZmqHelper {
+    inner: Lock<Inner>,
+}
+
+/// Inner struct to hold values in a `Lock`
+///
+/// `Lock` is there to enforce the synchronous order as zmq isn't thread safe
+struct Inner {
+    /// State of ZeroMQ socket
+    state: SocketState,
+    /// ZeroMQ socket
+    socket: Socket,
+}
+
+/// State of ZeroMQ socket
+#[derive(Debug)]
+enum SocketState {
+    /// Denotes that a new request can be sent on ZeroMQ socket
+    Ready,
+    /// Denotes that a request is sent to ZeroMQ socket but response is not received yet
+    RequestSent,
+    /// Denotes that response to a previously sent request is received
+    ResponseReceived(Message),
+}
+
+impl ZmqHelper {
+    /// Creates a new instance of ZeroMQ helper
+    pub fn new(connection_str: &str) -> Self {
+        let ctx = Context::new();
+        let socket = ctx.socket(REQ).expect("failed to init zmq context");
+        socket
+            .connect(connection_str)
+            .expect("failed to connect to the tx validation enclave zmq");
+
+        let inner = Inner {
+            state: SocketState::Ready,
+            socket,
+        };
+
+        log::debug!("Successfully created ZeroMQ helper");
+
+        Self {
+            inner: Lock::new(inner),
+        }
+    }
+
+    /// Sends a message to ZeroMQ socket
+    pub fn send<S: Into<Message>>(&mut self, message: S) -> Result<Async<()>> {
+        log::debug!("Sending message to ZeroMQ");
+        let mut inner = lock!(self.inner);
+
+        match inner.state {
+            SocketState::Ready => {
+                inner.socket.send(message, FLAGS)?;
+                inner.state = SocketState::RequestSent;
+                Ok(Async::Ready(()))
+            }
+            _ => {
+                log::debug!("Unable to send message. Previous response isn't processed");
+                Ok(Async::NotReady)
+            }
+        }
+    }
+
+    /// Returns `true` if previous request was sent. `false` otherwise.
+    pub fn is_request_sent(&mut self) -> Result<Async<bool>> {
+        let inner = lock!(self.inner);
+        Ok(Async::Ready(matches!(
+            inner.state,
+            SocketState::RequestSent
+        )))
+    }
+
+    /// Returns `true` if response was received. `false` otherwise.
+    pub fn is_response_received(&mut self) -> Result<Async<bool>> {
+        let inner = lock!(self.inner);
+        Ok(Async::Ready(matches!(
+            inner.state,
+            SocketState::ResponseReceived(_)
+        )))
+    }
+
+    /// Returns the length of message received from ZeroMQ
+    pub fn get_message_len(&mut self) -> Result<Async<usize>> {
+        let mut inner = lock!(self.inner);
+
+        let message = match inner.state {
+            SocketState::RequestSent => inner.socket.recv_msg(FLAGS)?,
+            _ => {
+                log::debug!("Unable to receive message. Previous request wasn't sent");
+                return Ok(Async::NotReady);
+            }
+        };
+
+        let message_len = message.len();
+        inner.state = SocketState::ResponseReceived(message);
+
+        Ok(Async::Ready(message_len))
+    }
+
+    /// Returns the message received from ZeroMQ
+    pub fn get_message(&mut self) -> Result<Async<Message>> {
+        log::debug!("Receiving response to ZeroMQ");
+        let mut inner = lock!(self.inner);
+
+        if matches!(inner.state, SocketState::ResponseReceived(_)) {
+            let mut socket_state = SocketState::Ready;
+            std::mem::swap(&mut socket_state, &mut inner.state);
+
+            if let SocketState::ResponseReceived(message) = socket_state {
+                Ok(Async::Ready(message))
+            } else {
+                unreachable!("Socket state cannot be anything other than `ResponseReceived`")
+            }
+        } else {
+            log::debug!("Unable to receive message. Previous request wasn't sent");
+            Ok(Async::NotReady)
+        }
+    }
+}
+
+impl Read for ZmqHelper {
+    fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
+        if try_poll!(self.is_request_sent()?) {
+            let message_len: u32 = try_poll!(self.get_message_len()?)
+                .try_into()
+                .expect("Message length exceeds `u32` bounds");
+
+            buf.copy_from_slice(&message_len.to_le_bytes());
+            Ok(core::mem::size_of::<u32>())
+        } else if try_poll!(self.is_response_received()?) {
+            let message = try_poll!(self.get_message()?);
+            buf.copy_from_slice(&message);
+            Ok(message.len())
+        } else {
+            Err(io::ErrorKind::WouldBlock.into())
+        }
+    }
+}
+
+impl Write for ZmqHelper {
+    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
+        try_poll!(self.send(buf)?);
+        Ok(buf.len())
+    }
+
+    fn flush(&mut self) -> io::Result<()> {
+        try_lock!(self.inner);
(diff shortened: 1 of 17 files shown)


devashishdxt:
> could client also be ported (e.g. under a feature flag for the moment) + integration test?

Tests will be ported in following PR for #1235. I'll add client code in this PR.

chain-tx-enclave/tx-query-next/enclave-app/src/handler/decryption_request.rs
```diff
@@ -0,0 +1,122 @@
...
+    zmq_stream: Arc<Mutex<TcpStream>>,
+) -> Result<DecryptionResponse, String> {
+    let enclave_request = EnclaveRequest::GetSealedTxData {
+        txids: decryption_request.body.txs.clone(),
+    }
+    .encode();
+
+    // TODO: check tx size
+    let max_result_len = decryption_request.body.txs.len() * 8000;
+    let mut result_buf = vec![0u8; max_result_len + 1];
```
tomtau, line 43: is this necessary? this was a temporary ugly hack in the original due to the ocall result buffer fiddling... here one could just keep reading off the stream and allocate as necessary? it may need some small change in `ZmqStreamHelper`

chain-tx-enclave/tx-query-next/enclave-app/src/handler/decryption_request.rs
```diff
@@ -0,0 +1,122 @@
...
+        Err(err) => Err(format!(
+            "Error while decoding response from ZeroMQ: {}",
+            err
+        )),
+    }
+}
+
+#[inline]
+fn is_allowed_view(allowed_views: &[TxAccessPolicy], view_key: &PublicKey) -> bool {
+    // TODO: policy != alldata + const eq?
```
tomtau, line 120: policies were reduced in the stabilised format, but constant equality will still be necessary to avoid leaking view keys

chain-tx-enclave/tx-query-next/enclave-app/src/main.rs
```diff
@@ -1,42 +1,157 @@
...
-    // currently, it's not possible to pass arguments
-    // FIXME: fix that or bind via Usercall extensions?
+    std::env::set_var("RUST_LOG", "debug");
+    env_logger::init();
+
+    log::info!("Connecting to ZeroMQ");
+    let zmq_stream = Arc::new(Mutex::new(TcpStream::connect("zmq")?));
+
+    // Currently, it's not possible to pass arguments to enclave code
+    // FIXME: Fix that or bind via Usercall extensions?
```
tomtau, line 34: the bind didn't work?

chain-tx-enclave/tx-query-next/enclave-app/src/main.rs
```diff
@@ -1,42 +1,157 @@
...
     // ref: https://github.com/fortanix/rust-sgx/issues/136
     // ref: https://github.com/fortanix/rust-sgx/blob/master/examples/usercall-extension-bind/runner/src/main.rs#L26
-    let listener = TcpListener::bind("0.0.0.0:3443")?;
-    println!("enclave: listening");
+    let addrs = "0.0.0.0:3443";
+    let num_threads = 4;
+    let config = EnclaveRaConfig {
+        sp_addr: "0.0.0.0:8989".to_string(),
```
tomtau, line 40: is that a valid address? It will connect to sp, not listen?

chain-tx-enclave/tx-query-next/enclave-app/src/main.rs
```diff
@@ -1,42 +1,157 @@
...
+                }
+                Err(err) => {
+                    log::error!("Error while decoding tx-query init request: {}", err);
+                    return;
                 }
-            }
-            Err(_) => {}
+            };
         }
+        Err(err) => log::error!("Error while reading bytes from TLS stream: {}", err),
```
tomtau, line 155: this is pretty complex and hard to read -- and shouldn't some of the errors at least be written back to the stream and/or send TLS `close_notify`?

client-core/src/cipher/default.rs
```diff
@@ -29,7 +39,23 @@ fn get_tls_config() -> Arc<rustls::ClientConfig> {
...
     client_cfg.versions.push(rustls::ProtocolVersion::TLSv1_2);
-    Arc::new(client_cfg)
+    Ok(Arc::new(client_cfg))
+}
+
+#[cfg(feature = "edp-obfuscation")]
+fn get_tls_config() -> Result<Arc<rustls::ClientConfig>> {
+    let verifier_config = EnclaveCertVerifierConfig {
+        signing_ca_cert_pem: IAS_CERT.into(),
+        valid_enclave_quote_statuses: vec!["OK".into(), "GROUP_OUT_OF_DATE".into()].into(),
```
tomtau, line 49: `GROUP_OUT_OF_DATE` isn't to be trusted

chain-tx-enclave/tx-query-next/enclave-app/src/main.rs
devashishdxt: Ideally, there should be a way to write error messages back to the stream. But, we'll have to modify `enclave-protocol` to do that as there is not way to send back error messages in current types.

chain-tx-enclave/tx-query-next/enclave-app/src/main.rs
devashishdxt: Yes. `tx-query` will connect to sp.

tomtau: if that's "0.0.0.0", what it will connect to? Or is this to be rebound in the runner?

chain-tx-enclave/tx-query-next/enclave-app/src/main.rs
tomtau: and `.send_close_notify()` ?

chain-tx-enclave/tx-query-next/enclave-app/src/main.rs
devashishdxt: This'll connect to `ra-sp-server` running on local machine.

chain-tx-enclave/tx-query-next/enclave-app/src/main.rs
devashishdxt: `.send_close_notify()` will not add any value here because the `TcpStream` gets dropped right after returning from this function (which closes the connection) and we don't write any additional message to the stream before closing connection.

client-core/src/cipher/default.rs
devashishdxt: Ok. I'll remove this. The reason I added this here is because during testing, I was not able to get `"OK"` status back. Maybe it is because I was using an old `SPID`?

chain-tx-enclave/tx-query-next/enclave-app/src/main.rs
devashishdxt: > If `connect_stream` returns `None`, the default implementation of `connect_stream` is used. [docs](https://docs.rs/enclave-runner/0.3.0/enclave_runner/usercalls/trait.UsercallExtension.html#method.connect_stream)

chain-tx-enclave/tx-query-next/enclave-app/src/handler/decryption_request.rs
devashishdxt: Ok. I'll replace it with [IoReader](https://docs.rs/parity-scale-codec/1.3.0/parity_scale_codec/struct.IoReader.html)

devashishdxt: Unfortunately, we cannot do this using `IoReader` because `rustls::Stream` does not implement `std::io::Seek` which is required by `IoReader`.

devashishdxt: > here one could just keep reading off the stream and allocate as necessary

We need to know the number of bytes we want to read from a stream. `IoReader` handled it internally but unfortunately, we cannot use that here.

tomtau: I meant in the runner / `ZmqStreamHelper` -- it could keep track of the cursor etc.?

chain-tx-enclave/tx-query-next/enclave-app/src/main.rs
tomtau: it'd be more graceful?

client-core/src/cipher/default.rs
tomtau: old Intel PSW (aesm etc.) or old CPU microcode / BIOS

chain-tx-enclave/tx-query-next/enclave-app/src/main.rs
devashishdxt: How? `.send_close_notify()` will only send something to the client when we write on the stream ([docs](https://docs.rs/rustls/0.17.0/rustls/trait.Session.html#tymethod.send_close_notify)).

devashishdxt:
> since the enclave itself is locking around zmq stream, perhaps the runner could be simplified not to use that async locking stuff?

Ok. I'll try to simplify this.

chain-tx-enclave/tx-query-next/enclave-app/src/main.rs
tomtau: + empty write?

chain-tx-enclave/tx-query-next/enclave-app/src/handler/encryption_request.rs
```diff
@@ -0,0 +1,126 @@
...
+                .write_all(&enclave_request)
+                .map_err(|err| format!("Error while writing request to ZeroMQ: {}", err))?;
+
+            let result_len = zmq_stream
+                .read(&mut result_buf)
+                .map_err(|err| format!("Error while reading response from ZeroMQ: {}", err))?;
+
+            if result_len > enclave_request.len() {
+                return Err("Not enough allocated space to return the sealed tx data".to_owned());
+            }
```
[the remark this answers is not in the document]
devashishdxt: The maximum it'll read is `result_len + 1` (size of buffer).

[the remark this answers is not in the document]
devashishdxt: > even with those one-time fixed result buffers (which don't seem necessary -- one could do better with stream reading more easily... e.g. some retry up to some max allocation), their handling doesn't seem correct

The implementation is trying to mimic code here (https://github.com/crypto-com/chain/blob/8dd02d3774c3bcbfc787a625ab1e5e97d2f01317/chain-tx-enclave/tx-query/enclave/src/lib.rs#L217)

Can you give me example of your thought? ("some retry up to some max allocation")

chain-tx-enclave/tx-query-next/enclave-app/src/handler/decryption_request.rs
```diff
@@ -0,0 +1,122 @@
...
+
+    let result_len = zmq_stream
+        .read(&mut result_buf)
+        .map_err(|err| format!("Error while reading response from ZeroMQ: {}", err))?;
+
+    if result_len > max_result_len {
+        return Err("Not enough allocated space to return the sealed tx data".to_owned());
+    }
+
+    match EnclaveResponse::decode(&mut result_buf.as_slice()) {
```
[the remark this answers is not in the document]
devashishdxt: Changed

chain-tx-enclave/tx-query-next/enclave-app/src/handler/encryption_request.rs
```diff
@@ -0,0 +1,126 @@
...
+
+            match EnclaveResponse::decode(&mut result_buf.as_slice()) {
```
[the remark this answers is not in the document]
devashishdxt: Changed

chain-tx-enclave/tx-query-next/enclave-app/src/handler/encryption_request.rs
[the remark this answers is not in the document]
tomtau: replied on slack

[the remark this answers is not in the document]
tomtau: For that `if result_len > enclave_request.len() {`

it seems like `if false {` ?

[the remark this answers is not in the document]
devashishdxt: Changed

client-core/src/cipher/default.rs
devashishdxt: Changed

chain-tx-enclave/tx-query-next/enclave-app/src/handler/decryption_request.rs
devashishdxt: Changed

chain-tx-enclave/tx-query-next/app-runner/src/zmq_helper.rs
```diff
@@ -0,0 +1,204 @@
...
+        if try_poll!(self.is_request_sent()?) {
+            let message_len: u32 = try_poll!(self.get_message_len()?)
+                .try_into()
+                .expect("Message length exceeds `u32` bounds");
+
+            buf.copy_from_slice(&message_len.to_le_bytes());
+            Ok(core::mem::size_of::<u32>())
+        } else if try_poll!(self.is_response_received()?) {
+            let message = try_poll!(self.get_message()?);
+            buf.copy_from_slice(&message);
```
tomtau, line 178: It'll be good to document this extra semantics on top of / from ZmqHelper, as one may get panics from `copy_from_slice` (e.g. if one uses `[u8; 8]` for the first buffer?)

Issue #1495, opened 2020-04-27
Problem: no way to get pubkey from extended pubkey


leejw51crypto:
add api to get pubkey from extended pubkey + index

leejw51crypto:
started

leejw51crypto:
done

PR #1496, opened 2020-04-27
Problem: no way to get pubkey from extended pubkey #1495

leejw51crypto:
Why?
there is no apis to make address from parent pubkey
to peek addresses, for non-hardened keys, it's possible from parent pubkey.
this feature is necessary for address recovery in sync (another PR)

What problem to solve?
previously, to get address of specific index, we have to derive private,publickeys, which is not necessary, especially for hardware wallet

Why i think this issue solved?
checked with unit-test


leejw51crypto:
yes, i'll change

PR #1500, opened 2020-04-27
Problem (Fix #1499): query api in tendermint client trait don't support height and prove

### LINKED ISSUE #1499: Problem: query api in tendermint client trait don't support height and prove
Separated issue from #1313.

yihuang:
Solution:
- Add the parameters, and do the nessesary changes.

### DIFF
--- a/client-common/src/tendermint/client.rs
+++ b/client-common/src/tendermint/client.rs
@@ -37,7 +37,16 @@ pub trait Client: Send + Sync + Clone {
     fn broadcast_transaction(&self, transaction: &[u8]) -> Result<BroadcastTxResponse>;
 
     /// Makes `abci_query` call to tendermint
-    fn query(&self, path: &str, data: &[u8]) -> Result<AbciQuery>;
+    ///
+    /// height: `None` means latest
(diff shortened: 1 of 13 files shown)


client-common/src/tendermint/client.rs
```diff
@@ -37,7 +37,13 @@ pub trait Client: Send + Sync + Clone {
     fn broadcast_transaction(&self, transaction: &[u8]) -> Result<BroadcastTxResponse>;
 
     /// Makes `abci_query` call to tendermint
-    fn query(&self, path: &str, data: &[u8]) -> Result<AbciQuery>;
+    fn query(
+        &self,
+        path: &str,
+        data: &[u8],
+        height: Option<Height>,
+        prove: bool,
```
leejw51crypto, line 45: some comments for prove?

leejw51crypto: can prove be optional?

yihuang: Added some comments now.
Prove is to say include proof or not, can pass `false` by default.

PR #1511, opened 2020-04-29
Problem (Fix #1313): light client doesn't verify the fetched staking state

### LINKED ISSUE #1313: Problem: light client doesn't verify the fetched staking state

yihuang:
Solution:
- Add staking_root in sync state
- Remove the "account" path in favor of the new "staking" path
- Client staking state command add wallet name parameter,
  and verify staking state and proof against the trusted staking_root in sync state
- Need to sync wallet before fetching the new staking state now. (you can still get any staking state without verification by request abci_query API directly)

### DIFF
--- a/client-network/src/network_ops/default_network_ops_client.rs
+++ b/client-network/src/network_ops/default_network_ops_client.rs
@@ -14,6 +14,7 @@ use chain_core::tx::data::input::TxoPointer;
 use chain_core::tx::data::output::TxOut;
 use chain_core::tx::fee::FeeAlgorithm;
 use chain_core::tx::{TxAux, TxPublicAux};
+use chain_storage::jellyfish::SparseMerkleProof;
 use chain_tx_validation::{check_inputs_basic, check_outputs_basic, verify_unjailed};
 use client_common::tendermint::types::AbciQueryExt;
 use client_common::tendermint::Client;
@@ -70,34 +71,6 @@ where
     /// Returns current underlying wallet client
     pub fn get_wallet_client(&self) -> &W {
         &self.wallet_client
-    }
-
-    /// Get account info
-    fn get_account(&self, staked_state_address: &[u8]) -> Result<StakedState> {
-        let bytes = self
-            .client
-            .query("account", staked_state_address, None, false)?
-            .bytes();
-
-        StakedState::decode(&mut bytes.as_slice()).chain(|| {
-            (
-                ErrorKind::DeserializationError,
-                format!(
-                    "Cannot deserialize staked state for address: {}",
-                    hex::encode(staked_state_address)
-                ),
-            )
-        })
-    }
-
-    /// Get staked state info
-    fn get_staked_state_account(
-        &self,
-        to_staked_account: &StakedStateAddress,
-    ) -> Result<StakedState> {
-        match to_staked_account {
-            StakedStateAddress::BasicRedeem(ref a) => self.get_account(&a.0),
-        }
     }
 
     /// Calculate the withdraw unbounded fee
@@ -164,11 +137,12 @@ where
         transactions: Vec<(TxoPointer, TxOut)>,
         to_address: StakedStateAddress,
         attributes: StakedStateOpAttributes,
+        verify_staking: bool,
     ) -> Result<(TxAux, TransactionPending)> {
         // if the to_address belongs to current wallet, we do not check the state
         let staking_addresses = self.wallet_client.staking_addresses(name, enckey)?;
         if !staking_addresses.contains(&to_address) {
-            let staked_state = self.get_staked_state(&to_address)?;
+            let staked_state = self.get_staked_state(name, &to_address, verify_staking)?;
             verify_unjailed(&staked_state).map_err(|e| {
                 Error::new(
                     ErrorKind::ValidationError,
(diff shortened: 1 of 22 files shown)


client-network/src/network_ops/default_network_ops_client.rs
```diff
@@ -472,8 +445,52 @@ where
...
     #[inline]
-    fn get_staked_state(&self, address: &StakedStateAddress) -> Result<StakedState> {
-        self.get_staked_state_account(address)
+    fn get_staked_state(
+        &self,
+        name: &str,
+        address: &StakedStateAddress,
+        verify: bool,
+    ) -> Result<StakedState> {
+        let mstaking = if verify {
```
leejw51crypto, line 454: instead of mstaking,
how about full word variable name ?

client-network/src/network_ops/default_network_ops_client.rs
```diff
@@ -472,8 +445,52 @@ where
...
+            let sync_state = self.wallet_client.get_sync_state(name)?;
+            let rsp = self.client.query(
```
leejw51crypto, line 456: full name with "response"

client-network/src/network_ops/default_network_ops_client.rs
```diff
@@ -472,8 +445,52 @@ where
...
+                address.as_ref(),
+                Some(sync_state.last_block_height.into()),
+                true,
+            )?;
+            let mstaking = <Option<StakedState>>::decode(&mut rsp.bytes().as_slice())
+                .err_kind(ErrorKind::DeserializationError, || {
+                    format!("Cannot deserialize staked state for address: {}", address)
+                })?;
+
+            let mut proof_bytes = rsp.proof.as_ref().unwrap().ops[0].data.as_slice();
```
leejw51crypto, line 467: expect and
assert!(ops.len()>0)?

client-network/src/network_ops/default_network_ops_client.rs
yihuang: I believe the name style of `mxxx` is adopted from Haskell, because I've seen this style in our code base several times, so I keep on using it. Basically in these cases `mxxx` means `optional_xxx`.

client-network/src/network_ops/default_network_ops_client.rs
yihuang: Personally, I think well-known shortcuts should be ok to keep the code short, actually easier to read.

client-network/src/network_ops/default_network_ops_client.rs
yihuang: Changed to return proper error for invalid response.

yihuang:
> > Tx construction in network-ops doesn't verify staking state for now, because the unit tests are hard to mock the env.
> > Need to sync wallet before fetching the new staking state now.
> 
> maybe the verification option could be added as extra argument -- in local setting, one could disable it if one knows it's querying against the user's own full node; in mocked unit tests it could also be disabled?

Added a `verify_staking` parameter, unit tests pass `false`, client-cli and client-rpc passes `true`.

tomtau:
multi-node integration test fails with
> jsonrpcclient.exceptions.ReceivedErrorResponseError: Deserialization error: Unable to deserialize global state for wallet with name target

tomtau:
also this should update the changelog breaking change, as the query path was removed

yihuang:
> multi-node integration test fails with
> 
> > jsonrpcclient.exceptions.ReceivedErrorResponseError: Deserialization error: Unable to deserialize global state for wallet with name target

This is actually a separated issue, solved in https://github.com/crypto-com/chain/pull/1524.

chain-abci/src/app/query.rs
```diff
@@ -170,26 +170,6 @@ impl<T: EnclaveProxy> ChainNodeApp<T> {
...
-                        None => {
-                            resp.log += "account lookup failed: account not exists";
-                            resp.code = 1;
-                        }
-                    }
-                } else {
-                    resp.log += "account lookup failed (either invalid address or node not correctly restored / initialized)";
-                    resp.code = 3;
-                }
-            }
```
tomtau, line 192: is there a need to do a breaking change? the original default behaviour should identical to "staking" path without any parameters? it could possibly be a non-breaking change with `"staking" | "account" => {...` pattern match... it could log some deprecation warning for "account" path that it'd be removed in the future

yihuang:
> looks all right, but perhaps no need to make this a breaking change immediately?

Do you mean keep the account path and staking path at the same time?

tomtau:
> > looks all right, but perhaps no need to make this a breaking change immediately?
> 
> Do you mean keep the account path and staking path at the same time?

yes, for the time being

yihuang:
> > > looks all right, but perhaps no need to make this a breaking change immediately?
> > 
> > 
> > Do you mean keep the account path and staking path at the same time?
> 
> yes, for the time being

Ok, done.

PR #1514, opened 2020-04-30
Problem (Fix #1038): no total supply sanity checks

### LINKED ISSUE #1038: Problem: no total supply sanity checks
Currently coins can reside in several places:
1. staked states (bonded or unbonded)
2. rewards pool (not-yet-minted coins or bonus from fees and punishments)
3. transfer transaction outputs

different state updates then move these around.
It'll be good to have some high-level statistics tracking on these 3 areas, which will be updated on each change + an invariant that sum(1.'s amount + 2.'s amount + 3.'s amount) == maximum supply.
there could be fuzzing tests that would try to break this invariant

yihuang:
Solution:
- Record utxo coins
- Do the check at the end of commit events.
- Optimize `sum_coins` to use `try_fold`
  ```
  utxo = withdraw - deposit - utxo transfer fee
  utxo + staking + reward_pool = initial distribution + minted
  minted <= expansion_cap
  initial distribution + expansion_cap = max supply
  ```

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -21,6 +21,7 @@ steps:
   commands:
   - export CARGO_HOME=$PWD/drone/cargo
   - export CARGO_TARGET_DIR=$PWD/drone/target
+  - export CRYPTO_CHAIN_ENABLE_SANITY_CHECKS=1
   - cargo test
 
 - name: integration-tests
@@ -38,6 +39,7 @@ steps:
   commands:
   - export CARGO_TARGET_DIR=$PWD/drone/target
   - export PYTHON_VENV_DIR=$PWD/drone/venv
+  - export CRYPTO_CHAIN_ENABLE_SANITY_CHECKS=1
   - LD_LIBRARY_PATH=/opt/intel/sgx-aesm-service/aesm /opt/intel/sgx-aesm-service/aesm/aesm_service
   - ./integration-tests/run.sh
 
@@ -56,6 +58,7 @@ steps:
   commands:
   - export CARGO_TARGET_DIR=$PWD/drone/target
   - export PYTHON_VENV_DIR=$PWD/drone/venv
+  - export CRYPTO_CHAIN_ENABLE_SANITY_CHECKS=1
   - LD_LIBRARY_PATH=/opt/intel/sgx-aesm-service/aesm /opt/intel/sgx-aesm-service/aesm/aesm_service
   - ./integration-tests/run_multinode.sh
 
@@ -128,6 +131,6 @@ trigger:
 
 ---
 kind: signature
-hmac: 9a1fa3521f6d925ed079f7953dd0b2d721e207895928a04119659a9f99fe7de2
+hmac: fbabf3a62adcb7697a34675624f62c408eda1197524bbab22a904694cb3c2fc8
 
 ...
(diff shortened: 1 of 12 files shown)


tomtau:
the failing tests are because of https://github.com/crypto-com/chain/pull/1516 ?

yihuang:
> the failing tests are because of #1516 ?

not sure, I might not pushed the most recent version yet. I'm working on #1516 first.

leejw51crypto:
could you add logic of computing staking sum?
some comment of explaining macro?

i want to know the performance of computing staking amounts

chain-storage/src/jellyfish.rs
```diff
@@ -230,6 +233,36 @@ fn decode_stale_node_index(data: &[u8]) -> Result<StaleNodeIndex> {
     })
 }
 
+/// Iterate through all stakings
+pub fn iter_stakings<S: GetKV>(
+    storage: &S,
+    version: Version,
+) -> impl Iterator<Item = StakedState> + '_ {
```
leejw51crypto, line 240: in production, this will take a lot of time

yihuang:
> could you add logic of computing staking sum?
> some comment of explaining macro?
> 
> i want to know the performance of computing staking amounts.
> is it iterate all merkletrie to fetch amount?

Yes, it's simply iterate all leaf nodes of merkle trie and sum the `bonded + unbonded`.

> 
> i think this is enabled in debug mode, it's ok though.

Yes, it's supposed to only run in debug mode, not in production, at least not run per block
.
> 
> on second thought,
> how about computing when staking account is updated

Then you need to make sure this update logic is bug free.

chain-storage/src/jellyfish.rs
yihuang: yes

yihuang:
> looks all right -- as for "iterating through all states", I think it may be good to make this optional under some flag (instead of for all debug build) and/or add a more lightweight check (counting the total stake as in utxo), because one may want to run a debug chain-abci build on some larger network data like the public testnet.
> that's more for the future; will merge this after 0.5 release

Yes, it might make sense to make it a runtime flag, default to false, enabled in integration tests.

yihuang:
Or use an environment variable instead of a command-line argument, so it doesn't get exposed to the normal users.

.drone.yml
```diff
@@ -56,6 +58,7 @@ steps:
   commands:
   - export CARGO_TARGET_DIR=$PWD/drone/target
   - export PYTHON_VENV_DIR=$PWD/drone/venv
+  - export CRYPTO_CHAIN_ENABLE_SANITY_CHECKS=1
```
tomtau, line 61: drone hmac won't verify / needs updating -- I can possibly sign it later if you can't

yihuang: Still get this when trying to sign:
```
$ drone sign crypto-com/chain
client error 404: {"message":"Not Found"}
```

yihuang:
> I guess this will also be good to enable in the fuzzer target: https://github.com/crypto-com/chain/blob/master/chain-abci/fuzz/fuzz_targets/abci_cycle.rs -- could possibly be done with `std::env::set_var` ?

Done

PR #1516, opened 2020-04-30
Problem (Fix #1515): unbond tx don't subtract fee from bonded

### LINKED ISSUE #1515: Problem: unbond tx don't subtract fee from bonded.
unbond tx added fee into reward pool, but don't subtract from bonded.

yihuang:
Solution:
- Fix the bug and add unit test

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,9 +15,11 @@
 
 ### Bug Fixes
 
+- *chain-abci* [1516](https://github.com/crypto-com/chain/pull/1516): Fix unbond tx didn't subtract fee from bonded
+
 *Unreleased*
 
-## V0.4.1
+## v0.4.1
 
 ### Bug Fixes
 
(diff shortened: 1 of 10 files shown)


tomtau:
bors retry

tomtau:
or needs to call
```
 rpc.staking.unbond(addr, int(state['bonded'] - fee), enckey=enckey, name='target')
```

instead of
```
 rpc.staking.unbond(addr, int(state['bonded']), enckey=enckey, name='target')
```
?

(or something like `state['bonded']-min_required+1` ?)

yihuang:
> or needs to call
> 
> ```
>  rpc.staking.unbond(addr, int(state['bonded'] - fee), enckey=enckey, name='target')
> ```
> 
> instead of
> 
> ```
>  rpc.staking.unbond(addr, int(state['bonded']), enckey=enckey, name='target')
> ```
> 
> ?
> 
> (or something like `state['bonded']-min_required+1` ?)

Yes, fixed now, `drone exec` runs successfully locally now.

Issue #874, opened 2020-01-10
Problem: not clear from name what unit a parameter is in


tomtau:
Some parameters are in seconds, some are in blocks...
it may be clearer if the name said it, e.g. by post fixing with `_sec`

lezzokafka:
Here are  some examples: 

- In seconds:  
```
    "unbonding_period"
    "jail_duration"
    "slash_wait_period"
    "distribution_period" 
```

- In blocks:
```
       "block_signing_window"
       "missed_block_threshold"
       "evidence"
```

yihuang:
Also:
```
monetary_expansion_r0: millis,
monetary_expansion_decay: macros,
```

tomtau:
Some of the parameters disappeared (e.g.     "unbonding_period", "jail_duration", "slash_wait_period" are all replaced by evidence age), some were renamed ("reward_period_seconds")
and some seem OK ("block_signing_window" / "missed_block_threshold" kind of hints it's in blocks?).

the reward formula parameters... I don't know if they can be given some meaningful unit name given their ranges: https://github.com/crypto-com/chain-docs/blob/master/docs/protocol/reward-and-punishments.md#rewards-network-parameters-for-monetary-expansion

anyway, feel free to reopen issues for any parameters that could be good to rename

PR #1521, opened 2020-05-01
Problem (Fix #1520): `missed_block_threshold <= block_signing_window`not checked in init chain

### LINKED ISSUE #1520: Problem: `missed_block_threshold  <= block_signing_window` not checked in init chain
Add a check to prevent misconfiguration.

yihuang:
Solution:
- Add validation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,9 +9,13 @@
 
 ### Features
 
-- *chain-abci* [1458](https://github.com/crypto-com/chain/pull/1458): new abci_query path "sealed", query sealed log of transaction id.
+- *chain-abci* [1458](https://github.com/crypto-com/chain/pull/1458): new abci_query path "sealed", query sealed log of
+  transaction id
 
 ### Improvements
+
+- *chain-abci* [1521](https://github.com/crypto-com/chain/pull/1521): add network parameter validation rule:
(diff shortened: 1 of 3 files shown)


tomtau:
bors retry

chain-core/src/init/params.rs
```diff
@@ -1,3 +1,4 @@
+use super::config::DistributionError;
```
tomtau, line 1: Some options:

1. return `Result<(), ()>` in validation here
2. validation + import conditioned by `#[cfg(not(feature = "mesalock_sgx"))]`
3. move this validation to config

config module isn't compiled in sgx, because it's not needed for the existing functionality + it brings extra dependencies (base64, serde, serge-json,. ...)

yihuang: I think option 3 is simply enough.

yihuang: Also, there are some out of date network parameters and namings, we can solve them in another issue.

tomtau: do you mean this: https://github.com/crypto-com/chain/issues/874 ?

yihuang: > do you mean this: #874 ?

I mean the `jailing_config` has nothing to do with jail now, it seems that's the only thing out of date.

Issue #1483, opened 2020-04-24
Problem: client-rpc process is not necessary for python integration tests


yihuang:
With the new RPC style c API, python integration tests can run by calling the c API directly, no need the client-RPC process.
- Better testing the API
- Showcase multi-linguist client implementation.

If this approach proved to be successful, maybe we don't need to maintain client-rpc at all?

tomtau:
potentially, it can also decrease the number of dependencies.

there are a few external places where client-rpc is used e.g. https://github.com/crypto-com/sample-chain-wallet (there I guess it'll make sense to replace it with the Node.JS library)
and I think @calvinaco has written a few internal bash scripts that `curl` client-rpc...
not sure if any external parties on the testnet are using client-rpc -- @lezzokafka ?
anyway, if we ever decide to deprecate it, it should be graceful.

lezzokafka:
> not sure if any external parties on the testnet are using client-rpc -- @lezzokafka ?

Not that I know of. I think most of the external parties are using client-cli to perform node-join and staking related operations @tomtau.

leejw51crypto:
like ethereum, geth  has cli + interactive+ rpc.
maybe we can consider similar approach.
combine cli+ rpc together

but rpc via http is still useful and valid using with other languages.
because it's by tcp level.

yihuang:
- Drop the jsonrpc dependency, but keep the RPC schema deriving stuff, probably cherrypick the related code from jsonrpc library.
- Write a simple HTTP server process incoming requests one by one, since the current APIs are not supposed to be called concurrently.

PR #1544, opened 2020-05-06
Problem (#1483): client-rpc library and server dependencies are mixed

yihuang:
Solution:
- Seperate them, less dependencies in client-rpc library


client-rpc/src/lib.rs
```diff
@@ -1,8 +1,21 @@
-mod program;
-/// Used by c api
+use std::fmt::Debug;
+
 pub mod rpc;
-mod server;
 
-pub fn run() {
-    crate::program::run_electron();
+pub fn to_rpc_error<E: ToString + Debug>(error: E) -> jsonrpc_core::Error {
```
devashishdxt, line 5: `Display` is more idiomatic?
```suggestion
pub fn to_rpc_error<E: Display + Debug>(error: E) -> jsonrpc_core::Error {
```

yihuang: Maybe use `Error` directly here is more proper?

devashishdxt: Yeah. But, some error types may not implement `Error`. You can try that. If it compiles fine, then it’s good.

yihuang: Then I think `ToString + Debug` is actually accurate here, `ToString` for the `to_string` method, `Debug` for the logging.

Issue #1546, opened 2020-05-06
Problem: the name `jailing_config` in network parameter is out dated


yihuang:
The content of `jailing_config` don't have anything to do with jailing now.
Propose to merge `jailing_config` and `slashing_config` into `punishment_config`.
It's a breaking change.

Issue #1168, opened 2020-03-03
`check_wallet_syncer_app_hash_on_multiple_tx` hardcoded app hash not modifiable


yihuang:
The unit test `check_wallet_syncer_app_hash_on_multiple_tx` hardcoded app hash which is difficult to modify, when state logic changed, the test will fail.

tomtau:
@yihuang it's ok for a few hardcoded tests, as sanity checks that state logic didn't accidentally change. Perhaps a more precise problem here is that those asset files are difficult to update?

yihuang:
> @yihuang it's ok for a few hardcoded tests, as sanity checks that state logic didn't accidentally change. Perhaps a more precise problem here is that those asset files are difficult to update?

Yeah, the issue title is "the hardcoded app hash not modifiable"

tomtau:
maybe in dev-utils, there could be another command "gen-test-assets" which will generate the asset files and print out expected app hashes?

yihuang:
> maybe in dev-utils, there could be another command "gen-test-assets" which will generate the asset files and print out expected app hashes?

Just confirmed that the reason for the mismatch is I re-ordered a field in `NetworkParameter`, so in this case, have `dev-utils` able to re-generate the app hash based on the snapshotted state and corresponding block results would help.
Even better, we can update them on the fly in the unit test code.

tomtau:
> Even better, we can update them on the fly in the unit test code.
that's a bit dangerous, as it'll silently skip accidental / unwanted state logic changes

PR #1547, opened 2020-05-06
Problem (Fix #1016): jail period/unbonding time is not verified against maximum evidence age

### LINKED ISSUE #1016: Problem: jail period/unbonding time is not verified against maximum evidence age
also https://github.com/crypto-com/chain/issues/856

yihuang:
Solution:
- Remove unbonding_period parameter, use max_evidence_age of consensus
  parameter instead.

### DIFF
--- a/chain-abci/tests/validator_changes.rs
+++ b/chain-abci/tests/validator_changes.rs
@@ -58,10 +58,11 @@ fn check_unbonding_without_removing_validator() {
 #[test]
 fn check_unbonding_with_removing_validator() {
     // Init Chain
-    let (env, storage) =
+    let (mut env, storage) =
         ChainEnv::new_with_customizer(Coin::max(), Coin::zero(), 2, |parameters| {
             parameters.required_council_node_stake = (Coin::max() / 10).unwrap();
         });
+    env.max_evidence_age = 61;
     let mut app = env.chain_node(storage);
     let _rsp = app.init_chain(&env.req_init_chain());
     let state = app.last_state.as_ref().unwrap();
@@ -145,13 +146,12 @@ fn check_unbonding_with_removing_validator() {
 #[test]
 fn check_rejoin() {
     // Init Chain
-    let (env, storage) = ChainEnv::new_with_customizer(
+    let (mut env, storage) = ChainEnv::new_with_customizer(
         (Coin::max() / 2).unwrap(),
         (Coin::max() / 2).unwrap(),
         2,
         |parameters| {
             // tweaking times + parameters, something more than 0.0... gets minted
-            parameters.unbonding_period = 3600 * 24 * 10000;
             parameters.rewards_config.reward_period_seconds = 365 * 24 * 3600;
             parameters.required_council_node_stake = (Coin::max() / 4).unwrap();
             parameters.rewards_config.monetary_expansion_r0 = Milli::try_new(1, 0).unwrap();
@@ -159,6 +159,7 @@ fn check_rejoin() {
             parameters.rewards_config.monetary_expansion_decay = 0;
         },
     );
+    env.max_evidence_age = 3600 * 24 * 10000;
     let mut app = env.chain_node(storage);
     let _rsp = app.init_chain(&env.req_init_chain());
     let state = app.last_state.as_ref().unwrap();
--- a/client-core/src/wallet/syncer.rs
+++ b/client-core/src/wallet/syncer.rs
@@ -728,6 +728,7 @@ mod tests {
     }
 
     #[test]
+    #[ignore]
     fn check_wallet_syncer_app_hash_on_multiple_tx() {
         #[derive(Clone)]
         struct MockTendermintClient {}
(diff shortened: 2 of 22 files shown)


client-core/src/wallet/syncer.rs
```diff
@@ -728,6 +728,7 @@ mod tests {
     }
 
     #[test]
+    #[ignore]
```
devashishdxt, line 731: Why is this ignored?

chain-abci/tests/validator_changes.rs
```diff
@@ -145,20 +146,20 @@ fn check_unbonding_with_removing_validator() {
...
             // tweaking times + parameters, something more than 0.0... gets minted
-            parameters.unbonding_period = 3600 * 24 * 10000;
             parameters.rewards_config.reward_period_seconds = 365 * 24 * 3600;
             parameters.required_council_node_stake = (Coin::max() / 4).unwrap();
             parameters.rewards_config.monetary_expansion_r0 = Milli::try_new(1, 0).unwrap();
             parameters.rewards_config.monetary_expansion_tau = 1000_0000_0000_0000_0000;
             parameters.rewards_config.monetary_expansion_decay = 0;
         },
     );
+    env.max_evidence_age = 3600 * 24 * 10000;
```
devashishdxt, line 162: `unbonding_period` was in seconds and `max_evidence_age` is in nanoseconds. I'm not sure if this is correct? Check this at other places also.

yihuang: Yes, but we only take the seconds part of evidence age duration in app. This is handled in init chain event.

client-core/src/wallet/syncer.rs
yihuang: Because it’s not easy to update the hashes in the support files of this test case :(

yihuang: I suggest ignoring this test case until we develop a convenient tool to update the support files.

tomtau: https://github.com/crypto-com/chain/issues/1168

yihuang:
> maybe a bit simpler change (non-breaking) for now will be to keep unbonding period and add validation that unbonding period == (or >= ?) max age ( in seconds instead of nanoseconds)?
> otherwise app hash changes everywhere etc.: https://travis-ci.org/github/crypto-com/chain/jobs/683687610#L1315
> breaking genesis format changes (renaming, moving or removing fields) could be done later in one go?

- I forgot to update `./ci-scripts/check-docker-app-hash.sh` which is updated now.
- I suggest ignoring the unit test(`check_wallet_syncer_app_hash_on_multiple_tx`) until we develop a convenient tool to update the supporting files.
  We can still detect breaking changes regarding app hash by `./ci-scripts/check-docker-app-hash.sh`.
- Breaking change in `CHANGELOG` is updated too.

yihuang:
> sgx sealing test is also affected: https://github.com/crypto-com/chain/blob/master/chain-abci/src/enclave_bridge/real/test/seal.rs#L163

Done.

> 
> maybe bump the versions to 0.6.0?

We can decide the version number at the release time, #1546 can also be done in the same release.

chain-abci/tests/validator_changes.rs
yihuang: https://github.com/crypto-com/chain/blob/master/chain-abci/src/app/app_init.rs#L400

yihuang:
Fix clippy warnings, use `ChainInfo` to reduce the number of function parameters.

yihuang:
> there's one clippy warning: https://travis-ci.org/github/crypto-com/chain/jobs/683712926#L1591
> it can potentially be ignored, but I guess this could be a good suggestion for small refactoring to group some of the parameters into new types.
> 
> e.g. these parameters
> 
> ```
>         block_time: Timespec,
>         block_height: BlockHeight,
> ```
> 
> seem to always come together.
> so there could potentially be something like:
> 
> ```
> struct BlockInfo {
>         block_time: Timespec,
>         block_height: BlockHeight,
> }
> ```
> 
> or `type BlockInfo = (Timespec, BlockHeight)`
> 
> or group even more parameters if they are always passed together.

Currently, I re-used `ChainInfo` to do this.

tomtau:
> 
> Currently, I re-used `ChainInfo` to do this.
> In this case, it has two un-used fields out of five fields in total, and we don't need to create a new type.

I suggest not to reuse `ChainInfo`, as it is meant for tx validation (fee field is meant to be computed based on the tx size) -- otherwise, it may be error-prone, as there may be two ChainInfos in scope ("normal" one and the "fake" one with a zero fee).
Either a dedicated type for staking table usage, or extracting out needed part from ChainInfo into a new type (and nesting it in ChainInfo)

tomtau:
Or if this refactoring that was meant to be small turns out to be big, perhaps just revert back to that original version and place allow warning annotations on those two methods with a todo comment to refactor this later

yihuang:
Maybe just a BeginBlockInfo dedicated for begin block event for now, generalize only when there’s real need in the future.

PR #1584, opened 2020-05-11
Problem: (fix #1437) No multisignature address generation in client-cli

### LINKED ISSUE #1437: Problem: No multisignature address generation in client-cli
**The issue:**
Currently,  MultiSig operations in client-cli only support creating the public key and listing them

> - list-address-public-keys     _List public keys for MultiSig address_
> - new-address-public-key     _Creates a new public key for MultiSig address_

**Expected outcome:**
 `multiSig_createAddress` is supported already in the rpc, just have to bring it to the cli as well, the flow can be something like:
- Choose the value for n and m; 
- Insert those public keys; 
- Tada - here’s your multisig address.

linfeng-crypto:
* add command `new-address` in `client-cli multisig`
* classify the transfer addresses in client-cli print out, it looks like:

```
Address: dcro18cae94afx0qxjm3cvayyavxu9vnvuqrlexxhgqyt3vyumfm2spnq0f5kcf
MultiSig Address: dcro1w489pglfzv9lam9r6e656tmdu28sxhtgk6psfcw7rghhhfcx2qkq7qk9mr(3/4)
```

### DIFF
--- a/client-cli/src/command/address_command.rs
+++ b/client-cli/src/command/address_command.rs
@@ -1,6 +1,6 @@
 use std::str::FromStr;
 
-use quest::{ask, success, text};
+use quest::{ask, error, success, text};
 use structopt::StructOpt;
 use unicase::eq_ascii;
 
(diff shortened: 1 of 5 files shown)


linfeng-crypto:
> Looks good. One future improvement can be to support a non-interactive mode where we take all the details (public keys, number of required signatures) in command line arguments.

Ok

linfeng-crypto:
> Tried locally, works fine.
> 
> Just one suggestion, when listing the address in client-cli, should we add an extra category for multisig address? ( currently we have only _staking_ and _transfer_ )
> 
> This is because when a multisig address is being generated, it will be added to the wallet under the list of "Transfer" address. At the end, someone could accidentally use one of these addresses to receive funds without noticing it is, in fact, a multi-sig address, which "locks" the funds.

sure, I'll add.

PR #1594, opened 2020-05-13
Problem: (fix #1545)light client doesn't verify genesis hash

### LINKED ISSUE #1545: Problem: light client doesn't verify genesis hash
Pass a genesis hash from command line and verify the genesis response against it.

linfeng-crypto:
check the genesis app hash from tendermint with the trusted one when `get_genesis_sync_state`

### DIFF
--- a/client-core/src/wallet/syncer.rs
+++ b/client-core/src/wallet/syncer.rs
@@ -15,7 +15,7 @@ use chain_core::tx::TransactionId;
 use chain_storage::jellyfish::compute_staking_root;
 use chain_tx_filter::BlockFilter;
 use client_common::tendermint::types::{
-    Block, BlockExt, BlockResults, BlockResultsResponse, StatusResponse, Time,
+    Block, BlockExt, BlockResults, BlockResultsResponse, Genesis, StatusResponse, Time,
 };
 use client_common::tendermint::Client;
 use client_common::{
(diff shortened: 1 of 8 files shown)


client-core/src/wallet/syncer.rs
```diff
@@ -547,9 +547,31 @@ impl<
     }
 }
 
+/// testnet v0.5
+const GENESIS_APP_HASH: &str = "F62DDB49D7EB8ED0883C735A0FB7DE7F2A3FA322FCD2AA832F452A62B38607D5";
+
+fn check_genesis_hash(genesis_hash: &[u8]) -> Result<()> {
+    let hash_setted = std::env::var("GENESIS_APP_HASH").unwrap_or_else(|_| GENESIS_APP_HASH.into());
```
tomtau, line 554: it would be good to list this extra environment variable in the command line help, as it lists all the other client environment variables

client-core/src/wallet/syncer.rs
```diff
@@ -547,9 +547,31 @@ impl<
...
+            genesis_hash_str, hash_setted
+        );
+        Err(Error::new(ErrorKind::VerifyError, msg))
+    }
+}
+
 pub fn get_genesis_sync_state<C: Client>(client: &C) -> Result<SyncState> {
-    // FIXME verify against trusted genesis hash
     let genesis = client.genesis()?;
+    check_genesis_hash(&genesis.app_hash)?;
```
yihuang, line 574: It's not enough to check app hash of genesis, but the hash of whole genesis.

linfeng-crypto: tendermint gives the genesis, if the app_hash is wrong in genesis, tendermint can not start up.

client-core/src/wallet/syncer.rs
leejw51crypto: yes, that would be good feature

client-core/src/wallet/syncer.rs
yihuang: > tendermint gives the genesis, if the app_hash is wrong in genesis, tendermint can not start up.

We can't trust the whole genesis response from the node, and app hash only covers the app state, don't cover other important informations, like genesis_time, consensus_params, validators.

devashishdxt: I think `app_hash` does cover genesis time and consensus parameters. I may be wrong.

yihuang: > I think `app_hash` does cover genesis time and consensus parameters. I may be wrong.

No, it only covers `ChainState` plus valid tx ids in the last block.

linfeng-crypto: yes, I'll the whole genesis app hash instead of app hash

devashishdxt: https://github.com/crypto-com/chain/blob/ae5271ca1ebc9bf8fd24bd340cf828a9a2e76ed1/chain-abci/src/app/app_init.rs#L224

linfeng-crypto:
Some unit test failed, because the genesis hash is dynamic, I'm debugging

integration-tests/bot/chainbot.py
```diff
@@ -333,7 +343,15 @@ async def fix_genesis(genesis, cfg):
                 f'--genesis_dev_config_path "{fp_cfg.name}" '
                 f'--tendermint_genesis_path "{fp_genesis.name}"'
             )
-        return json.load(open(fp_genesis.name))
+            hash = await run(
```
yihuang, line 346:
```suggestion
            hash = await communicate(
```

PR #1600, opened 2020-05-14
Problem:(cro-1589) incorrect error message of client-rpc , when tx-qu…

leejw51crypto:
Solution: show sync error when sync fails
also keep trying to sync until tx-query is ready again

### DIFF
--- a/client-rpc/src/rpc/sync_rpc.rs
+++ b/client-rpc/src/rpc/sync_rpc.rs
@@ -16,6 +16,7 @@ use std::sync::Mutex;
 use std::thread;
 // seconds
 const NOTIFICATION_TIME: u64 = 2;
+const ERROR_NOTIFICATION_TIME: u64 = 30;
 pub trait CBindingCallback: Send + Sync {
     fn progress(&mut self, current: u64, start: u64, end: u64) -> i32;
     fn set_user(&mut self, user: u64);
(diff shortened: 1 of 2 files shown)


client-rpc/src/rpc/sync_rpc.rs
```diff
@@ -61,7 +62,7 @@ impl Default for SyncRequest {
     }
 }
 
-#[rpc(server)]
+#[rpc]
```
tomtau, line 65: is this annotation change necessary?

client-rpc/src/rpc/sync_rpc.rs
```diff
@@ -1,6 +1,6 @@
 use super::sync_worker::SyncWorker;
 use super::sync_worker::WorkerShared;
-use crate::to_rpc_error;
+use crate::server::to_rpc_error;
```
yihuang, line 3:
```suggestion
use crate::to_rpc_error;
```

client-rpc/src/rpc/sync_worker.rs
```diff
@@ -1,5 +1,5 @@
 use super::sync_rpc::{CBindingCallback, RunSyncProgressResult};
-use crate::rpc_error_from_string;
+use crate::server::rpc_error_from_string;
```
yihuang, line 2:
```suggestion
use crate::rpc_error_from_string;
```

client-rpc/src/rpc/sync_rpc.rs
yihuang, line 65:
```suggestion
#[rpc(server)]
```

leejw51crypto:
i'll modify

client-rpc/src/rpc/sync_rpc.rs
leejw51crypto: this is not necessary, i'll change back

leejw51crypto:
sure, i'll do it

tomtau:
bors try

Issue #1549, opened 2020-05-06
Problem: no enclave-to-enclave mutually attested TLS transmission of tx+key payloads


tomtau:
https://github.com/crypto-com/chain-docs/blob/master/plan.md#datakey-retrieval-enclave-to-enclave-communication
so one will need to write protocol to choose whether to fetch key or txids.
for server-client certs, IAS etc., it can probably reuse or adapt some of the code written for new tx query?

what txids to fetch or what enclave IP to connect is out of the scope of this issue -- they could be provided as command line arguments (or have some other way of inputting them)

PR #1612, opened 2020-05-19
Problem: No way to create mutually attested enclave-to-enclave TLS connections

devashishdxt:
Solution: Added constructs for creating attested enclave-to-enclave TLS connections

- `EnclaveCertVerifier` now implements `ClientCertVerifier`
- Added some utility functions for creating `rustls::ServerConfig` and `rustls::ClientConfig`

Related to #1549.


devashishdxt:
> wouldn't one also need to set certificate+key in `ClientConfig`?

Added a new utility function in `Cerificate` to `configure_client_config`.

PR #1618, opened 2020-05-19
Problem: edp related crates conflicts with hacks for old enclave code

yihuang:
Solution:
- Put edp related crates in a standalone directory


yihuang:
> + changelog in ADR?

Done

PR #1619, opened 2020-05-19
Problem (#1483): client rpc setup code is duplicated between server and c-api

yihuang:
Solution:
- consolidate the client-rpc setup code
- cro-clib add mock mode api when build in debug
- build cro-clib as cdylib, can be used directly by other language.
- make callback parameter in c api optional

### DIFF
--- a/cro-clib/Cargo.toml
+++ b/cro-clib/Cargo.toml
@@ -6,7 +6,7 @@ edition = "2018"
 
 [lib]
 name = "cro_clib"
-crate-type =["staticlib"]
+crate-type = ["staticlib"]
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
--- a/cro-clib/chain.h
+++ b/cro-clib/chain.h
@@ -19,6 +19,8 @@ typedef struct CroJsonRpc CroJsonRpc;
 
 typedef struct CroTx CroTx;
 
+typedef struct Option_ProgressCallback Option_ProgressCallback;
+
 typedef struct CroResult {
   int result;
 } CroResult;
@@ -30,12 +32,6 @@ typedef CroFee *CroFeePtr;
 typedef CroHDWallet *CroHDWalletPtr;
 
 typedef CroJsonRpc *CroJsonRpcPtr;
-
-/**
- * current, start, end, userdata
- * return: 1: continue, 0: stop
- */
-typedef int32_t (*ProgressCallback)(uint64_t, uint64_t, uint64_t, const void*);
 
 typedef CroTx *CroTxPtr;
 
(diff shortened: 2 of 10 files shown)


cro-clib/Cargo.toml
```diff
@@ -6,7 +6,7 @@ edition = "2018"
 
 [lib]
 name = "cro_clib"
-crate-type =["staticlib"]
+crate-type =["staticlib", "cdylib"]
```
leejw51crypto, line 9: dynamic library is intentionally removed.
because of ssl related error

cro-clib/chain.h
```diff
@@ -19,6 +19,8 @@ typedef struct CroJsonRpc CroJsonRpc;
 
 typedef struct CroTx CroTx;
 
+typedef struct Option_ProgressCallback Option_ProgressCallback;
```
leejw51crypto, line 22: where is Option_ProgressCallback?
opaque pointer?

leejw51crypto: how about "OptionProgressCallback"?
removing _

yihuang: This is auto-generated by cbindgen from `Option<ProgressCallback>`.

tomtau: I guess the example needs updating: https://github.com/crypto-com/chain/blob/master/cro-clib/examples/rpc.c#L12 ?

tomtau: or if one passes non-null function pointer, it'll work as intended / before?

cro-clib/Cargo.toml
tomtau: @yihuang is this necessary? if it is, these instructions will need update: https://github.com/crypto-com/chain/blob/master/cro-clib/android.md as one may need to remove `cdylib` manually

cro-clib/chain.h
yihuang: > or if one passes non-null function pointer, it'll work as intended / before?

yes, the null pointer will be treated as a `None`.

cro-clib/Cargo.toml
yihuang: How about moving the c wrapper for jsonrpc API into the `client-rpc-core` crate, produce `cdylib` there. `cro-clib` produce staticlib and cross-compiling as before.

tomtau: I think that works -- probably https://github.com/crypto-com/chain/blob/master/cro-clib/build.rs will need updating?

yihuang: Removed the `cdylib`, change the `crate-type` in `build.sh` with tool on the fly to build the shared library.

PR #1623, opened 2020-05-19
Problem (Fix #1483): client-rpc process is not necessary for python integration tests

### LINKED ISSUE #1483: Problem: client-rpc process is not necessary for python integration tests

yihuang:
Solution:
- python call the rpc shared library directly
- chainbot don't start client-rpc by default,
  but integration tests still starts client-rpc for type script tests.

fast forward is disabled by default through c API, because it's problematic in zero-fee test env.
Opened an issue to configure it through c API: #1627

### DIFF
--- a/cro-clib/src/jsonrpc.rs
+++ b/cro-clib/src/jsonrpc.rs
@@ -275,7 +275,7 @@ unsafe fn create_rpc(
         })),
     });
     let options = SyncerOptions {
-        enable_fast_forward: true,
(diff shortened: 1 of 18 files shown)


tomtau:
bors try

PR #1625, opened 2020-05-20
Problem (Fix #1621): Can't deposit into other's new staking address

### LINKED ISSUE #1621: Problem: Can't deposit into other's new staking address
The intension of the restriction is to prevent wrong deposit tx, but it also block valid cases.

Options:
- remove the restriction
- add confirmation in client ui
- whitelist the operator

Decision: Option2
- Remove the restriction from client network ops
- add a check and double confirmation in client-cli.

yihuang:
Solution:
- Remove the restriction from client network ops code
- Add check and double confirmation in client-cli

### DIFF
--- a/client-network/src/network_ops/default_network_ops_client.rs
+++ b/client-network/src/network_ops/default_network_ops_client.rs
@@ -139,18 +139,14 @@ where
         attributes: StakedStateOpAttributes,
         verify_staking: bool,
     ) -> Result<(TxAux, TransactionPending)> {
-        // if the to_address belongs to current wallet, we do not check the state
-        let staking_addresses = self.wallet_client.staking_addresses(name, enckey)?;
-        if !staking_addresses.contains(&to_address) {
-            let staked_state = self.get_staked_state(name, &to_address, verify_staking)?;
-            verify_unjailed(&staked_state).map_err(|e| {
+        if let Some(staking) = self.get_staking(name, &to_address, verify_staking)? {
+            verify_unjailed(&staking).map_err(|e| {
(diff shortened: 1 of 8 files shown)


client-network/src/network_ops/default_network_ops_client.rs
```diff
@@ -137,20 +137,7 @@ where
...
-        // if the to_address belongs to current wallet, we do not check the state
-        let staking_addresses = self.wallet_client.staking_addresses(name, enckey)?;
-        if !staking_addresses.contains(&to_address) {
-            let staked_state = self.get_staked_state(name, &to_address, verify_staking)?;
-            verify_unjailed(&staked_state).map_err(|e| {
-                Error::new(
-                    ErrorKind::ValidationError,
-                    format!("Failed to validate staking account: {}", e),
-                )
-            })?;
```
tomtau, line 151: maybe the jailing test could be optionally run (if the state exists?)

yihuang: Done

yihuang:
> maybe also bump version numbers in the affected client-* Cargo.toml

Done

tomtau:
> > maybe also bump version numbers in the affected client-* Cargo.toml
> 
> Done

client-rpc is also affected if client-network changed?

yihuang:
> > > maybe also bump version numbers in the affected client-* Cargo.toml
> > 
> > 
> > Done
> 
> client-rpc is also affected if client-network changed?

Yes, just bumped the version too, and reflected that in CHANGELOG.

PR #1626, opened 2020-05-20
Problem: Remote attestation does not verify enclave measurement and svn information

### LINKED ISSUE #1622: Problem: Remote attestation need to verify more information
https://github.com/crypto-com/chain/pull/1604#pullrequestreview-414120899

devashishdxt:
Solution: Added optional verification of enclave measurement and svn info if `enclave_info` is provided in verifier config.

Fixes #1622.


tomtau:
+ cargo fmt https://travis-ci.org/github/crypto-com/chain/jobs/689084482#L1317

devashishdxt:
> for `cpu_svn` and `isv_svn`, it's better if to check that they are greater or equal to -- or in other words, reject if the expected enclave info values are greater than the quote's ones.
> In that way, for (non-enclave -> enclave), non-enclave clients can still connect to "newer" / "updated" enclaves (from the same signer / author); for (enclave <-> enclave), newer enclaves will reject to connect to the older ones
> (If that's done, `mr_enclave` should be made optional, as one may want to connect to a newer version of the enclave code from the same signer)

Changed

PR #1637, opened 2020-05-21
Problem: `tx-query-next` does not create a new certificate after old one is expired.

devashishdxt:
Solution:
- Moved the code to set TLS certificate inside of individual connection threads.
- Some `clippy` fixes.


tomtau:
the root cause was that it didn't call "configure_server_config"?

devashishdxt:
> the root cause was that it didn't call "configure_server_config"?

It did call but only at startup. `EnclaveRaContext` regenerates new certificate after old ones are expired. But, we did not set the updated certificate in TLS sessions.

PR #1656, opened 2020-05-25
Problem: (Fix#1533) Error messages is ambiguous for node-join/unbond Txs 

lezzokafka:
Solution:
- Update the error message.


devashishdxt:
bors retry

PR #1663, opened 2020-05-26
Problem: (fix #1658) no api for `minimum unbonded time` in client-rpc

### LINKED ISSUE #1658: Problem: no api for `minimum unbonded time` in client-rpc
add api to fetch minimum unbonded time in client-rpc
that api will be used in sample wallet client.

leejw51crypto:
to display unbonded time in sample wallet client.
use evidence,max_age_duration in genesis(which is nano seconds)
added genesis, status api to client-rpc, which uses tendermint rpc (26657 port)
also when withdrawal fails by unbonded time, also returns remaining time for withdrawal

### DIFF
--- a/client-network/src/network_ops/default_network_ops_client.rs
+++ b/client-network/src/network_ops/default_network_ops_client.rs
@@ -16,7 +16,7 @@ use chain_core::tx::fee::FeeAlgorithm;
 use chain_core::tx::{TxAux, TxPublicAux};
 use chain_storage::jellyfish::SparseMerkleProof;
 use chain_tx_validation::{check_inputs_basic, check_outputs_basic, verify_unjailed};
-use client_common::tendermint::types::AbciQueryExt;
+use client_common::tendermint::types::{AbciQueryExt, Genesis, StatusResponse};
 use client_common::tendermint::Client;
 use client_common::{
     Error, ErrorKind, Result, ResultExt, SecKey, SignedTransaction, Storage, Transaction,
@@ -28,6 +28,7 @@ use client_core::{TransactionObfuscation, UnspentTransactions, WalletClient};
 use tendermint::{block::Height, Time};
 
 /// Default implementation of `NetworkOpsClient`
+#[derive(Clone)]
 pub struct DefaultNetworkOpsClient<W, S, C, F, E>
 where
     W: WalletClient,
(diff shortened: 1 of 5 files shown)


client-network/src/network_ops/default_network_ops_client.rs
```diff
@@ -254,11 +254,12 @@ where
...
         let last_block_time = self.get_last_block_time()?;
         let staked_state = self.get_staked_state(name, from_address, verify_staking)?;
-
         if staked_state.unbonded_from > last_block_time {
+            let seconds = staked_state.unbonded_from - last_block_time;
+            let duration = std::time::Duration::from_secs(seconds);
             return Err(Error::new(
                 ErrorKind::ValidationError,
-                "Staking state is not yet unbonded",
+                format!("Staking state is not yet unbonded time:left {:?}", duration),
```
linfeng-crypto, line 262: `time left:`

leejw51crypto: ok

yihuang:
Why add these api into client-rpc if you can just call tendermint rpc directly?

leejw51crypto:
just design, UI only talks with client-rpc
later it may be talk via ffi

yihuang:
> just design simplicity, UI only talks with client-rpc
> later it may be talk via ffi

How about a common proxy api for tendermint rpc then, no need to add separated ones for each tendermint rpc api?

leejw51crypto:
ok, that seems better design
i'll change it

client-rpc/src/handler.rs
```diff
@@ -72,10 +73,17 @@ impl RpcHandler {
...
+            storage.clone(),
+            tendermint_client.clone(),
+            fee_policy,
+            obfuscator.clone(),
+        )?;
 
         let multisig_rpc = MultiSigRpcImpl::new(wallet_client.clone());
         let transaction_rpc = TransactionRpcImpl::new(network_id);
         let staking_rpc = StakingRpcImpl::new(wallet_client.clone(), ops_client, network_id);
+        let info_rpc = InfoRpcImpl::new(info_ops_client);
```
yihuang, line 86:
```suggestion
        let info_rpc = InfoRpcImpl::new(ops_client.clone());
```

leejw51crypto: ok

leejw51crypto: done

Issue #1616, opened 2020-05-19
Problem: keypackage not verified in nodejointx / council node data


tomtau:
- (disabled in mock mode)
- tx validation should check that cert field in `ConfidentialInit` is a valid keypackage
- (it can also extract the security version number (ISVSVN) -- if the number is greater than the existing one it has from genesis / previously, it can display a warning)
- some basic TDBE runner that will generate keypackage + expose it somehow (dump to TCP? or just printout? custom stream?)
- client/tool support for fetching the keypackage for node join tx construction + preparation of genesis
(+ integration test modifications)

yihuang:
The `ConfidentialInit` is not accurate since the keypackage in `StakedState::Validator::CouncilNode` can also be updated?
How about simply rename to `keypackage`, and store `KeyPackage` directly?

tomtau:
> The `ConfidentialInit` is not accurate since the keypackage in `StakedState::Validator::CouncilNode` can also be updated?
> How about simply rename to `keypackage`, and store `KeyPackage` directly?

do you mean to rename `cert` to `keypackage` or renaming `ConfidentialInit`?
I think the former is fine, as the TX wire format won't change (genesis will break, but it's already had a breaking change on the current master anyway)

yihuang:
> > The `ConfidentialInit` is not accurate since the keypackage in `StakedState::Validator::CouncilNode` can also be updated?
> > How about simply rename to `keypackage`, and store `KeyPackage` directly?
> 
> do you mean to rename `cert` to `keypackage` or renaming `ConfidentialInit`?
> I think the former is fine, as the TX wire format won't change (genesis will break, but it's already had a breaking change on the current master anyway)

I was trying to rename `ConfidentialInit`, but I guess rename `cert` to `keypackage` is good enough?

tomtau:
> 
> I was trying to rename `ConfidentialInit`, but I guess rename `cert` to `keypackage` is good enough?

changing `ConfidentialInit` to `KeyPackage` would break the tx wire format -- this single `keypackage` field is ok

yihuang:
> - (it can also extract the security version number (ISVSVN) -- if the number is greater than the existing one it has from genesis / previously, it can display a warning)

Is it to warn that when a new version of enclave is appeared?

tomtau:
> Is it to warn that a new version of enclave is appeared?

Yes, it's to signal security upgrades -- more logic around that will be needed, but this is just for the start

PR #1668, opened 2020-05-26
Problem (WIP #1616): keypackage not verified in nodejointx / council node data

yihuang:
Solution:
- record most recent isv_svn, warn when a new version appears
- add mock data and fix unit/integration tests temporarily.

TBD: generate real keypackage when construct node-join tx, or prepare genesis.


chain-abci/src/staking/tx.rs
```diff
@@ -24,9 +24,9 @@ impl StakingTable {
         heap: &mut impl StoreStaking,
         block_time: Timespec,
         max_evidence_age: Timespec,
-        ra_verifier: &EnclaveCertVerifier,
+        mut recent_isv_svn: u16,
         tx: &NodeJoinRequestTx,
-    ) -> Result<(), PublicTxError> {
+    ) -> Result<u16, PublicTxError> {
```
tomtau, line 29: why it needs it `mut` when it also returns it?

chain-tx-enclave-next/enclave-ra/ra-client/src/config.rs
```diff
@@ -19,7 +19,8 @@ impl<'a> EnclaveCertVerifierConfig<'a> {
     pub fn new() -> Self {
         Self {
             signing_ca_cert_pem: IAS_CERT.into(),
-            valid_enclave_quote_statuses: vec!["OK".into()].into(),
+            // FIXME remove SW_HARDENING_NEEDED after https://github.com/rust-lang/llvm-project/pull/58
```
tomtau, line 22: I think `SW_HARDENING_NEEDED` will still appear as a response, so it may not be possible to remove it ("OK" or "UP-TO-DATE" will only appear on unaffected platforms which are only some very recent CPUs, i.e. mid-2019 and newer) -- it's more of a reminder that one should apply those software hardening measures: https://software.intel.com/security-software-guidance/insights/deep-dive-load-value-injection#mitigationguidelines

Since it affects many platforms, applying the mitigations and releasing only one version with them for all platforms is ok.

chain-abci/src/staking/tx.rs
yihuang: It'll get assigned if `info.quote.report_body.isv_svn > recent_isv_svn`, and returned at the end.

chain-tx-enclave-next/enclave-ra/ra-client/src/config.rs
yihuang: Oh, then OK, UP-TO-DATE, SW_HARDENING_NEEDED all should be considered as valid responses?

chain-abci/src/staking/tx.rs
leejw51crypto: how about 
`recent_isv_svn:&mut u16` ?
for readibility

chain-abci/src/staking/tx.rs
```diff
@@ -24,9 +24,9 @@ impl StakingTable {
         heap: &mut impl StoreStaking,
         block_time: Timespec,
         max_evidence_age: Timespec,
-        ra_verifier: &EnclaveCertVerifier,
+        mut recent_isv_svn: u16,
```
leejw51crypto, line 27: some full-word comment would be useful

chain-abci/src/storage/mod.rs
```diff
@@ -117,7 +116,12 @@ pub enum TxPublicAction {
         unbond: (StakedStateAddress, Coin),
         unbonded_from: Timespec,
     },
-    NodeJoin(StakedStateAddress, CouncilNode),
+    NodeJoin {
+        address: StakedStateAddress,
+        council_node: CouncilNode,
+        /// most recent isv_svn
```
leejw51crypto, line 122: this remark for documentation?
// ?

chain-abci/src/staking/tx.rs
yihuang: Used a new variable to return now, it would be less confusing.

chain-core/src/init/config.rs
```diff
@@ -2,85 +2,61 @@ use crate::common::Timespec;
 use crate::init::address::RedeemAddress;
```
tomtau, line 2: just one note that genesis processing may need more thought / design: https://github.com/crypto-com/chain-docs/issues/141#issuecomment-634387139

chain-abci/src/storage/mod.rs
yihuang: Done

chain-abci/src/staking/tx.rs
yihuang: added a little bit more comment here about isv svn:
https://github.com/crypto-com/chain/pull/1668/files#diff-3cfbc20473df3695f7d44c7c3d8c1508R57

chain-tx-enclave-next/enclave-ra/ra-client/src/config.rs
tomtau: The API docs (https://api.trustedservices.intel.com/documents/sgx-attestation-api-spec.pdf) don't mention "UP-TO-DATE" (it's only mentioned in that LVI mitigation article), so perhaps only "OK" and "SW_HARDENING_NEEDED"

yihuang: Done

chain-abci/fuzz/Cargo.lock
```diff
@@ -0,0 +1,3908 @@
+# This file is automatically @generated by Cargo.
```
tomtau, line 1: needed? or could this go to .gitignore?

tomtau:
A few clippy suggestions: https://travis-ci.org/github/crypto-com/chain/jobs/691601119#L1686

chain-abci/fuzz/Cargo.lock
yihuang: I think general guideline is like libraries should ignore Cargo.lock but binaries/applications should check-in Cargo.lock, I guess fuzz can be considered as application?

yihuang:
> A few clippy suggestions: https://travis-ci.org/github/crypto-com/chain/jobs/691601119#L1686

Done

chain-abci/fuzz/Cargo.lock
yihuang: Maybe we can put the fuzz crate in workspace, it'll share the same Cargo.lock file?

yihuang: Removed the lock file.

PR #1671, opened 2020-05-27
Problem:(fix #1649) no graceful close in client-cli after sync

### LINKED ISSUE #1649: Problem: no graceful close in client-cli after sync
E[2020-05-22|15:52:14.037] Failed to read request                       module=rpc-server protocol=websocket remote=127.0.0.1:39560 err="websocket: close 1006 (abnormal closure): unexpected EOF"
E[2020-05-22|15:52:14.037] Failed to read request                       module=rpc-server protocol=websocket remote=127.0.0.1:39572 err="websocket: close 1006 (abnormal closure): unexpected EOF"

before finishing sync, graceful close websocket,
also check with client-rpc, too

leejw51crypto:
- send `close` websocket message before shutting down websocket
- to fix two `EOF` errors per syncing

### DIFF
--- a/client-cli/src/command.rs
+++ b/client-cli/src/command.rs
@@ -384,9 +384,10 @@ impl Command {
                 disable_address_recovery,
                 block_height_ensure,
             } => {
+                let enckey = ask_seckey(None)?;
                 let tendermint_client = WebsocketRpcClient::new(&tendermint_url())?;
                 let tx_obfuscation = get_tx_query(tendermint_client.clone())?;
-                let enckey = ask_seckey(None)?;
+
                 let storage = SledStorage::new(storage_path())?;
                 let config = ObfuscationSyncerConfig::new(
                     storage.clone(),
@@ -399,7 +400,8 @@ impl Command {
                         block_height_ensure: *block_height_ensure,
                     },
                 );
-                Self::resync(config, name.clone(), enckey, *force, storage)
+                Self::resync(config, name.clone(), enckey, *force, storage)?;
+                Ok(())
             }
             Command::MultiSig { multisig_command } => {
                 let storage = SledStorage::new(storage_path())?;
@@ -691,6 +693,7 @@ fn print_sync_warning() {
 
 fn get_wallet_client(storage: SledStorage) -> Result<AppWalletClient> {
     let tendermint_client = WebsocketRpcClient::new(&tendermint_url())?;
+
     let hw_key_service = HwKeyService::default();
 
     let signer_manager = WalletSignerManager::new(storage.clone(), hw_key_service.clone());
--- a/client-common/src/tendermint/rpc_client/sync_rpc_client.rs
+++ b/client-common/src/tendermint/rpc_client/sync_rpc_client.rs
@@ -11,6 +11,7 @@ use tendermint::{lite, validator};
 use tokio::runtime::Runtime;
 
 use chain_core::state::ChainState;
+use std::sync::Mutex;
 
 use super::async_rpc_client::AsyncRpcClient;
 use crate::{
@@ -18,13 +19,19 @@ use crate::{
     Error, ErrorKind, Result, ResultExt,
 };
 
+use futures_util::sink::SinkExt;
+use tokio_tungstenite::tungstenite::protocol::frame::coding::CloseCode;
+use tokio_tungstenite::tungstenite::protocol::frame::CloseFrame;
+use tokio_tungstenite::tungstenite::Message;
+
 const RESPONSE_TIMEOUT: Duration = Duration::from_secs(10);
 
 /// Wraps asynchronous RPC client and executes it in tokio runtime
 #[derive(Clone)]
 pub struct SyncRpcClient {
-    runtime: Arc<Runtime>,
-    async_rpc_client: AsyncRpcClient,
+    runtime: Arc<Mutex<Runtime>>,
+    /// ASYNC RPC CLIENT
+    pub async_rpc_client: AsyncRpcClient,
 }
 
 impl SyncRpcClient {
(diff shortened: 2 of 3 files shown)


client-common/src/tendermint/rpc_client/sync_rpc_client.rs
```diff
@@ -309,4 +322,31 @@ impl Client for SyncRpcClient {
...
+/// close websocket connection by sending CLOSE message
+pub fn close_connection(writer: Arc<Mutex<WebSocketWriter>>) -> Result<()> {
+    // close
+
+    let runtime = Runtime::new().chain(|| {
+        (
+            ErrorKind::InitializationError,
+            "Unable to start tokio runtime",
+        )
+    })?;
```
devashishdxt, line 340: Why are you creating a new `Runtime` just to close the connection?

leejw51crypto:
it's the same with calling stream.close
that api also sends CloseFrame, because stream is owned in async-rpc-client.
but i'll check Drop , too.

client-common/src/tendermint/rpc_client/sync_rpc_client.rs
leejw51crypto: because `close` is actually `sending close message`, it's async , future.
so to run it, needed runtime.
sharing runtime with original is a little complex, so chose making a new tokio runtime to run the future

devashishdxt: If you do it in `Drop` implementation of `SyncRpcClient`, you'll also have access to the main `Runtime`?

leejw51crypto: ok, i'll try to do it on Drop

leejw51crypto:
ok, i'll rebase

Drop working fine,
i'm changing that, too.

leejw51crypto:
i tried to use self.runtime,
but it should be mutable to use block_on,
so i created runtime

client-common/src/tendermint/rpc_client/sync_rpc_client.rs
```diff
@@ -310,3 +316,27 @@ impl Client for SyncRpcClient {
             .collect()
     }
 }
+
+impl Drop for SyncRpcClient {
+    fn drop(&mut self) {
+        let sender = self.async_rpc_client.websocket_writer.clone();
+
+        let mut runtime = Runtime::new()
```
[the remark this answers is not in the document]
leejw51crypto: it's not allowed in Arc,
but i'll try again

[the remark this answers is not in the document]
leejw51crypto: ooooh, it works
i submitted

client-cli/src/command.rs
```diff
@@ -385,7 +385,9 @@ impl Command {
                 block_height_ensure,
             } => {
                 let tendermint_client = WebsocketRpcClient::new(&tendermint_url())?;
-                let tx_obfuscation = get_tx_query(tendermint_client.clone())?;
+                let tendermint_client_tx_query = WebsocketRpcClient::new(&tendermint_url())?;
```
yihuang, line 388: Why two connections?

yihuang:
Just a side note, someone might merge this much simpler version of implementation(https://gist.github.com/yihuang/a512e18b5bc59953b05255ffcf4ddc44) ;D

devashishdxt:
> Just a side note, shall we merge this much simpler version of implementation sometime (https://gist.github.com/yihuang/a512e18b5bc59953b05255ffcf4ddc44) ;D

I think `smol` is still very new and there is not much ecosystem around it. `tokio` runtime is used in most places, for example, it is also used in our explorer backend code so its very easy to integrate that with `client-*` crates. We can think of using `smol` when there are good frameworks on top of `smol`. `smol` is definitely very simple but it is also very new.

client-cli/src/command.rs
leejw51crypto: if it cloned,  some disconnect message, occurs, 
i'll show it shortly

tomtau:
bors try

leejw51crypto:
i'll check

client-cli/src/command.rs
leejw51crypto: remove two connections

tomtau:
bors try

client-common/src/tendermint/rpc_client/sync_rpc_client.rs
```diff
@@ -11,20 +11,29 @@ use tendermint::{lite, validator};
...
 /// Wraps asynchronous RPC client and executes it in tokio runtime
 #[derive(Clone)]
 pub struct SyncRpcClient {
-    runtime: Arc<Runtime>,
-    async_rpc_client: AsyncRpcClient,
+    runtime: Arc<Mutex<Runtime>>,
+    /// ASYNC RPC CLIENT
+    pub async_rpc_client: AsyncRpcClient,
+    /// graceful close on Drop
+    pub disconnect_on_drop: bool,
```
devashishdxt, line 36: Why do we need a flag for this? Shouldn't this be the default behaviour?

leejw51crypto: RpcClient is cloned, without that, integration test breaks.
also don't to shutdown it gracefully for all cases

client-common/src/tendermint/rpc_client/sync_rpc_client.rs
```diff
@@ -310,3 +320,20 @@ impl Client for SyncRpcClient {
             .collect()
     }
 }
+
+impl Drop for SyncRpcClient {
+    fn drop(&mut self) {
+        if self.disconnect_on_drop {
```
devashishdxt, line 326: Instead of having a boolean flag, you can check following condition:
```
if Arc::strong_count(&self.runtime) == 1 {
    // Close connection
}
```

leejw51crypto: ok, that's better solution.
i'll check it

tomtau:
bors try

PR #1678, opened 2020-05-28
problem: (fix #1610) can not find wallet name in wallet list after import success from a file

### LINKED ISSUE #1610: Problem: can not find wallet name in wallet list after import success from a file
import some wallet from a file using `client-cli wallet import -f a_file`, wallet name is not listed in the result of `client-cli wallet list`, but other functions like `client-cli view-key` or `client-cli wallet delete` can work well.

linfeng-crypto:
solution:
* fix encode/decode method for Wallet
* add key pairs and multisig addresses in WalletInfo
* restore key pair, private keys, multisig addresses and root hash when import wallet

### DIFF
--- a/client-core/src/service/wallet_service.rs
+++ b/client-core/src/service/wallet_service.rs
@@ -14,8 +14,10 @@ use secstr::SecUtf8;
 use serde::de::{self, Visitor};
 use serde::export::PhantomData;
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
+use std::collections::BTreeMap;
 use std::fmt;
 use std::str;
+
 /// Key space of wallet
 const KEYSPACE: &str = "core_wallet";
 
@@ -101,11 +103,15 @@ pub struct WalletInfo {
     /// wallet meta data
     #[serde(deserialize_with = "deserde_from_str", serialize_with = "serde_to_str")]
     pub wallet: Wallet,
-    /// private key of the wallet
+    /// private key of view key pair
     #[serde(deserialize_with = "deserde_from_str", serialize_with = "serde_to_str")]
     pub private_key: PrivateKey,
     /// passphrase used when import wallet
     pub passphrase: Option<SecUtf8>,
+    ///public_key -> encoded private_key pairs, private key is None for hardware wallet
+    pub key_pairs: BTreeMap<PublicKey, Option<Vec<u8>>>,
+    /// hex encoded root_hash -> parity_scale_codec encoded multisig_address pairs
+    pub multisig_address_pair: BTreeMap<String, Vec<u8>>,
 }
 
 /// Wallet meta data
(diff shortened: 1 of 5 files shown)


tomtau:
bors try

client-core/src/service/wallet_service.rs
```diff
@@ -101,11 +103,15 @@ pub struct WalletInfo {
     /// wallet meta data
     #[serde(deserialize_with = "deserde_from_str", serialize_with = "serde_to_str")]
     pub wallet: Wallet,
-    /// private key of the wallet
+    /// private key of view key
```
tomtau, line 106: of view key pair?

client-core/src/service/wallet_service.rs
```diff
@@ -101,11 +103,15 @@ pub struct WalletInfo {
...
     #[serde(deserialize_with = "deserde_from_str", serialize_with = "serde_to_str")]
     pub private_key: PrivateKey,
     /// passphrase used when import wallet
     pub passphrase: Option<SecUtf8>,
+    /// private_key: public_key-> private_key paire
+    pub key_pairs: BTreeMap<PublicKey, Option<Vec<u8>>>,
```
tomtau, line 112: in watch-only it'll be None? or why is it `Option`? and perhaps `SecretVec<u8>`?

client-core/src/service/wallet_service.rs
```diff
@@ -101,11 +103,15 @@ pub struct WalletInfo {
...
+    /// multisig_address: hex encoded root_hash -> multisig_address pair
+    pub multisig_addresses: BTreeMap<String, Vec<u8>>,
```
tomtau, line 114: `MultiSigAddress` scale-encoded is the Vec<u8>?

client-core/src/service/wallet_service.rs
linfeng-crypto: if hardware wallet, we can't get private key

client-core/src/service/wallet_service.rs
linfeng-crypto: Ok

client-core/src/service/wallet_service.rs
linfeng-crypto: yes, it's parity_scale_codec encoded multisig_address.

yihuang:
What is the root cause of original issue: "can't find wallet name and addresses after import a wallet", why the solution involves changes to `WalletInfo`?

client-core/src/service/wallet_service.rs
```diff
@@ -101,11 +103,15 @@ pub struct WalletInfo {
...
-    /// private key of the wallet
+    /// private key of view key pair
     #[serde(deserialize_with = "deserde_from_str", serialize_with = "serde_to_str")]
     pub private_key: PrivateKey,
     /// passphrase used when import wallet
     pub passphrase: Option<SecUtf8>,
+    ///public_key -> encoded private_key pairs, private key is None for hardware wallet
+    pub key_pairs: BTreeMap<PublicKey, Option<Vec<u8>>>,
+    /// hex encoded root_hash -> parity_scale_codec encoded multisig_address pairs
+    pub multisig_address_pair: BTreeMap<String, Vec<u8>>,
```
yihuang, line 114: What's the purpose of new fields: "key_pairs", "multisig_address_pair"

client-core/src/service/wallet_service.rs
tomtau: ok, it'll be good to mention in the comment

client-core/src/service/wallet_service.rs
linfeng-crypto: * key-pairs: before the key-pairs stored in `Wallet`, and @leejw51 moved them out of it and restored as key-value directly in DB, so we should add it.

* multisig_address_pair: same as key-pair, or the transfer addresses and multisig address will be lost after wallet imported.

PR #1686, opened 2020-05-29
Problem (Fix #1667): staking address witness checking is not delayed

### LINKED ISSUE #1667: Problem: address witness checking is not delayed
Fix the TODO in 
https://github.com/crypto-com/chain/blob/master/chain-abci/src/storage/mod.rs#L296

delay witness checking can potentially save some cpu.

yihuang:
Solution:
- Move witness checking after logic processing, could save some cpu
  since witness checking is more costly then logic.


yihuang:
> those `staking_table.<something>` modify staking_store -- what if everything in the tx is ok, but not the signature? then the `staking_table` gets modified with invalid tx?

The tx logic only mutates after all verification passed, if a failed verification leave state mutated, it's definitely a bug.

PR #1693, opened 2020-06-01
Problem: `tx-query2-app-runner` build fails

### LINKED ISSUE #1235: Problem: ported tx-query not tested

### LINKED ISSUE #1651: Problem: build tx-query2-app-runner failed
We need to upgrade tx-query2-app-runner to tokio 0.2
```
error[E0277]: the trait bound `zmq_helper::ZmqHelper: tokio::io::async_read::AsyncRead` is not satisfied
  --> chain-tx-enclave/tx-query-next/app-runner/src/main.rs:33:62
   |
33 |                     let boxed_stream: Box<dyn AsyncStream> = Box::new(stream);
   |                                                              ^^^^^^^^^^^^^^^^ the trait `tokio::io::async_read::AsyncRead` is not implemented for `zmq_helper::ZmqHelper`
   |
   = note: required because of the requirements on the impl of `enclave_runner::usercalls::AsyncStream` for `zmq_helper::ZmqHelper`
   = note: required for the cast to the object type `dyn enclave_runner::usercalls::AsyncStream`

error[E0277]: the trait bound `zmq_helper::ZmqHelper: tokio::io::async_write::AsyncWrite` is not satisfied
  --> chain-tx-enclave/tx-query-next/app-runner/src/main.rs:33:62
   |
33 |                     let boxed_stream: Box<dyn AsyncStream> = Box::new(stream);
   |                                                              ^^^^^^^^^^^^^^^^ the trait `tokio::io::async_write::AsyncWrite` is not implemented for `zmq_helper::ZmqHelper`
   |
   = note: required because of the requirements on the impl of `enclave_runner::usercalls::AsyncStream` for `zmq_helper::ZmqHelper`
   = note: required for the cast to the object type `dyn enclave_runner::usercalls::AsyncStream`

error: aborting due to 2 previous errors

For more information about this error, try `rustc --explain E0277`.
error: could not compile `tx-query2-app-runner`.
```

devashishdxt:
Solution:
- Upgraded `tokio` to `v0.2`.
- Started using `tx-query2-enclave-app` in integration tests.
- Made `edp-obfuscation` default feature in `client-core`.
- Fixes #1651
- Fixes #1235


devashishdxt:
bors try

.gitignore
```diff
@@ -37,3 +37,8 @@ id_ecdsa
 /integration-tests/bot/.venv
 
 **/chain-abci/Enclave_u.*
+
+
+# Drone temparary files
+secrets.txt
```
yihuang, line 43: There's an existing ignore item for this: `/.drone.secret`

client-core/Cargo.toml
```diff
@@ -53,7 +53,7 @@ ripemd160 = "0.8.0"
 test-common = { path = "../test-common" }
 
 [features]
-default = ["sled", "sgx-obfuscation"]
+default = ["sled", "edp-obfuscation"]
```
yihuang, line 56: It seems to me, no need to keep features `edp-obfuscation` and `sgx-obfuscation`? only need one of them, and they can be build in normal environment, right?

integration-tests/run.sh
```diff
@@ -14,8 +14,10 @@ source $PYTHON_VENV_DIR/bin/activate
 export CARGO_TARGET_DIR=${CARGO_TARGET_DIR:-"../target"}
 BUILD_PROFILE=${BUILD_PROFILE:-debug}
 BUILD_MODE=${BUILD_MODE:-sgx}
-ln -sf $CARGO_TARGET_DIR/$BUILD_PROFILE/tx_query_enclave.signed.so .
+EDP_TARGET_DIR=$CARGO_TARGET_DIR/x86_64-fortanix-unknown-sgx/$BUILD_PROFILE
 ln -sf $CARGO_TARGET_DIR/$BUILD_PROFILE/tx_validation_enclave.signed.so .
+ln -sf $EDP_TARGET_DIR/tx-query2-enclave-app.sgxs .
+ln -sf $EDP_TARGET_DIR/tx-query2-enclave-app.sig .
```
yihuang, line 20: You can cleanup these two symlinks in `integration-tests/cleanup.sh`.

integration-tests/run_multinode.sh
```diff
@@ -14,7 +14,6 @@ source $PYTHON_VENV_DIR/bin/activate
 CARGO_TARGET_DIR=${CARGO_TARGET_DIR:-"../target"}
 BUILD_PROFILE=${BUILD_PROFILE:-debug}
 BUILD_MODE=${BUILD_MODE:-sgx}
-ln -sf $CARGO_TARGET_DIR/$BUILD_PROFILE/tx_query_enclave.signed.so .
 ln -sf $CARGO_TARGET_DIR/$BUILD_PROFILE/tx_validation_enclave.signed.so .
 export PATH=$CARGO_TARGET_DIR/$BUILD_PROFILE:$PATH
```
yihuang, line 19: I think you need to create the symlinks of tx-query2 enclave here too?

chain-tx-enclave-next/tx-query-next/app-runner/src/zmq_helper.rs
```diff
@@ -0,0 +1,189 @@
...
+    fn poll_read(
+        self: Pin<&mut Self>,
+        cx: &mut Context,
+        buf: &mut [u8],
+    ) -> Poll<io::Result<usize>> {
+        let mut inner_future = Box::pin(self.inner.lock());
+        let mut inner = ready!(inner_future.as_mut().poll(cx));
+
+        if inner.state.is_ready() {
+            log::debug!("Unable to receive message. Previous request wasn't sent");
```
yihuang, line 92: Should this be an error or panic?

chain-tx-enclave-next/tx-query-next/app-runner/src/zmq_helper.rs
```diff
@@ -0,0 +1,189 @@
...
+            match response {
+                Ok(_) => {
+                    inner.state = SocketState::RequestSent;
+                    Poll::Ready(Ok(buf.len()))
+                }
+                Err(e) => Poll::Ready(Err(into_io_err(e))),
+            }
+        } else {
+            log::debug!("Unable to send message. Previous response isn't processed");
+            Poll::Pending
```
yihuang, line 157: https://github.com/crypto-com/chain/pull/1693/files#r433167721

client-core/Cargo.toml
devashishdxt: Eventually, we'll depricate `sgx-obfuscation` after `edp-obfuscation` is sufficiently tested.

chain-tx-enclave-next/tx-query-next/app-runner/src/zmq_helper.rs
devashishdxt: I don't think we should panic here. I'll add a way to wake all the tasks waiting for state change.

chain-tx-enclave-next/tx-query-next/app-runner/src/zmq_helper.rs
devashishdxt: https://github.com/crypto-com/chain/pull/1693#discussion_r433363468

chain-tx-enclave-next/tx-query-next/app-runner/src/main.rs
```diff
@@ -0,0 +1,112 @@
...
+}
+
+#[allow(clippy::type_complexity)]
+impl UsercallExtension for Options {
+    fn connect_stream<'future>(
+        &'future self,
+        addr: &'future str,
+        _local_addr: Option<&'future mut String>,
+        _peer_addr: Option<&'future mut String>,
+    ) -> Pin<Box<dyn Future<Output = io::Result<Option<Box<dyn AsyncStream>>>> + 'future>> {
```
tomtau, line 45: the desugared type is still needed even after the upgrade?

chain-tx-enclave-next/tx-query-next/app-runner/src/main.rs
```diff
@@ -0,0 +1,112 @@
...
+
+    let mut device = Device::new()
+        .expect("SGX device was not found")
+        .einittoken_provider(AesmClient::new())
+        .build();
+    let enclave_path = options.enclave_path.clone();
+    let mut enclave_builder = EnclaveBuilder::new(enclave_path.as_ref());
+
+    if options.test {
+        enclave_builder.dummy_signature();
```
tomtau, line 99: perhaps log a warning that it's launched in test / with dummy signature? could be good to know when testing with some scripts that will add these flags

chain-tx-enclave-next/tx-query-next/app-runner/src/zmq_helper.rs
```diff
@@ -0,0 +1,189 @@
+use std::{
+    convert::TryInto,
+    future::Future,
+    pin::Pin,
+    task::{Context, Poll},
+    vec::IntoIter,
+};
+
+use async_zmq::{request, Message, MultipartIter, Request};
```
tomtau, line 9: what's a reason for `async_zmq` ? Existing zmq bridging used synchronous REQ-REP which still be wrapped?

chain-tx-enclave-next/tx-query-next/app-runner/src/zmq_helper.rs
```diff
@@ -0,0 +1,189 @@
...
+impl AsyncRead for ZmqHelper {
+    fn poll_read(
+        self: Pin<&mut Self>,
+        cx: &mut Context,
+        buf: &mut [u8],
+    ) -> Poll<io::Result<usize>> {
+        let mut inner_future = Box::pin(self.inner.lock());
+        let mut inner = ready!(inner_future.as_mut().poll(cx));
+
+        if inner.state.is_ready() {
```
tomtau, line 91: why have these if-elses with `unreachable!` branch if one could just pattern match on `inner.state` directly ?

chain-tx-enclave-next/tx-query-next/app-runner/src/zmq_helper.rs
```diff
@@ -0,0 +1,189 @@
...
+    }
+}
+
+impl AsyncRead for ZmqHelper {
+    fn poll_read(
+        self: Pin<&mut Self>,
+        cx: &mut Context,
+        buf: &mut [u8],
+    ) -> Poll<io::Result<usize>> {
+        let mut inner_future = Box::pin(self.inner.lock());
```
tomtau, line 88: how does this ensure, if there are multiple concurrent requests/replies, that one isn't locking and reading reply from another request?

ci-scripts/build-tx-query.sh
```diff
@@ -0,0 +1,15 @@
+set -e
+
+APP_NAME=tx-query2-enclave-app
+RUNNER_NAME=tx-query2-app-runner
+PRIVATE_KEY_PATH=chain-tx-enclave/tx-query/enclave/Enclave_private.pem
+
+# Build enclave app
+RUSTFLAGS="-Ctarget-feature=+aes,+sse2,+sse4.1,+ssse3,+pclmul -Clinker=gold -Clink-args=-shared" cargo build --package $APP_NAME --target x86_64-fortanix-unknown-sgx
```
tomtau, line 8: is this script ok? a few days ago, it seemed this config was problematic, so one should still go with rust-lld and just add `CFLAGS="-gz=none"`?

tomtau: or is this script used?

client-core/src/cipher/default.rs
```diff
@@ -44,18 +40,7 @@ fn get_tls_config() -> Result<Arc<rustls::ClientConfig>> {
...
-        report_validity_secs: 86400,
-        enclave_info: None, // TODO: Get enclave details from command line or env variables?
-    };
-    let verifier = EnclaveCertVerifier::new(verifier_config).chain(|| {
-        (
-            ErrorKind::InitializationError,
-            "Cannot initialize enclave certificate verifier",
-        )
-    })?;
+    let verifier = EnclaveCertVerifier::default();
```
tomtau, line 43: fixme: compile-time enclave details for verification

ci-scripts/build-tx-query.sh
devashishdxt: This is not used anymore. I'll delete this.

chain-tx-enclave-next/tx-query-next/app-runner/src/zmq_helper.rs
devashishdxt: We cannot send two concurrent requests. Before sending a request, we check if the socket is in ready state and only send request if it is in ready state.

chain-tx-enclave-next/tx-query-next/app-runner/src/zmq_helper.rs
devashishdxt: `async-zmq` hooks `zmq`'s notification file descriptors with `tokio` runtime to get async IO.

chain-tx-enclave-next/tx-query-next/app-runner/src/main.rs
devashishdxt: Yes. `async fn`s are not allowed in traits.

chain-tx-enclave-next/tx-query-next/app-runner/src/zmq_helper.rs
devashishdxt: Pattern matching on `inner.state` takes reference to the state and we cannot modify state if we have a reference to it. This is a workaround for that.

devashishdxt:
bors try

devashishdxt:
bors try

chain-tx-enclave-next/tx-query-next/app-runner/src/zmq_helper.rs
devashishdxt: I added a waking mechanism for tasks waiting for ready state. At other places, it is ok to panic because those are logically unreachable.

chain-tx-enclave-next/tx-query-next/app-runner/src/zmq_helper.rs
devashishdxt: https://github.com/crypto-com/chain/pull/1693#discussion_r433602073

chain-tx-enclave-next/tx-query-next/app-runner/src/zmq_helper.rs
yihuang: I feel zmq is not worth it in our use case, more complex and probably less efficient than plain tcp connections.

devashishdxt: I agree. In future, we may remove ZeroMQ in favour of attested TCP channel.

chain-tx-enclave-next/tx-query-next/app-runner/src/zmq_helper.rs
tomtau: Tried rewriting it as:
```
        let mut wakeup = false;
        let result = match inner_state {
            SocketState::Ready => {
                unreachable!("Unable to receive message. Previous request wasn't sent")
            },
            SocketState::RequestSent => {
                let response = {
                        let mut future = Box::pin(inner.socket.recv());
                        ready!(future.as_mut().poll(cx))
                    };
                match response {
                    Ok(multipart) => {
                        // We don't expect multipart messages from ZeroMQ. Assuming only one message
                        assert_eq!(
                            1,
                            multipart.len(),
                            "Received multiple messages in ZeroMQ response"
                        );

                        let response = multipart.into_iter().next().unwrap();
                        let response_len: u32 = response
                            .len()
                            .try_into()
                            .expect("Message length exceeds `u32` bounds");
                        buf.copy_from_slice(&response_len.to_le_bytes());

                        inner.state = SocketState::ResponseReceived(response);

                        Poll::Ready(Ok(std::mem::size_of::<u32>()))
                    }
                    Err(e) => Poll::Ready(Err(into_io_err(e))),
                }
            }
            SocketState::ResponseReceived(msg) => {
                buf.copy_from_slice(&msg);
                wakeup = true;

                Poll::Ready(Ok(msg.len()))
            }
        };
        if wakeup {
            inner.wake_next();
        }
        result
```
it seems to compile and is simpler?

devashishdxt: Both are pretty much the same things.

devashishdxt: I used the functions approach because that is the idiomatic way for doing the same thing for `Option`. [Example](https://github.com/crypto-com/chain/blob/a05b1cd101e3cab55a010fe86c9814bf116d903e/chain-abci/src/main.rs#L49).

chain-tx-enclave-next/tx-query-next/app-runner/src/zmq_helper.rs
tomtau: where does that make a difference? that `inner.socket.send` / `unner.socket.recv`won't block?

devashishdxt: Yes. They won't block. They are `async` functions (https://docs.rs/async_zmq/0.3.2/async_zmq/request/struct.Request.html).

chain-tx-enclave-next/tx-query-next/app-runner/src/zmq_helper.rs
tomtau: It's the same -- it's just a suggestion to avoid boilerplate, as it's perhaps 50 lines shorter with pattern match (no need for that `impl Socket` stuff etc.) and contains fewer foot guns for future code modifications (it's an exhaustive match, so no ```else {
             unreachable!()
        ``` } catch all)

devashishdxt: Ok. I’ll change.

devashishdxt: Changed.

tomtau:
bors try

devashishdxt:
> looks ok -- before moving to chain-tx-enclave-next, it may need some extra fixes or feature guarded (against target_env), as kcov compilation fails: https://travis-ci.org/github/crypto-com/chain/jobs/693715918#L2424

Is there a way to remove `tx-query-next/enclave-app`

tomtau:
> > looks ok -- before moving to chain-tx-enclave-next, it may need some extra fixes or feature guarded (against target_env), as kcov compilation fails: https://travis-ci.org/github/crypto-com/chain/jobs/693715918#L2424
> 
> Is there a way to remove `tx-query-next/enclave-app`

Some options:
- put it back to `chain-tx-enclave`, then this hack does the trick: https://github.com/crypto-com/chain/blob/master/.travis.yml#L81
- move that main and stuff into a separate module, and in main.rs have two mains -- one for sgx target_env (runs enclave_main / the current main), and one "not sgx target_env" (does nothing / prints warning?)

@yihuang any ideas?

yihuang:
> > > looks ok -- before moving to chain-tx-enclave-next, it may need some extra fixes or feature guarded (against target_env), as kcov compilation fails: https://travis-ci.org/github/crypto-com/chain/jobs/693715918#L2424
> > 
> > 
> > Is there a way to remove `tx-query-next/enclave-app`
> 
> Some options:
> 
> * put it back to `chain-tx-enclave`, then this hack does the trick: https://github.com/crypto-com/chain/blob/master/.travis.yml#L81
> * move that main and stuff into a separate module, and in main.rs have two mains -- one for sgx target_env (runs enclave_main / the current main), and one "not sgx target_env" (does nothing / prints warning?)
> 
> @yihuang any ideas?

Maybe easier to feature guard the whole crate, but keep the crate itself, like https://github.com/crypto-com/chain/blob/master/chain-tx-enclave-next/mls/src/main.rs.

devashishdxt:
>     * move that main and stuff into a separate module, and in main.rs have two mains -- one for sgx target_env (runs enclave_main / the current main), and one "not sgx target_env" (does nothing / prints warning?)

Done.

chain-tx-enclave-next/tx-query-next/app-runner/src/zmq_helper.rs
leejw51crypto: but , zeromq is handy to use, for raw tcp channel, need to make a layer like zmq

Issue #1703, opened 2020-06-03
Problem: incorrect tree hashing in mls modules


tomtau:
1. some of the tree math methods are to be fixed in the spec: https://github.com/mlswg/mls-protocol/pull/327/files#diff-a87cc9081154f564150420544170f1c9L2898

2. currently, the code operates on plain usize, but it's potentially error-prone, as there are two kinds of indices (node position vs leaf position) -- similarly to the reference implementation: https://github.com/cisco/go-mls/blob/master/tree-math.go#L28 it'll be good to have opaque types (`LeafSize` and `NodeSize` as with Rust naming conventions?) to distinguish between these two

3. tree implementation misses some parts, the existing code may be wrong (e.g. due to 2) -- passing node index instead of leaf index and vice versa) and integrity checking isn'd done in the welcome message processing:
> For each non-empty parent node, verify that exactly one of the node's children are non-empty and have the hash of this node set as their parent_hash value (if the child is another parent) or has a parent_hash extension in the KeyPackage containing the same value (if the child is a leaf).
>
>For each non-empty leaf node, verify the signature on the KeyPackage.

PR #1740, opened 2020-06-08
Problem (WIP #1703): incorrect tree hashing in mls modules

yihuang:
Solution:
- included proposed fixes (5ee9f5d) to the tree math draft
- explicit types for node and leaf indices

### DIFF
--- a/chain-tx-enclave-next/mls/src/tree.rs
+++ b/chain-tx-enclave-next/mls/src/tree.rs
@@ -4,6 +4,7 @@ use crate::keypackage::{
     self as kp, KeyPackage, Timespec, MLS10_128_DHKEMP256_AES128GCM_SHA256_P256,
 };
 use crate::message::*;
+use crate::tree_math::{LeafSize, NodeSize};
 use crate::utils::{decode_option, encode_option, encode_vec_u32, read_vec_u32};
 use ra_client::AttestedCertVerifier;
 use rustls::internal::msgs::codec::{self, Codec, Reader};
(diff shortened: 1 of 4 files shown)


chain-tx-enclave-next/mls/src/tree_math.rs
```diff
@@ -0,0 +1,334 @@
...
+    let lvl = level(x);
+    if lvl == 0 {
+        None
+    } else {
+        let nodes = node_width(leafs);
+        // 01x -> 10x
+        let mut r = x ^ (0b11 << (lvl - 1));
+        while r >= nodes {
+            // TODO proof
+            r = left(r).unwrap();
```
tomtau, line 92: or `r = left(r)?;`

chain-tx-enclave-next/mls/src/tree_math.rs
```diff
@@ -0,0 +1,334 @@
...
+#[inline]
+pub fn parent(x: NodeSize, leafs: LeafSize) -> Option<NodeSize> {
+    let nodes = node_width(leafs);
+    assert!(x < nodes);
+    if x == root(leafs) {
+        return None;
+    }
+    let mut p = parent_step(x);
+    while p >= nodes {
+        p = parent_step(p)
```
tomtau, line 185: works without a semicolon at the end?

chain-tx-enclave-next/mls/src/tree_math.rs
```diff
@@ -0,0 +1,334 @@
...
+/// The direct path of a node, ordered from the root
+/// down, including the root
+#[inline]
+pub fn direct_path(node_pos: NodeSize, leafs: LeafSize) -> Vec<NodeSize> {
+    let mut d = Vec::new();
+    if let Some(mut p) = parent(node_pos, leafs) {
+        let r = root(leafs);
+        while p != r {
+            d.push(p);
+            p = parent(p, leafs).unwrap();
```
tomtau, line 199: proof / no panic?

chain-tx-enclave-next/mls/src/tree_math.rs
```diff
@@ -0,0 +1,334 @@
...
+
+/// The level of a node in the tree. Leaves are level 0, their parents are
+/// level 1, etc. If a node's children are at different levels, then its
+/// level is the max level of its children plus one.
+#[inline]
+pub fn level(x: NodeSize) -> u32 {
+    #[cfg(feature = "nightly")]
+    let n = x.trailing_ones();
+    #[cfg(not(feature = "nightly"))]
+    let n = {
```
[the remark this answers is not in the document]
yihuang: I kept the simpler version in the comment, we can use that when the feature become stable. it's also more efficient using special instructions.

chain-tx-enclave-next/mls/src/tree_math.rs
```diff
@@ -0,0 +1,334 @@
...
+        if x == 0 {
+            return 0;
+        }
+
+        let mut k = 0;
+        while (x >> k) > 0 {
+            k += 1
+        }
+        k - 1
+    };
```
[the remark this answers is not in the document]
yihuang: I kept the simpler version in the comment, we can use that when the feature become stable. it's also more efficient using special instructions.

chain-tx-enclave-next/mls/src/tree_math.rs
```diff
@@ -0,0 +1,334 @@
...
+//!     /    \
+//!   P00X  P10X
+//!   ```
+//!   - `P` is a shared prefix
+//!   - `X` is a sequence of ones
+//!
+//!   So moving from parent to children or vice versa, is only about switching one or two bits.
+//!
+//! The left sub-tree of the root node is always complete, but the right part might not.
+type NodeSize = usize;
```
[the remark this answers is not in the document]
yihuang: Changed to newtype and methods.

chain-tx-enclave-next/mls/src/tree.rs
```diff
@@ -158,117 +159,6 @@ pub struct Tree {
...
-        p = parent(p, n);
-    }
-    if node_pos != r {
-        d.push(r);
-    }
-    d
-}
-
 impl Tree {
     pub fn get_package(&self, leaf_index: usize) -> Option<&KeyPackage> {
```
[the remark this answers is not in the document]
yihuang: Done

chain-tx-enclave-next/mls/src/tree_math.rs
yihuang: Done

chain-tx-enclave-next/mls/src/tree_math.rs
yihuang: It seems to work without it, but still I added the semicolon.

chain-tx-enclave-next/mls/src/tree_math.rs
yihuang: Added no panic comment

PR #1741, opened 2020-06-08
Problem: `der-parser` 4.0 fails to compile

### LINKED ISSUE #1728: Problem: der-parser 4.0.0 fails to compile
https://travis-ci.org/github/crypto-com/chain/jobs/694997080#L903

devashishdxt:
Solution: Upgrade `der-parser` to `4.0` and `x509-parser` to `0.8.0-beta1`. Fixes #1728.


chain-tx-enclave-next/enclave-ra/ra-client/src/verifier.rs
```diff
@@ -108,7 +108,8 @@ impl EnclaveCertVerifier {
             return Err(EnclaveCertVerifierError::CertificateExpired);
         }
 
-        let attestation_report_oid = Oid::from(OID_EXTENSION_ATTESTATION_REPORT);
+        let attestation_report_oid = Oid::from(OID_EXTENSION_ATTESTATION_REPORT)
+            .expect("Unable to parse attestation report OID");
```
tomtau, line 112: so the parser panics if one constructs a malicious cert? 
extending and returning `EnclaveCertVerifierError` seems safer?

devashishdxt: This is just creating an Object ID. `OID_EXTENSION_ATTESTATION_REPORT` is a compile time constant.

tomtau: ok that's all right

Issue #1491, opened 2020-04-27
Problem: client-* breaks PoLP


tomtau:
Currently, no matter what action is being done, client internals allow accessing or doing anything -- e.g. if one wants to only read the balance or history, one could easily grab the private keys for spending at that moment -- which breaks [PoLP](https://en.wikipedia.org/wiki/Principle_of_least_privilege).
One start could be adding capabilities to seckey: https://github.com/crypto-com/chain/blob/master/client-common/src/seckey.rs#L14

yihuang:
Aren't we using the same encryption key for all the data currently? Maybe use a different encryption key for spending key?

tomtau:
> Aren't we using the same encryption key for all the data currently? Maybe use a different encryption key for spending key?

yeah, that will be the second step.
the first step may be to add the capabilities to seckey (make the current seckey to have all capabilities) and mark different internal APIs what capability they require (it may be a big change, so can be a PR on its own).
the second step will be to derive multiple keys

tomtau:
one potential thing to look into: https://github.com/casbin/casbin-rs

tomtau:
examples:
https://web.archive.org/web/20180129173236/http://zsck.co/writing/capability-based-apis.html
https://docs.rs/secp256k1/0.17.2/secp256k1/#enums

hsluoyz:
Hi @tomtau I'm from Casbin team. Casbin is a very popular authorization library in Go (https://casbin.org/) and we have ported it to Rust: https://github.com/casbin/casbin-rs . It supports classic access control models like ACL, RBAC, ABAC, RESTful, etc. Please have a try with it and let me know if you have any questions.

tomtau:
Hi @hsluoyz , thanks! Casbin is very nice, but it may be an overkill here (a simple command line interface or rpc application) -- maybe some basic mode without users could work.
ideally, we would also want to separate out internals on the type level as in https://github.com/crypto-com/chain/issues/1491#issuecomment-665554430 -- we'd probably start with that, see how it evolves and whether it'd make sense to add this sort of authorization middleware later

PR #1743, opened 2020-06-08
Problem: no 12 words mnemonics option for HDWallet (fix #1694)

### LINKED ISSUE #1694: Problem: no 12 words mnemonics option for HDWallet
in creating hdwallet, receive mnemonic types
Words12, Words24

also write unit-test for above

leejw51crypto:
Solution: add mnemonics word count in wallet creation
for example)
word count ==12, 12 mnemonics
word count ==24, 24 mnemonics

### DIFF
--- a/client-cli/src/command/wallet_command.rs
+++ b/client-cli/src/command/wallet_command.rs
@@ -35,6 +35,14 @@ pub enum WalletCommand {
             case_insensitive = true
         )]
         wallet_type: WalletKind,
+        #[structopt(
+            name = "wallet mnemonic count",
+            short = "m",
+            long = "mnemonics_word_count",
+            default_value = "24",
+            help = "Number of words in mnemonics"
+        )]
+        mnemonics_word_count: u32,
     },
     #[structopt(name = "export", about = "Backup wallet to a file")]
     Export {
@@ -120,9 +128,11 @@ pub enum WalletCommand {
 impl WalletCommand {
     pub fn execute<T: WalletClient>(&self, wallet_client: T) -> Result<()> {
         match self {
-            WalletCommand::New { name, wallet_type } => {
-                Self::new_wallet(wallet_client, name, *wallet_type)
-            }
+            WalletCommand::New {
+                name,
+                wallet_type,
+                mnemonics_word_count,
+            } => Self::new_wallet(wallet_client, name, *wallet_type, *mnemonics_word_count),
             WalletCommand::List => Self::list_wallets(wallet_client),
             WalletCommand::Restore { name } => Self::restore_wallet(wallet_client, name),
             WalletCommand::RestoreBasic { name } => Self::restore_basic_wallet(wallet_client, name),
@@ -141,6 +151,7 @@ impl WalletCommand {
         wallet_client: T,
         name: &str,
         wallet_kind: WalletKind,
+        mnemonics_word_count: u32,
     ) -> Result<()> {
         let passphrase = ask_passphrase(None)?;
         let confirmed_passphrase = ask_passphrase(Some("Confirm passphrase: "))?;
@@ -151,7 +162,8 @@ impl WalletCommand {
                 "Passphrases do not match",
             ));
         }
-        let (enckey, mnemonic) = wallet_client.new_wallet(name, &passphrase, wallet_kind)?;
+        let (enckey, mnemonic) =
+            wallet_client.new_wallet(name, &passphrase, wallet_kind, Some(mnemonics_word_count))?;
 
         if let WalletKind::HD = wallet_kind {
             ask("Please store following mnemonic safely to restore your wallet later: ");
--- a/client-core/src/mnemonic.rs
+++ b/client-core/src/mnemonic.rs
@@ -9,7 +9,6 @@ use serde::{Deserialize, Serialize, Serializer};
 
 use client_common::{ErrorKind, Result, ResultExt};
 
-const MNEMONIC_TYPE: MnemonicType = MnemonicType::Words24;
 const MNEMONIC_LANGUAGE: Language = Language::English;
 
 /// Mnemonic wrapped in secures string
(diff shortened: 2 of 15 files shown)


client-cli/src/command/wallet_command.rs
```diff
@@ -35,6 +35,13 @@ pub enum WalletCommand {
             case_insensitive = true
         )]
         wallet_type: WalletKind,
+        #[structopt(
+            name = "wallet mnemonic count",
+            short = "m",
+            long = "mnemonics_word_count",
+            help = "Number of words in mnemonics"
+        )]
+        mnemonics_word_count: u32,
```
devashishdxt, line 44: Can user enter any value here (other than `12` and `24`)? If not, Maybe it's better to have an enum `MnemonicType` with variants `Long` and `Short` to represent this? Also, set the default value to `Long`.

leejw51crypto: also support
12, 15,18,21,24
because it's word count, just using value seems easier.

but using enum also make sense.
then enum value will be represented as `String`

client-core/src/mnemonic.rs
```diff
@@ -38,8 +38,19 @@ impl Mnemonic {
...
+        let flag = if 12 == mnemonics_word_count {
+            MnemonicType::Words12
+        } else if 15 == mnemonics_word_count {
+            MnemonicType::Words15
+        } else if 18 == mnemonics_word_count {
+            MnemonicType::Words18
+        } else if 21 == mnemonics_word_count {
+            MnemonicType::Words21
+        } else {
+            MNEMONIC_TYPE
```
tomtau, line 51: Not sure if it's good to silently put the default -- maybe it'll be good to at least display some info/warning message ("Invalid mnemonic word count type, using the default 24 words")?

client-core/src/mnemonic.rs
```diff
@@ -38,8 +38,19 @@ impl Mnemonic {
     /// Generate and returns mnemonic words
     #[allow(clippy::new_without_default)]
     #[inline]
-    pub fn new() -> Self {
-        let mnemonic = bip39::Mnemonic::new(MNEMONIC_TYPE, MNEMONIC_LANGUAGE);
+    pub fn new(mnemonics_word_count: u32) -> Self {
+        let flag = if 12 == mnemonics_word_count {
```
tomtau, line 42: maybe pattern match can make it a bit simpler?
```
let flag = match mnemonics_word_count {
  12 => MnemonicType::Words12,
  15 => MnemonicType::Words15,
...
  _ => ...
};

File: client-core/src/service/mock_hw_key_service.rs [hunk 4]
```diff
@@ -83,7 +83,7 @@ impl Default for MockHardwareWallet {
         let socket_address = format!("127.0.0.1:{}", server_port)
             .parse()
             .expect("invalid mock hardware wallet tcp server address");
-        //let mnemonic = Mnemonic::new();
+        //let mnemonic = Mnemonic::new(24);
```
tomtau, line 86: can be deleted?

client-core/src/wallet.rs
```diff
@@ -96,6 +96,7 @@ pub trait WalletClient: Send + Sync {
         name: &str,
         passphrase: &SecUtf8,
         wallet_kind: WalletKind,
+        mnemonics_word_count: u32,
```
tomtau, line 99: probably OK in this PR, but this argument doesn't make sense WalletKind::Basic, so perhaps a TODO comment for some API redesign/refactoring (e.g. as a part of this https://github.com/crypto-com/chain/issues/1491 ).
A bit cleaner way may be either to have separate functions `new_wallet_basic(...)` and `new_wallet_hd(...,  mnemonics_word_count: ...)` or e.g. put the parameter inside `WalletKind` enum's HD variant

integration-tests/bot/chainrpc.py
```diff
@@ -160,7 +160,7 @@ def create(self, name=DEFAULT_WALLET, type='Basic', passphrase=None):
         :param name: Name of the wallet. [defualt: Default]
         :param type: Type of the wallet. [Basic|HD] [default: Basic]
         '''
-        return self.client.call('wallet_create', [name, passphrase or get_passphrase()], type)
+        return self.client.call('wallet_create', [name, passphrase or get_passphrase()], type,24)
```
tomtau, line 163: is the argument necessary? I thought it'd automatically take 24 as default?

integration-tests/client-rpc/test/hdwallet-management.test.ts
```diff
@@ -77,10 +81,11 @@ describe("HDWallet management", () => {
 		const walletRequest = newCreateWalletRequest(walletName, DEFAULT_PASSPHRASE);
 
 		const walletCreateResult = await client.request("wallet_create", [
-			walletRequest
-			, "HD"
+			walletRequest,
+			"HD",
+			24,
```
tomtau, line 86: needed?

client-cli/src/command/wallet_command.rs
```diff
@@ -35,6 +35,13 @@ pub enum WalletCommand {
             case_insensitive = true
         )]
         wallet_type: WalletKind,
+        #[structopt(
+            name = "wallet mnemonic count",
+            short = "m",
```
tomtau, line 40: perhaps one more argument:
```
default_value = "24"
```

so the old commands still work as before?

leejw51crypto: yes, old commands is the same.
i'll add default value.

integration-tests/client-rpc/test/hdwallet-management.test.ts
leejw51crypto: yes, need to give word count

client-core/src/service/mock_hw_key_service.rs
leejw51crypto: ok

client-core/src/mnemonic.rs
leejw51crypto: that's simpler, i'll change it

client-core/src/mnemonic.rs
leejw51crypto: i'll change to
12,15,18,21,24 count only
otherwise display error message, and not continue

client-core/src/wallet.rs
leejw51crypto: i'll add TODO

integration-tests/bot/chainrpc.py
leejw51crypto: i'll process with default value

leejw51crypto:
make sense, i'll change it

leejw51crypto:
restored back integration test

leejw51crypto:
checking travis issue now

client-core/src/mnemonic.rs
```diff
@@ -191,14 +203,14 @@ mod mnemonic_tests {
 
     #[test]
     fn should_display_as_secret() {
-        let mnemonic = Mnemonic::new();
+        let mnemonic = Mnemonic::new(24).expect("get 24 wrods mnemonics");
```
tomtau, line 206:
```suggestion
        let mnemonic = Mnemonic::new(24).expect("get 24 words mnemonics");
```

leejw51crypto:
ok

PR #1754, opened 2020-06-10
Problem (WIP #1703): incorrect tree hashing in mls modules

yihuang:
Solution:
- Remove redundant optional in spec: https://github.com/mlswg/mls-protocol/pull/345
- Verify parent_hash

### DIFF
--- a/chain-tx-enclave-next/mls/src/group.rs
+++ b/chain-tx-enclave-next/mls/src/group.rs
@@ -46,7 +46,7 @@ impl GroupAux {
     fn get_sender(&self) -> Sender {
         Sender {
             sender_type: SenderType::Member,
-            sender: self.tree.my_pos as u32,
+            sender: self.tree.my_pos.0 as u32,
         }
     }
 
@@ -162,7 +162,7 @@ impl GroupAux {
         }
         Welcome {
             version: PROTOCOL_VERSION_MLS10,
-            cipher_suite: self.tree.cs.clone() as u16,
+            cipher_suite: self.tree.cs as u16,
             secrets,
             encrypted_group_info,
         }
@@ -413,13 +413,28 @@ impl GroupAux {
         );
         let group_info =
             cs.open_group_info(&welcome.encrypted_group_info, welcome_key, welcome_nonce);
-        // * "Verify the signature on the GroupInfo object..."
-        let signer = match group_info
+
+        // convert Vec<Option<Node>> to Vec<Node>
+        let nodes = group_info
             .payload
             .tree
-            .get(group_info.payload.signer_index as usize)
-        {
-            Some(Some(Node::Leaf(Some(kp)))) => Ok(kp.clone()),
+            .clone()
+            .into_iter()
+            .enumerate()
+            .map(|(i, node)| {
+                node.unwrap_or_else(|| {
+                    if i % 2 == 0 {
+                        Node::Leaf(None)
+                    } else {
+                        Node::Parent(None)
+                    }
+                })
+            })
+            .collect::<Vec<_>>();
+
+        // * "Verify the signature on the GroupInfo object..."
+        let signer = match nodes.get(group_info.payload.signer_index as usize) {
+            Some(Node::Leaf(Some(kp))) => Ok(kp.clone()),
             _ => Err(kp::Error::KeyPackageNotFound),
         }?;
         let identity_pk =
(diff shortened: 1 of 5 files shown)


chain-tx-enclave-next/mls/src/group.rs
```diff
@@ -561,7 +561,8 @@ pub struct GroupInfoPayload {
     /// that is processed)
     pub epoch: u64,
     /// 1..2^32-1
-    pub tree: Vec<Option<Node>>,
+    /// https://github.com/mlswg/mls-protocol/pull/345
+    pub tree: Vec<Node>,
```
tomtau, line 565: The PR may not be merged, so this can be reversed and put some FIXME note.
The final form is to be specified; it may end up something like `Vec<Option<NodeHash>>`

chain-tx-enclave-next/mls/src/tree.rs
```diff
@@ -162,72 +173,66 @@ pub struct Tree {
 impl Tree {
     pub fn get_package(&self, leaf_index: LeafSize) -> Option<&KeyPackage> {
         match self.nodes[NodeSize::from_leaf_index(leaf_index).0] {
-            Node::Leaf(Some(ref pk)) => Some(pk),
-            _ => None,
+            Node::Leaf(ref kp) => kp.as_ref(),
+            _ => panic!("invalid node type"),
```
tomtau, line 177: why panic? what if it's an empty leaf?

yihuang: empty leaf will match the `Node::Leaf` pattern, it'll come to here only when node index and node type don't match.

chain-tx-enclave-next/mls/src/group.rs
yihuang: Done.

yihuang:
Added copath implementation in tree-math

chain-tx-enclave-next/mls/src/tree.rs
```diff
@@ -155,79 +166,72 @@ pub struct Tree {
...
+                        nodes[left_pos.0].parent_hash(),
+                        nodes[right_pos.0].parent_hash(),
+                    ) {
+                        (None, Some(hash)) => hash,
+                        (Some(hash), None) => hash,
                         _ => {
                             return Err(kp::Error::TreeIntegrityError);
                         }
+                    };
+                    if parent_hash != node_hash(&nodes, cs, x, leaf_len) {
```
tomtau, line 233: `.ct_eq` for sanity -- normal `!=` short-circuits / has a timing-leak

chain-tx-enclave-next/mls/src/tree.rs
```diff
@@ -350,7 +328,37 @@ impl Tree {
...
+            let right_index = index.right(leaf_size).unwrap();
+            let right_hash = node_hash(nodes, cs, right_index, leaf_size);
+            let mut inp = Vec::new();
+            ParentNodeHashInput {
+                node_index: index.0 as u32,
+                parent_node: pn.clone(),
+                left_hash,
+                right_hash,
+            }
+            .encode(&mut inp);
```
tomtau, line 360: nit: `.get_encoding()` is a short-hand for this, so could be:
```
let inp = ParentNodeHashInput {
                  node_index: index.0 as u32,
                  parent_node: pn.clone(),
                  left_hash,
                  right_hash,
              }.get_encoding();
```
?

chain-tx-enclave-next/mls/src/tree_math.rs
```diff
@@ -199,23 +209,52 @@ impl NodeSize {
...
-            d
         } else {
-            vec![]
+            None
+        }
+    }
+
+    /// The copath of the node is the siblings of the nodes on its direct
+    /// path (including the node itself)
+    pub fn copath(self, leafs: LeafSize) -> Vec<NodeSize> {
```
tomtau, line 247: I guess these are not yet used in the code? but they'll be necessary for populating the paths -- perhaps add the tests with precomputed values

yihuang: done

chain-tx-enclave-next/mls/src/tree.rs
yihuang: done

chain-tx-enclave-next/mls/src/tree.rs
yihuang: done

yihuang:
It turns out `copath` is not used by tree evolution, removed.

PR #1759, opened 2020-06-11
Problem: ambiguous error message in check_genesis_hash (fix #1670)

### LINKED ISSUE #1670: Problem: ambiguous error message in check_genesis_hash
"genesis app hash from tendermint {} is not match setted genesis hash {}"
is ambiguous
it should
"genesis hash from tendermint {} is not match setted genesis hash {}"

we can summarize terms
genesis hash: hash of the genesis
app hash: hash of app_state in genesis

also check in both
client-cli, client-rpc
to display correct error message when "genesis hash" doesn't match from the tendermint

leejw51crypto:
Solution: change from genesis hash  to genesis fingerprint


leejw51crypto:
i'll change it

leejw51crypto:
change env variable

tomtau:
bors try

PR #1760, opened 2020-06-11
Problem (Fix #1703): incorrect tree hashing in mls modules

### LINKED ISSUE #1703: Problem: incorrect tree hashing in mls modules

yihuang:
Solution:
- Implement ratchet tree evolution
- Populate DirectPath for init_commit
- support apply the path secrets when apply commit message

### DIFF
--- a/chain-tx-enclave-next/mls/src/ciphersuite.rs
+++ b/chain-tx-enclave-next/mls/src/ciphersuite.rs
@@ -1,4 +1,5 @@
 use crate::group::GroupInfo;
+use crate::key::{HPKEPrivateKey, HPKEPublicKey};
 use crate::keypackage::{KeyPackage, OwnedKeyPackage};
 use crate::message::*;
 use crate::utils::{encode_vec_u32, encode_vec_u8_u8, read_vec_u32, read_vec_u8_u8};
@@ -7,11 +8,11 @@ use hkdf::{Hkdf, InvalidLength};
 use hpke::{
     aead::{AeadTag, AesGcm128},
     kex::{Marshallable, Unmarshallable},
-    EncappedKey,
+    EncappedKey, HpkeError,
 };
 use rustls::internal::msgs::codec::{Codec, Reader};
 use secrecy::{ExposeSecret, SecretVec};
-use sha2::digest::generic_array::{ArrayLength, GenericArray};
+use sha2::digest::generic_array::GenericArray;
 use sha2::digest::{BlockInput, FixedOutput, Input, Reset};
 use sha2::{Digest, Sha256};
 
@@ -77,12 +78,9 @@ impl Codec for ApplicationContext {
 }
 
 /// Additional methods for Kdf
-pub trait HkdfExt<D>
-where
-    D: Input + BlockInput + FixedOutput + Reset + Default + Clone,
-    D::BlockSize: ArrayLength<u8>,
-    D::OutputSize: ArrayLength<u8>,
-{
+///
+/// draft-ietf-mls-protocol.md#key-schedule
+pub trait HkdfExt {
     fn expand_label(
         &self,
         group_context_hash: Vec<u8>,
@@ -106,7 +104,7 @@ where
     ) -> Result<Vec<u8>, hkdf::InvalidLength>;
 }
 
-impl<D> HkdfExt<D> for Hkdf<D>
+impl<D> HkdfExt for Hkdf<D>
 where
     D: Input + BlockInput + FixedOutput + Reset + Default + Clone,
 {
@@ -305,6 +303,112 @@ impl CipherSuite {
     pub fn hash_len(self) -> usize {
         match self {
             CipherSuite::MLS10_128_DHKEMP256_AES128GCM_SHA256_P256 => 32,
+        }
+    }
+
+    /// spec: draft-ietf-mls-protocol.md#key-schedule
+    pub fn expand_label(
+        self,
+        secret: &SecretVec<u8>,
+        group_context_hash: Vec<u8>,
+        label: &str,
+        context: &[u8],
+        length: u16,
+    ) -> Result<SecretVec<u8>, InvalidLength> {
+        match self {
+            CipherSuite::MLS10_128_DHKEMP256_AES128GCM_SHA256_P256 => {
+                Hkdf::<Sha256>::new(None, secret.expose_secret())
+                    .expand_label(group_context_hash, label, context, length)
+                    .map(<SecretVec<u8>>::new)
+            }
+        }
+    }
+
+    pub fn secret_size(self) -> u16 {
+        match self {
+            CipherSuite::MLS10_128_DHKEMP256_AES128GCM_SHA256_P256 => 32,
+        }
+    }
+
+    pub fn keypair_secret_size(self) -> u16 {
+        match self {
+            CipherSuite::MLS10_128_DHKEMP256_AES128GCM_SHA256_P256 => 32,
+        }
+    }
+
+    pub fn derive_private_key(self, secret: &SecretVec<u8>) -> HPKEPrivateKey {
+        match self {
+            CipherSuite::MLS10_128_DHKEMP256_AES128GCM_SHA256_P256 => {
+                HPKEPrivateKey::derive(&secret.expose_secret())
+            }
+        }
+    }
+
+    /// encrypt to public key
+    pub fn encrypt(
+        self,
+        mut msg: Vec<u8>,
+        recip_pk: &HPKEPublicKey,
+        aad: &[u8],
+    ) -> Result<HPKECiphertext, HpkeError> {
+        match self {
+            CipherSuite::MLS10_128_DHKEMP256_AES128GCM_SHA256_P256 => {
+                let mut csprng = rand::thread_rng();
+                let (kem_output, mut context) = hpke::setup_sender::<
+                    AesGcm128,
+                    hpke::kdf::HkdfSha256,
+                    hpke::kem::DhP256HkdfSha256,
+                    _,
+                >(
+                    &hpke::OpModeS::Base,
+                    recip_pk.kex_pubkey(),
+                    b"",
+                    &mut csprng,
+                )?;
+                let tag = context.seal(&mut msg, aad)?;
+                msg.extend_from_slice(&tag.marshal());
+                Ok(HPKECiphertext {
+                    kem_output: kem_output.marshal().to_vec(),
+                    ciphertext: msg,
+                })
+            }
+        }
+    }
+
+    /// decrypt ciphertext
+    pub fn decrypt(
+        self,
+        private_key: &HPKEPrivateKey,
+        aad: &[u8],
+        ct: &HPKECiphertext,
+    ) -> Result<Vec<u8>, HpkeError> {
+        match self {
+            CipherSuite::MLS10_128_DHKEMP256_AES128GCM_SHA256_P256 => {
+                let encapped_key = EncappedKey::<
+                    <hpke::kem::DhP256HkdfSha256 as hpke::kem::Kem>::Kex,
+                >::unmarshal(&ct.kem_output)
+                .expect("valid encapped key");
+                let mut context = hpke::setup_receiver::<
+                    AesGcm128,
+                    hpke::kdf::HkdfSha256,
+                    hpke::kem::DhP256HkdfSha256,
+                >(
+                    &hpke::OpModeR::Base,
+                    private_key.kex_secret(),
+                    &encapped_key,
+                    b"",
+                )?;
+
+                let payload_len = ct.ciphertext.len();
+                let mut payload = ct.ciphertext[0..payload_len - 16].to_vec();
+                let tag_bytes = &ct.ciphertext[payload_len - 16..payload_len];
+                let tag = AeadTag::<AesGcm128>::unmarshal(tag_bytes).expect("valid tag");
+
+                context
+                    .open(&mut payload, aad, &tag)
+                    .expect("decryption failed");
+                Ok(payload)
+            }
         }
     }
 }
--- a/chain-tx-enclave-next/mls/src/group.rs
+++ b/chain-tx-enclave-next/mls/src/group.rs
@@ -175,19 +175,41 @@ impl GroupAux {
             .collect();
         let mut updated_tree = self.tree.clone();
         let positions = updated_tree.update(&add_proposals, &[], &[]);
-        let commit_secret = vec![0; self.tree.cs.hash_len()];
+
+        // generate init key
+        self.owned_kp.update_init_key();
+
+        // update path secrets
+        let (path_nodes, parent_hash, commit_secret) = updated_tree.evolve(
+            &self.context.get_encoding(),
+            self.owned_kp.init_private_key.marshal(),
+        );
+
+        // update keypackage's parent_hash extension
+        self.owned_kp
+            .keypackage
+            .payload
+            .put_extension(&ext::ParentHashExt(parent_hash));
+        self.owned_kp.update_signature();
+        updated_tree.set_my_package(self.owned_kp.keypackage.clone());
+
         let commit = Commit {
             updates: vec![],
             removes: vec![],
             adds: add_proposals_ids,
-            path: None,
+            path: Some(DirectPath {
+                leaf_key_package: self.owned_kp.keypackage.clone(),
+                nodes: path_nodes,
+            }),
         };
         let updated_epoch = self.context.epoch + 1;
         let confirmed_transcript_hash = self.get_init_confirmed_transcript_hash(&commit);
-        let mut updated_group_context = self.context.clone();
-        updated_group_context.epoch = updated_epoch;
-        updated_group_context.tree_hash = updated_tree.compute_tree_hash();
-        updated_group_context.confirmed_transcript_hash = confirmed_transcript_hash;
+        let updated_group_context = GroupContext {
+            tree_hash: updated_tree.compute_tree_hash(),
+            epoch: updated_epoch,
+            confirmed_transcript_hash,
+            ..self.context.clone()
+        };
         let updated_group_context_hash = self.tree.cs.hash(&updated_group_context.get_encoding());
         let epoch_secrets = self.secrets.generate_new_epoch_secrets(
             &commit_secret,
@@ -234,16 +256,16 @@ impl GroupAux {
         remove_proposals: &[MLSPlaintext],
         ra_verifier: &impl AttestedCertVerifier,
         now: Timespec,
-    ) -> Result<(), kp::Error> {
+    ) -> Result<(), ProcessCommitError> {
         // "Verify that the epoch field of the enclosing MLSPlaintext message
         // is equal to the epoch field of the current GroupContext object"
         if self.context.epoch != commit.content.epoch {
-            return Err(kp::Error::GroupEpochError);
+            return Err(ProcessCommitError::GroupEpochError);
         }
         let kp = self
             .tree
             .get_package(LeafSize(commit.content.sender.sender as usize))
-            .ok_or(kp::Error::KeyPackageNotFound)?;
+            .ok_or(ProcessCommitError::KeyPackageNotFound)?;
         let pk = IdentityPublicKey::new_unsafe(kp.verify(ra_verifier, now)?.public_key);
         // "Verify that the signature on the MLSPlaintext message verifies
         //  using the public key from the credential stored at the leaf in the tree indicated by the sender field."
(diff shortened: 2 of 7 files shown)


chain-tx-enclave-next/mls/src/ciphersuite.rs
```diff
@@ -77,12 +78,7 @@ impl Codec for ApplicationContext {
 }
 
 /// Additional methods for Kdf
-pub trait HkdfExt<D>
-where
-    D: Input + BlockInput + FixedOutput + Reset + Default + Clone,
-    D::BlockSize: ArrayLength<u8>,
-    D::OutputSize: ArrayLength<u8>,
-{
+pub trait HkdfExt {
```
leejw51crypto, line 81: some comment about HkdftExt , can be replaced with fullword ?

chain-tx-enclave-next/mls/src/ciphersuite.rs
```diff
@@ -296,15 +292,88 @@ impl CipherSuite {
...
+    ) -> Result<Vec<u8>, InvalidLength> {
+        match self {
+            CipherSuite::MLS10_128_DHKEMP256_AES128GCM_SHA256_P256 => Hkdf::<Sha256>::new(
+                None, secret,
+            )
+            .expand_label(group_context_hash, label, context, length),
+        }
+    }
+
+    pub fn secret_size(self) -> u16 {
```
tomtau, line 324: key_pair secret?

chain-tx-enclave-next/mls/src/ciphersuite.rs
```diff
@@ -153,13 +149,13 @@ where
 }
 
 impl CipherSuite {
-    pub fn aead_key_len(&self) -> usize {
+    pub fn aead_key_len(self) -> usize {
```
leejw51crypto, line 152: comment of aead

chain-tx-enclave-next/mls/src/ciphersuite.rs
```diff
@@ -296,15 +292,88 @@ impl CipherSuite {
...
+        }
+    }
+
+    pub fn secret_size(self) -> u16 {
+        match self {
+            CipherSuite::MLS10_128_DHKEMP256_AES128GCM_SHA256_P256 => 32,
+        }
+    }
+
+    pub fn derive_private_key(
```
tomtau, line 330: from Ciphersuites def?

chain-tx-enclave-next/mls/src/ciphersuite.rs
```diff
@@ -278,7 +274,7 @@ impl CipherSuite {
 
     /// TODO: use generic array?
     pub fn encrypt_group_info(
-        &self,
+        self,
```
leejw51crypto, line 277: just &[u8] is fastest, not necessary to use generic array

chain-tx-enclave-next/mls/src/ciphersuite.rs
```diff
@@ -296,15 +292,88 @@ impl CipherSuite {
     }
 
     /// TODO: use generic array?
-    pub fn hash(&self, data: &[u8]) -> Vec<u8> {
+    pub fn hash(self, data: &[u8]) -> Vec<u8> {
```
leejw51crypto, line 295: this is fine, not necessary to use generic array

chain-tx-enclave-next/mls/src/group.rs
```diff
@@ -174,19 +175,38 @@ impl GroupAux {
...
+        let (path_nodes, parent_hash, commit_secret) =
+            updated_tree.encap(&updated_group_context.get_encoding(), leaf_secret);
+
+        // update keypackage's parent_hash extension
+        self.owned_kp
+            .keypackage
+            .payload
+            .put_extension(&ext::ParentHashExt(parent_hash));
+        self.owned_kp.update_signature();
+        updated_tree.set_my_package(self.owned_kp.keypackage.clone());
```
tomtau, line 196: > The generator of the Commit starts by using the HPKE secret key "leaf_hpke_secret" associated with the new leaf KeyPackage (see {{key-packages}}) to compute "path_secret[0]"

chain-tx-enclave-next/mls/src/key.rs
```diff
@@ -100,6 +104,14 @@ impl HPKEPrivateKey {
     pub fn kex_secret(&self) -> &<hpke::kex::DhP256 as hpke::KeyExchange>::PrivateKey {
         &self.0
     }
+
+    pub fn unmarshal(secret: &[u8]) -> Result<Self, HpkeError> {
```
leejw51crypto, line 108: consider `decode` for unmarshal, because we're using `encode`, `decode` terms

chain-tx-enclave-next/mls/src/ciphersuite.rs
tomtau: probably fixme or change to use https://github.com/cfrg/draft-irtf-cfrg-hpke/blob/master/draft-irtf-cfrg-hpke.md#derivekeypair-derive-key-pair

chain-tx-enclave-next/mls/src/keypackage.rs
```diff
@@ -81,6 +81,19 @@ impl KeyPackagePayload {
         ))
     }
 
+    /// insert or update extension
+    pub fn put_extension<T: MLSExtension>(&mut self, ext: &T) {
```
leejw51crypto, line 85: pub fn put_extension(&mut self, ext: &MLSExtension) { 
is simpler

chain-tx-enclave-next/mls/src/tree.rs
```diff
@@ -350,7 +390,148 @@ impl Tree {
...
+            my_pos: LeafSize(0),
         })
     }
+
+    /// spec: draft-ietf-mls-protocol.md#ratchet-tree-evolution
+    ///
+    /// update path secrets
+    ///
+    /// returns: direct path nodes, parent_hash, root_secret
+    pub fn encap(
```
leejw51crypto, line 402: full word for `encap`

chain-tx-enclave-next/mls/src/tree.rs
```diff
@@ -350,7 +390,148 @@ impl Tree {
...
+    match node {
+        Node::Leaf(kp) => cs.hash(
+            &LeafNodeHashInput {
+                node_index: index.0 as u32,
+                key_package: kp.clone(),
+            }
+            .get_encoding(),
+        ),
+        Node::Parent(pn) => {
+            let left_index = index.left().unwrap();
```
leejw51crypto, line 492: expect()?

chain-tx-enclave-next/mls/src/ciphersuite.rs
tomtau: the problem here is that not all secrets from that "key pair expand label" may be valid, so the unmarshal may panic

chain-tx-enclave-next/mls/src/key.rs
```diff
@@ -88,6 +91,7 @@ impl Codec for HPKEPublicKey {
 
 /// p-256 private key
 /// used for obtaining the initial sealed secrets (HPKE)
+#[derive(Clone)]
 pub struct HPKEPrivateKey(<hpke::kex::DhP256 as hpke::KeyExchange>::PrivateKey);
```
tomtau, line 95: is this needed to be clone?

chain-tx-enclave-next/mls/src/group.rs
tomtau: init key is to be updated -- https://github.com/cisco/go-mls/blob/master/state.go#L370 ?

chain-tx-enclave-next/mls/src/tree.rs
```diff
@@ -350,7 +390,148 @@ impl Tree {
...
+
+    /// spec: draft-ietf-mls-protocol.md#ratchet-tree-evolution
+    ///
+    /// update path secrets
+    ///
+    /// returns: direct path nodes, parent_hash, root_secret
+    pub fn encap(
+        &mut self,
+        ctx: &[u8],
+        leaf_secret: Vec<u8>,
```
tomtau, line 405: SecretVec?

chain-tx-enclave-next/mls/Cargo.toml
```diff
@@ -19,6 +19,7 @@ aead = "0.2.0"
 rand = "0.7"
 ra-client = { path = "../enclave-ra/ra-client" }
 subtle = "2.2.3"
+rand_core = "0.5"
```
tomtau, line 22: why rand_core? isn't that exposed from rand?

chain-tx-enclave-next/mls/src/tree.rs
tomtau: or something different -- `Encap` has already a defined meaning in HPKE, but here it's something else

chain-tx-enclave-next/mls/src/key.rs
tomtau: these are keywords used in the spec

chain-tx-enclave-next/mls/src/ciphersuite.rs
tomtau: generic array would prevent some problems at compile-time, because there are various `Vec<u8>` (hashes, secrets, nonces, pubkeys...) which can be quite error-prone

chain-tx-enclave-next/mls/src/ciphersuite.rs
tomtau: generic array is meant rather for welcome_key, welcome_nonce ... not the (variable-length) plaintext input

chain-tx-enclave-next/mls/src/ciphersuite.rs
tomtau: it has a comment "Additional methods for Kdf".
It expands the "Hkdf" trait with extra methods defined in: https://github.com/mlswg/mls-protocol/blob/master/draft-ietf-mls-protocol.md#key-schedule

tomtau: maybe it can reference it?

chain-tx-enclave-next/mls/src/group.rs
tomtau: > When used as InitKeys, KeyPackages are intended to be used only once and SHOULD NOT be reused except in case of last resort.

chain-tx-enclave-next/mls/src/keypackage.rs
devashishdxt: Maybe use `pub fn put_extension(&mut self: ext: &impl MLSExtension)`?

yihuang: done

chain-tx-enclave-next/mls/src/tree.rs
yihuang: maybe "evolve"

chain-tx-enclave-next/mls/src/tree.rs
yihuang: done

chain-tx-enclave-next/mls/src/ciphersuite.rs
yihuang: The [go-mls](https://github.com/cisco/go-mls/blob/master/crypto.go#L69) used the name `SecretSize`, so

chain-tx-enclave-next/mls/src/ciphersuite.rs
yihuang: added a FIXME for now, the go version seems don't handle this currently.

chain-tx-enclave-next/mls/src/group.rs
yihuang: done

chain-tx-enclave-next/mls/src/key.rs
yihuang: removed

chain-tx-enclave-next/mls/src/tree.rs
yihuang: done

chain-tx-enclave-next/mls/Cargo.toml
yihuang: removed

chain-tx-enclave-next/mls/src/ciphersuite.rs
tomtau: so what's this `secret_size` if not key pair secret size? the output hash length?

yihuang: It’s used by expand label “path”, when generating path secrets.

yihuang:
support apply path secrets when apply commit message now, existing unit tests passed.

chain-tx-enclave-next/mls/src/ciphersuite.rs
yihuang: done

chain-tx-enclave-next/mls/src/key.rs
yihuang: The `Clone` is added back, because we need to store private keys inside tree, and tree need to be able to be cloned.

chain-tx-enclave-next/mls/src/ciphersuite.rs
tomtau: https://github.com/mlswg/mls-protocol/blob/master/draft-ietf-mls-protocol.md#ratchet-tree-evolution

`KEM.Nsk` -- Nsk is the length of the secret key, so should be the same `keypair_secret_size`?

tomtau: some of this changed, it may be good to put Git commit sha of spec this refers to

chain-tx-enclave-next/mls/src/ciphersuite.rs
tomtau: that definitions isn't longer in the spec?

tomtau: that expand_label with "key pair"?

chain-tx-enclave-next/mls/src/ciphersuite.rs
yihuang: "Each ciphersuite has a `Derive-Key-Pair` function that maps octet strings of
length `Nsk` (a KEM-specific constant defined by HPKE) to HPKE key pairs", it seems they are still different?
And in go-mls, the [P521_AES256GCM_SHA512_P521](https://github.com/cisco/go-mls/blob/master/crypto.go#L108) has SecretSize 64, but the [DHKEM_P256](https://github.com/cisco/go-mls/blob/master/crypto.go#L292) has keyPairSecretSize 32.

yihuang: > some of this changed, it may be good to put Git commit sha of spec this refers to

update the sha in lib.rs to refer to current master?

tomtau: > "Each ciphersuite has a `Derive-Key-Pair` function that maps octet strings of
> length `Nsk` (a KEM-specific constant defined by HPKE) to HPKE key pairs", it seems they are still different?
> And in go-mls, the [P521_AES256GCM_SHA512_P521](https://github.com/cisco/go-mls/blob/master/crypto.go#L108) has SecretSize 64, but the [HPKEKEM](https://github.com/cisco/go-mls/blob/master/crypto.go#L292) has keyPairSecretSize 32.

66, not 32: https://github.com/cfrg/draft-irtf-cfrg-hpke/blob/master/draft-irtf-cfrg-hpke.md#key-encapsulation-mechanisms-kems-kem-ids

`shared_secret_size`? I'd need to read through more in detail whether that context is correct

tomtau: as HPKE library doesn't implement P-521 or X448 yet, I'm not sure if it even distinguishes these sizes :) but still good to keep it consistent and distinguish them if that's in the spec

chain-tx-enclave-next/mls/src/ciphersuite.rs
yihuang: "This function is defined to be HPKE's `DeriveKeyPair` function.", I think it refers to the [hpke spec](https://github.com/cfrg/draft-irtf-cfrg-hpke/blob/master/draft-irtf-cfrg-hpke.md#derivekeypair-derive-key-pair) now.

yihuang: The DeriveKeyPair function should ideally implemented in hpke library?

yihuang: the hpke library just implemented derive_keypair, we changed to use that now.

chain-tx-enclave-next/mls/src/ciphersuite.rs
yihuang: oh, I guess they have a bug here to [use DHKEM_P256 in ciphersuite P521_AES256GCM_SHA512_P521](https://github.com/cisco/go-mls/blob/master/crypto.go#L109)

yihuang:
> maybe splitting the tree into private and public portions similarly to https://github.com/cisco/go-mls/pull/71/files#diff-53faa623cefecaf31c5a72febc6526c9R363 ?
> need to yet read through the code in more detail, as some spec parts changed a bit

how about do this in another pr, I think we need better code change to make the tree implementation more robust [sketched here](https://gist.github.com/yihuang/3155363033f6b7c4c17f4b9d42894df2)

tomtau:
> > maybe splitting the tree into private and public portions similarly to https://github.com/cisco/go-mls/pull/71/files#diff-53faa623cefecaf31c5a72febc6526c9R363 ?
> > need to yet read through the code in more detail, as some spec parts changed a bit
> 
> how about do this in another pr, I think we need more code changes to make the tree implementation more robust [sketched here](https://gist.github.com/yihuang/3155363033f6b7c4c17f4b9d42894df2)

should be ok, I'll try to do a final read of this PR later today -- these other code improvements can be in a different PR

chain-tx-enclave-next/mls/src/tree.rs
```diff
@@ -335,34 +441,264 @@ impl Tree {
...
+                right_hash,
+            }
+            .get_encoding()
         }
+    };
+    cs.hash(&payload)
+}
+
+/// no blank nodes return
+fn resolve(nodes: &[Node], index: NodeSize) -> Vec<NodeSize> {
```
tomtau, line 677: could be good to link "Ratchet Tree Nodes" section -- I was trying to search "resolve" where it is defined

chain-tx-enclave-next/mls/src/tree.rs
```diff
@@ -335,34 +441,264 @@ impl Tree {
...
+                        .expect("decrypt failed");
+                    return (ancestor, SecretVec::new(secret));
+                }
+            }
+            panic!("don't find private key to decrypt secret");
+        }
+    }
+
+    /// returns (leaf_parent_hash, commit_secret)
+    pub fn apply_path_secrets(
```
tomtau, line 586: "Synchronizing Views of the Tree"?

chain-tx-enclave-next/mls/src/tree.rs
```diff
@@ -335,34 +441,264 @@ impl Tree {
...
+        for node_index in overlap.direct_path(leaf_len).iter() {
+            last_secret = self
+                .cs
+                .expand_label(&last_secret, vec![], "path", &[], self.cs.secret_size())
+                .expect("invalid length");
+            self.nodes[node_index.0]
+                .parent_node_mut()
+                .expect("implant invalid node")
+                .set_secret(self.cs, &last_secret);
+        }
```
tomtau, line 622: A bit confused about this part -- my assumption is, as this is called by the recipient in processing commit in group init, that this is "Synchronizing Views of the Tree".
There's one step:
> The recipient SHOULD verify that the received public keys agree with the public keys derived from the new path_secret values.

I'm more sure whether that's done?

chain-tx-enclave-next/mls/src/tree.rs
yihuang: done

chain-tx-enclave-next/mls/src/tree.rs
yihuang: done

chain-tx-enclave-next/mls/src/tree.rs
yihuang: Changed the `set_secret` to `set_secret_verify_public`, the verification is done inside. the public key is updated in `merge_public_key`.

yihuang:
> looks ok -- there may be one step missing or I missed that when reading the code -- in general, it'd be good to add pointers to spec in the more involved parts

Fixed, also changed the error types to be more specific.

yihuang:
> one small clippy fix: https://travis-ci.org/github/crypto-com/chain/jobs/699644837#L1837

done

tomtau:
bors retry

PR #1772, opened 2020-06-15
Problem: reading genesis time from tendermint make `dev init` fail(fi…

leejw51crypto:
## why , i make this pr
Genesis time is UTC time , and should be later then keypackage time.
also if it's too distant from current utc time, tendermint will not work correctly.

## how i solved 
1. overwrite genesis time with current utc time,
can be disabled by `--no_genesistime_overwrite` option

2. add evidence parameters to dev.json

### DIFF
--- a/dev-utils/example-dev-conf.json
+++ b/dev-utils/example-dev-conf.json
@@ -24,6 +24,10 @@
         "base_fee": "1.1",
         "per_byte_fee": "1.25"
     },
+    "evidence": {
+        "max_age_duration": "172800000000000",
+        "max_age_num_blocks": "100000"
+    },
     "council_nodes": {
         "0x3ae55c16800dc4bd0e3397a9d7806fb1f11639de": [
             "test",
--- a/ci-scripts/check-docker-app-hash.sh
+++ b/ci-scripts/check-docker-app-hash.sh
@@ -15,7 +15,7 @@ function check_command_exist() {
 
 check_command_exist "jq"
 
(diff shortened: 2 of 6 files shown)


leejw51crypto:
checking travis

dev-utils/example-dev-conf.json
```diff
@@ -33,8 +37,8 @@
                 "value": "pPxJH5wUamXcCvSxgywzmANuO0UNR5x3nCbUzsrCeX8="
             },
             {
-                "keypackage": "RklYTUU="
+                "keypackage": "generate keypackage. for e.g.) `ra-sp-server --quote-type Unlinkable --ias-key $IAS_API_KEY --spid $SPID &`,`dev-utils keypackage generate --path ./mls.sgxs --output key.txt` "
```
tomtau, line 40: there's that `test_dev_config_file` unit test that tries to make sure the "example" file is somewhat valid -- so you may need to include some example keypackage, or at least some base64 payload.

leejw51crypto: ok, i'll replace with valid one

leejw51crypto:
ok, i'll add an option `not overwrite time`

tomtau:
> ok, i'll add an option `not overwrite time`

you may need to update the CI script as well, or set this option default to "true"

leejw51crypto:
ok, i'll check drone now

leejw51crypto:
added evidence option, checked with local drone

PR #1773, opened 2020-06-15
Problem:(fix #issue-1710) client-rpc can't initialize without tendermint rpc

linfeng-crypto:
solution:
* do not connect tendermint when `new`  a tendermint-client
* add mock-enclave feature for client-common
* impl FeeAlgorithm and TransactionObfuscation for tendermint-client
* using tendermint-client to create AppWalletClient and AppOpsClient instead of using fee_policy and obfuscator which need to connect tendermint
* move cipher from client-core to client-common


client-common/src/tendermint/rpc_client/sync_rpc_client.rs
```diff
@@ -31,32 +47,101 @@ const RESPONSE_TIMEOUT: Duration = Duration::from_secs(10);
 pub struct SyncRpcClient {
     runtime: Arc<Mutex<Runtime>>,
     /// ASYNC RPC CLIENT
-    pub async_rpc_client: AsyncRpcClient,
+    pub async_rpc_client: Arc<Mutex<Option<AsyncRpcClient>>>,
+    url: String,
+}
+
+impl GenesisExt for SyncRpcClient {
```
yihuang, line 54: Is this necessary?

client-common/src/tendermint/rpc_client/sync_rpc_client.rs
```diff
@@ -31,32 +47,101 @@ const RESPONSE_TIMEOUT: Duration = Duration::from_secs(10);
...
+        static OBFUCATION: OnceCell<std::result::Result<AppTransactionObfuscation, (ErrorKind, String)>> =
+            OnceCell::new();
+        let obfucation = OBFUCATION.get_or_init(|| {
+            AppTransactionObfuscation::from_tx_query(self).map_err(|e| (e.kind(), e.message().into()))
+        });
+        let obfucation = match obfucation {
+            Ok(o) => Ok(o.clone()),
+            Err(e) => Err(e.clone()),
+        };
+        obfucation
```
yihuang, line 123:
```suggestion
        obfucation.clone()
```

client-common/src/tendermint/rpc_client/sync_rpc_client.rs
```diff
@@ -31,32 +47,101 @@ const RESPONSE_TIMEOUT: Duration = Duration::from_secs(10);
...
+    fn calculate_fee(&self, num_bytes: usize) -> std::result::Result<Fee, CoinError> {
+        self.fee_policy().calculate_fee(num_bytes)
+    }
+
+    fn calculate_for_txaux(&self, txaux: &TxAux) -> std::result::Result<Fee, CoinError> {
+        self.fee_policy().calculate_for_txaux(txaux)
+    }
+}
+
+impl TransactionObfuscation for SyncRpcClient {
```
yihuang, line 77: I think this impl can define in client-core, so no need to move cipher from client-core to client-common.

linfeng-crypto: `TransactionObfuscation` defined in cipher, but `client-core` needs the `client-common`, so I can't use `client-core` in `client-common`,  if  I do this in client-core, I need to move `SyncRpcClient` to `client-core`

client-common/src/tendermint/rpc_client/sync_rpc_client.rs
linfeng-crypto: I'll remove

client-common/src/tendermint/rpc_client/sync_rpc_client.rs
linfeng-crypto: yeah, I'll try

linfeng-crypto: the `DefaultTransactionObfuscation` and `MockTransactionObfuscation` also located in `client-core/cipher`, but the `client-common` can not use the `client-core`,  so I have to move the cipher to `client-common`

tomtau:
bors try

client-common/src/tendermint/rpc_client/sync_rpc_client.rs
yihuang: I mean you can do this in `client-core`:
```
use client_common::... SyncRpcClient;

impl TransactionObfuscation for SyncRpcClient {
}
```

Because you only need either the trait or the type in current crate to write the impl I think.

linfeng-crypto: yeah,   I can impl the trait in the client-core, but the client-common still need the client-core if the client-common wants to use the functions in the trait.

client-common/src/cipher/AttestationReportSigningCACert.pem
```diff
@@ -0,0 +1,31 @@
+-----BEGIN CERTIFICATE-----
```
leejw51crypto, line 1: is this certificate necessary?
can user make it?

linfeng-crypto:
> edp-obfuscation

I'll check

client-common/src/cipher/AttestationReportSigningCACert.pem
linfeng-crypto: I don't think so, it hard coded in the ra-client.

client-common/src/tendermint/rpc_client/sync_rpc_client.rs
yihuang: ok, but you can move the code, rather than copy, right? I think client-core don't need them anymore.

client-common/src/tendermint/rpc_client/sync_rpc_client.rs
```diff
@@ -31,32 +47,98 @@ const RESPONSE_TIMEOUT: Duration = Duration::from_secs(10);
...
+                .fee_policy()
+        });
+        *policy
+    }
+
+    /// get the obfuscation from tx query
+    pub fn get_tx_query(
+        &self,
+    ) -> std::result::Result<impl TransactionObfuscation, (ErrorKind, String)> {
+        static OBFUCATION: OnceCell<
```
yihuang, line 113:
```suggestion
        static OBFUSCATION: OnceCell<
```

client-common/src/tendermint/rpc_client/sync_rpc_client.rs
```diff
@@ -31,32 +47,98 @@ const RESPONSE_TIMEOUT: Duration = Duration::from_secs(10);
...
+            std::result::Result<AppTransactionObfuscation, (ErrorKind, String)>,
+        > = OnceCell::new();
+        let obfucation = OBFUCATION.get_or_init(|| {
```
yihuang, line 116:
```suggestion
        let obfuscation = OBFUSCATION.get_or_init(|| {
```

client-common/src/tendermint/rpc_client/sync_rpc_client.rs
yihuang: I still think it would be better to not "intrude" into the websocket client implementation itself.
maybe wrap it up with something like this in client-core:
```
struct LazyRpcClient {
    pub rpc_client: Arc<Mutex<Option<SyncRpcClient>>>,
    url: String,
}
```

tomtau:
bors try

client-rpc/src/handler.rs
```diff
@@ -80,8 +72,12 @@ impl RpcHandler {
             StakingRpcImpl::new(wallet_client.clone(), ops_client.clone(), network_id);
         let info_rpc = InfoRpcImpl::new(ops_client);
 
-        let sync_wallet_client =
-            make_wallet_client(storage, tendermint_client, fee_policy, obfuscator)?;
+        let sync_wallet_client = make_wallet_client(
+            storage,
+            tendermint_client.clone(),
+            tendermint_client.clone(),
+            tendermint_client,
```
yihuang, line 79: maybe combine these parameters?

cro-clib/src/jsonrpc.rs
```diff
@@ -264,7 +264,7 @@ unsafe fn create_rpc(
     network_id: u8,
     progress_callback: Option<ProgressCallback>,
     user_data: *const std::ffi::c_void,
-    mock: bool,
+    _mock: bool,
```
yihuang, line 267: remove this parameter, I guess mock mode is controlled by compile feature. I guess you need to add feature "mock-enclave" to cro-clib too.

linfeng-crypto: yeah, then the interface aguments will be changed and other applications using the cro-clib will also need to be fixed, it's another issue i think, so I keep this.

client-common/src/tendermint/rpc_client/sync_rpc_client.rs
linfeng-crypto: Ok, I'll fix like this.

tomtau:
bors try

client-common/src/tendermint/rpc_client/sync_rpc_client.rs
linfeng-crypto: Actually, we can use `LazyRpcClient` as `fee_policy` and `obfuscation`, but we also need the websocket client(which impled the `Client` trait) to initialize the `WalletClient`, so the websocket client should be also `lazy`. Or I have to impl the `Client` trait on the `LazyRpcClient` and use it to create the `WalletClient`, but I don't think it makes sense to create two types of client.

PR #1795, opened 2020-06-21
Problem (Fix #1787 #1788): no update/remove proposal processing in mls module

### LINKED ISSUE #1787: Problem: no update proposal processing in mls

yihuang:
Solution:
- process update/remove proposal and add unit test

### DIFF
--- a/chain-tx-enclave-next/mls/src/group.rs
+++ b/chain-tx-enclave-next/mls/src/group.rs
@@ -1,9 +1,11 @@
+use std::convert::TryInto;
+
 use crate::ciphersuite::*;
 use crate::extensions as ext;
 use crate::key::IdentityPublicKey;
 use crate::keypackage::Timespec;
 use crate::keypackage::PROTOCOL_VERSION_MLS10;
-use crate::keypackage::{self as kp, KeyPackage, OwnedKeyPackage};
+use crate::keypackage::{self as kp, KeyPackage, KeyPackageSecret};
 use crate::message::*;
 use crate::secrets::*;
 use crate::tree::*;
@@ -16,7 +18,6 @@ use ra_client::AttestedCertVerifier;
 use rustls::internal::msgs::codec::{self, Codec, Reader};
 use secrecy::{ExposeSecret, SecretVec};
 use sha2::Sha256;
-use std::collections::BTreeMap;
 use std::collections::BTreeSet;
 use subtle::ConstantTimeEq;
 
@@ -25,11 +26,13 @@ pub struct GroupAux {
     pub context: GroupContext,
     pub tree: Tree,
     pub secrets: EpochSecrets<Sha256>,
-    pub owned_kp: OwnedKeyPackage,
+    pub kp_secret: KeyPackageSecret,
+    // record the new secret when waiting for self update commit
+    pub kp_secret_pending: Option<KeyPackageSecret>,
 }
 
 impl GroupAux {
-    fn new(context: GroupContext, tree: Tree, owned_kp: OwnedKeyPackage) -> Self {
+    fn new(context: GroupContext, tree: Tree, kp_secret: KeyPackageSecret) -> Self {
         let secrets: EpochSecrets<Sha256> = match &tree.cs {
             CipherSuite::MLS10_128_DHKEMP256_AES128GCM_SHA256_P256 => {
                 EpochSecrets::new(tree.cs.hash(&context.get_encoding()), tree.leaf_len())
@@ -39,7 +42,8 @@ impl GroupAux {
             context,
             tree,
             secrets,
-            owned_kp,
+            kp_secret,
+            kp_secret_pending: None,
         }
     }
 
@@ -50,7 +54,10 @@ impl GroupAux {
         }
     }
 
-    fn get_signed_add(&self, kp: &KeyPackage) -> MLSPlaintext {
+    /// Generate and sign add proposal
+    ///
+    /// TODO return `None` when nodes size overflow u32?
+    pub fn get_signed_add(&self, kp: &KeyPackage) -> MLSPlaintext {
         let sender = self.get_sender();
         let add_content = MLSPlaintextCommon {
             group_id: self.context.group_id.clone(),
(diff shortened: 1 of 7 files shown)


chain-tx-enclave-next/mls/src/group.rs
```diff
@@ -173,8 +173,12 @@ impl GroupAux {
             .iter()
             .map(|plain| ProposalId(self.tree.cs.hash(&plain.get_encoding())))
             .collect();
+        let adds = add_proposals
```
leejw51crypto, line 176: additions?

chain-tx-enclave-next/mls/src/group.rs
```diff
@@ -298,16 +304,37 @@ impl GroupAux {
...
-                    .cloned()
+                    .and_then(|add| add.get_add().cloned())
+                    .ok_or(ProcessCommitError::CommitError)
+            })
+            .collect::<Result<Vec<_>, _>>()?;
+        let updates = commit_content
+            .updates
+            .iter()
+            .map(|proposal_id| {
+                update_proposals_ids
```
leejw51crypto, line 315: how about separate this block for readibility?

chain-tx-enclave-next/mls/src/group.rs
```diff
@@ -838,4 +865,43 @@ mod test {
         // they should get to the same context
         assert_eq!(&added_group.context, &creator_group.context);
     }
+
+    #[test]
+    fn test_tree_update() {
+        let creator_kp = get_fake_keypackage();
```
leejw51crypto, line 871: creator_keypackage?

chain-tx-enclave-next/mls/src/tree.rs
```diff
@@ -386,18 +401,27 @@ impl Tree {
...
-        add_proposals: &[MLSPlaintext],
-        update_proposals: &[MLSPlaintext],
-        remove_proposals: &[MLSPlaintext],
+        adds: &[Add],
+        updates: &[(LeafSize, Update)],
+        removes: &[Remove],
     ) -> Vec<(NodeSize, KeyPackage)> {
-        for _update in update_proposals.iter() {
-            // FIXME
+        // spec: draft-ietf-mls-protocol.md#update
```
leejw51crypto, line 408: what's ietf?

tomtau: https://github.com/mlswg/mls-protocol/blob/master/draft-ietf-mls-protocol.md
https://www.ietf.org/about/who/

yihuang:
> the original `process_commit` was written for initial commit where it's processed by the group creator, while the other members retrieve the state from the welcome message.
> it is thus dangerous in its current form for general commit processing -- e.g. it takes proposals as `&[MLSPlaintext]`, but doesn't validate them (e.g. signed by valid member? we may want to impose further restrictions that update can only be proposed by that particular member being updated) ? this needs to be cleared up in comments and possibly types of what one should/can pass in; the function may potentially be split it up for initial commit and later commits (the internal parts may be shared in some internal functions).
> the test only tests the direct updates of the tree, but there's no test showing the intended commit processing workflows

Made the commit generate and process more realistic situation, the flow is showed in unit test `test_group_update`.
The rules around who gets to sign what messages is left to be decided by code outside of mls library.

chain-tx-enclave-next/mls/src/group.rs
```diff
@@ -168,42 +217,76 @@ impl GroupAux {
...
+        }
         let mut updated_tree = self.tree.clone();
-        let positions = updated_tree.update(&add_proposals, &[], &[]);
+        let positions = updated_tree.update(&additions, &updates, &removes);
 
-        // generate init key
-        self.owned_kp.update_init_key();
+        // pathRequired = haveUpdateOrRemove || haveNoProposalsAtAll
+        let should_populate_path =
+            proposals.is_empty() || !updates.is_empty() || !removes.is_empty();
```
tomtau, line 251: the fact that the path must be populated in these cases does not imply that in the other cases, it must not be populated.
In particular for the group initialisation / at genesis, it must be populated, as there's no "init_secret" (so no entropy from there, as one just passes empty or 0 vec) and we don't use PSK or something like that.

so either dedicated internal function `init_commit` or some extra parameter to this one?

chain-tx-enclave-next/mls/src/group.rs
```diff
@@ -262,103 +358,47 @@ impl GroupAux {
...
-        };
-        updated_group_context.tree_hash = updated_tree.compute_tree_hash();
+        let (_, commit_secret) = updated_tree.apply_commit(
+            &self.context.get_encoding(),
+            &self.owned_kp.init_private_key,
+            &commit_content,
+        )?;
+        // "If the path value is not populated: ..."
+        let commit_secret =
+            commit_secret.unwrap_or_else(|| SecretVec::new(vec![0; self.tree.cs.hash_len()]));
```
tomtau, line 381: should it also validate that for those two cases (`pathRequired = haveUpdateOrRemove || haveNoProposalsAtAll`), it really indeed populated the path instead of continuing with empty vec?

chain-tx-enclave-next/mls/src/group.rs
yihuang: > ## a few clippy suggestions: https://travis-ci.org/github/crypto-com/chain/jobs/702102278#L1710
> I guess for that `(Vec<(NodeSize, KeyPackage)>, Option<SecretVec<u8>>)`, it could either be a type alias with a comment explaining what these are, or struct with fields with meaningful names.
> looks ok, the only thing is the "genesis" initialisation -- I suppose we'd want to populate the path in those first "static" messages that would be put in genesis?

It seems the first part of the tuple is not used yet, I removed it.

yihuang: > the fact that the path must be populated in these cases does not imply that in the other cases, it must not be populated.
> In particular for the group initialisation / at genesis, it must be populated, as there's no "init_secret" (so no entropy from there, as one just passes empty or 0 vec) and we don't use PSK or something like that.
> 
> so either dedicated internal function `init_commit` or some extra parameter to this one?

it seems the commit message generated from genesis initialization is not used by anyone? other nodes are initialized with the welcome message?

chain-tx-enclave-next/mls/src/group.rs
yihuang: > should it also validate that for those two cases (`pathRequired = haveUpdateOrRemove || haveNoProposalsAtAll`), it really indeed populated the path instead of continuing with empty vec?

done

chain-tx-enclave-next/mls/src/group.rs
tomtau: the group creator would self-apply it :)

tomtau:
nightly fails to compile, but it's not relevant to this PR: https://travis-ci.org/github/crypto-com/chain/jobs/702242920#L873

chain-tx-enclave-next/mls/src/group.rs
tomtau: If the path isn't populated at genesis, the group creator would need to also immediately create another empty Commit distributed along with Welcome

tomtau: (but I guess it's easier if it's just populated in the first commit?)

yihuang: > (but I guess it's easier if it's just populated in the first commit?)

Yes, I agree.
Currently I got another testing failure, which is caused by self-update proposal.
We need to update the parent hash extension of the keypackage included in update proposal when committing, which is seems a bit hacky, not sure is it correct approach.

yihuang:
- Removed `OwnedKeypackage` in favor of seperated `KeyPackage` and `KeyPackageSecret`, so we don't have duplicated keypackage in tree leaf and `owned_kp`.

chain-tx-enclave-next/mls/src/group.rs
```diff
@@ -66,19 +70,68 @@ impl GroupAux {
...
+        let to_be_signed = MLSPlaintextTBS {
+            context: self.context.clone(),
+            content: content.clone(),
+        }
+        .get_encoding();
+        let signature = self.kp_secret.credential_private_key.sign(&to_be_signed);
+        MLSPlaintext { content, signature }
+    }
+
+    pub fn get_signed_remove(&self, index: LeafSize) -> MLSPlaintext {
```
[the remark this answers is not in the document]
yihuang: done

yihuang:
> the keypackage creation needs updating:
> 
> ```
> error[E0425]: cannot find value `credential_private_key` in this scope
> 1301	   --> chain-tx-enclave-next/mls/src/keypackage.rs:269:17
> 1302	    |
> 1303	269 |                 credential_private_key,
> 1304	    |                 ^^^^^^^^^^^^^^^^^^^^^^ a field by this name exists in `Self`
> 1305	
> 1306	error[E0425]: cannot find value `init_private_key` in this scope
> 1307	   --> chain-tx-enclave-next/mls/src/keypackage.rs:270:17
> 1308	    |
> 1309	270 |                 init_private_key,
> 1310	    |                 ^^^^^^^^^^^^^^^^ a field by this name exists in `Self`
> 1311	
> 1312	error[E0425]: cannot find value `keypackage` in this scope
> 1313	   --> chain-tx-enclave-next/mls/src/keypackage.rs:272:13
> 1314	    |
> 1315	272 |             keypackage,
> 1316	    |             ^^^^^^^^^^ not found in this scope
> 1317	
> 1318	error: aborting due to 3 previous errors
> ```

Fixed

PR #1803, opened 2020-06-23
Problem: one cannot optionally disable light client verification (fix…

leejw51crypto:
options is)
--disable-light-client-verifcation

CRYPTO_GENESIS_FINGERPRINT check will be skipped,
default options is the same with now. (enabled)

### DIFF
--- a/client-core/src/wallet/default_wallet_client.rs
+++ b/client-core/src/wallet/default_wallet_client.rs
@@ -1168,7 +1168,7 @@ where
         let sync_state = if let Some(sync_state) = mstate {
             sync_state
         } else {
-            get_genesis_sync_state(&self.tendermint_client)?
+            get_genesis_sync_state(&self.tendermint_client, true)?
         };
         Ok(sync_state)
     }
--- a/client-core/src/wallet/syncer.rs
+++ b/client-core/src/wallet/syncer.rs
@@ -257,7 +257,9 @@ impl<
         let sync_state = if let Some(sync_state) = mstate {
             sync_state
         } else {
-            get_genesis_sync_state(&env.client)?
+            // if fast-forward, don't check genesis fingerprint
+            let enable_genesis_fingerprint_check = !env.options.enable_fast_forward;
+            get_genesis_sync_state(&env.client, enable_genesis_fingerprint_check)?
         };
 
         let wallet_state =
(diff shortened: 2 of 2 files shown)


calvinlauco:
@leejw51crypto  I am reading the original issue. I am curious what would be the situation we want to disable the verification for client-rpc.

leejw51crypto:
for example, you trust a full node inside local lan, then you can use this option.
skipping client-side verification

slight performance benefit, but i think that's almost negligible

tomtau:
> @leejw51crypto I am reading the original issue. I am curious what would be the situation we want to disable the verification for client-rpc.

I'm not sure if this use case is possible, but one thing @lezzokafka did before was that he wanted to check the state of (any) staking state -- before the light client verification, it was possible without syncing (one would just retrieve from abci without the merkle proof)... now syncing is necessary.
But not sure if this use case is still relevant if one can display staking state info in the explorer?

client-cli/src/command.rs
```diff
@@ -156,6 +156,13 @@ pub enum Command {
             case_insensitive = false
         )]
         hardware: Option<HardwareKind>,
+
```
tomtau, line 159: I think those arguments should be *optional* -- and the default value should be false (one will want to run the light client verification by default) ?

client-rpc/server/src/program.rs
```diff
@@ -63,6 +63,14 @@ pub struct Options {
         help = "Disable address recovery when syncing wallet, which is not necessary, when addresses already exist"
     )]
     pub disable_address_recovery: bool,
+
+    #[structopt(
+        name = "disable-light-client-verification",
+        long,
+        help = "Disable light client verification, which doesn't check CRYPTO_GENESIS_FINGERPRINT"
```
tomtau, line 70: optional? false default?

client-cli/src/command.rs
leejw51crypto: ok, i'll change to option

client-rpc/server/src/program.rs
leejw51crypto: change to "option", default is false.

client-cli/src/command.rs
leejw51crypto: there is use case 
without genesis checking
and fast forward
-------------
default is
fast forward + genesis checking

tomtau: @leejw51crypto what about just disable genesis checking on fast forward? then no need for extra parameter -- just use "fast forward"?

leejw51crypto: ok, i'll remove option,
skip genesis checking(lite client check) on fast forward mode

leejw51crypto:
in fast-forward, skip genesis finger print check

yihuang:
> > @leejw51crypto I am reading the original issue. I am curious what would be the situation we want to disable the verification for client-rpc.
> 
> @calvinaco I'm not sure if this use case is possible, but one thing @lezzokafka did before was that he wanted to check the state of (any) staking state -- before the light client verification, it was possible without syncing (one would just retrieve from abci without the merkle proof)... now syncing is necessary.
> But not sure if this use case is still relevant if one can display staking state info in the explorer?

One can still use tendermint's abci_query directly, for example, with `chainrpc.py` one can do both:
```
$ chainrpc.py staking state 0x***
go through light client
$ chainrpc.py chain staking 0x***
go through abci_query
```

PR #1834, opened 2020-06-29
Problem (Fix #1833): mls tree math api don't separate leaf and parent in type

### LINKED ISSUE #1833: Problem: mls tree math api don't separate leaf and parent in type
We can make it explicit in type when we can make sure it's leaf index or parent index.

yihuang:
Solution:

- refactor tree math api,
- also change to use u32 since messages in mls use u32 to encode the node index.

### DIFF
--- a/chain-tx-enclave-next/mls/src/astree.rs
+++ b/chain-tx-enclave-next/mls/src/astree.rs
@@ -17,11 +17,9 @@ impl<D: Input + BlockInput + FixedOutput + Reset + Default + Clone> TreeBaseKeyS
     pub fn new(application_secret: SecretVec<u8>, leaf_count: LeafSize) -> Self {
         let root_index = NodeSize::root(leaf_count);
         let node_len = NodeSize::node_width(leaf_count);
-        let mut secrets = Vec::with_capacity(node_len.0);
-        for _ in 0..node_len.0 {
-            secrets.push(None);
-        }
-        secrets[root_index.0] = Some(application_secret);
+        let mut secrets = Vec::new();
+        secrets.resize_with(node_len.0 as usize, <Option<SecretVec<u8>>>::default);
+        secrets[root_index.node_index()] = Some(application_secret);
         Self {
             secrets,
             marker: PhantomData,
@@ -33,13 +31,14 @@ impl<D: Input + BlockInput + FixedOutput + Reset + Default + Clone> TreeBaseKeyS
         group_context_hash: Vec<u8>,
         sender_leaf: LeafSize,
     ) -> SecretVec<u8> {
-        let node_index = NodeSize::from_leaf_index(sender_leaf);
-        let leaf_count =
-            LeafSize::from_nodes(NodeSize(self.secrets.len())).expect("invalid node count");
+        let node_index = NodeSize::from(sender_leaf);
+        let leaf_count = NodeSize(self.secrets.len() as u32)
+            .leafs_len()
+            .expect("invalid node count");
         let d = node_index.direct_path(leaf_count);
         let mut found_i = None;
         for (i, node) in d.iter().enumerate() {
-            if self.secrets[node.0].is_some() {
+            if self.secrets[node.node_index()].is_some() {
                 found_i = Some(i);
                 break;
             }
@@ -47,12 +46,12 @@ impl<D: Input + BlockInput + FixedOutput + Reset + Default + Clone> TreeBaseKeyS
         use generic_array::typenum::Unsigned;
         let secret_len = D::OutputSize::to_u16();
         for node in d[..found_i.unwrap()].iter().rev() {
-            let l = node.left().expect("should have left");
-            let r = node.right(leaf_count).expect("should have right");
-            let secret = self.secrets[node.0].take().unwrap();
+            let l = node.left();
+            let r = node.right(leaf_count);
+            let secret = self.secrets[node.node_index()].take().unwrap();
             let skdf = Hkdf::<D>::new(None, secret.expose_secret());
 
-            self.secrets[l.0] = Some(SecretVec::new(
+            self.secrets[l.node_index()] = Some(SecretVec::new(
                 skdf.derive_app_secret(
                     group_context_hash.clone(),
                     "tree",
@@ -62,7 +61,7 @@ impl<D: Input + BlockInput + FixedOutput + Reset + Default + Clone> TreeBaseKeyS
                 )
                 .expect("left"),
             ));
-            self.secrets[r.0] = Some(SecretVec::new(
+            self.secrets[r.node_index()] = Some(SecretVec::new(
                 skdf.derive_app_secret(
                     group_context_hash.clone(),
                     "tree",
@@ -77,7 +76,9 @@ impl<D: Input + BlockInput + FixedOutput + Reset + Default + Clone> TreeBaseKeyS
             drop(secret);
         }
 
-        self.secrets[node_index.0].take().expect("sender secret")
+        self.secrets[node_index.node_index()]
+            .take()
+            .expect("sender secret")
     }
 }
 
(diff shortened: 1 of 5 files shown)


chain-tx-enclave-next/mls/src/astree.rs
```diff
@@ -17,11 +17,9 @@ impl<D: Input + BlockInput + FixedOutput + Reset + Default + Clone> TreeBaseKeyS
     pub fn new(application_secret: SecretVec<u8>, leaf_count: LeafSize) -> Self {
         let root_index = NodeSize::root(leaf_count);
         let node_len = NodeSize::node_width(leaf_count);
```
leejw51crypto, line 19: adding type for root_index?

chain-tx-enclave-next/mls/src/astree.rs
```diff
@@ -17,11 +17,9 @@ impl<D: Input + BlockInput + FixedOutput + Reset + Default + Clone> TreeBaseKeyS
...
-        let mut secrets = Vec::with_capacity(node_len.0);
-        for _ in 0..node_len.0 {
-            secrets.push(None);
-        }
-        secrets[root_index.0] = Some(application_secret);
+        let mut secrets = Vec::new();
```
leejw51crypto, line 20: how about
vec![...]. instead of resize_with

chain-tx-enclave-next/mls/src/astree.rs
```diff
@@ -157,10 +158,8 @@ pub struct GroupKeySource<D: Input + BlockInput + FixedOutput + Reset + Default
 impl<D: Input + BlockInput + FixedOutput + Reset + Default + Clone> GroupKeySource<D> {
     pub fn new(application_secret: SecretVec<u8>, leaf_count: LeafSize) -> Self {
         let base_secret_source = TreeBaseKeySource::new(application_secret, leaf_count);
-        let mut ratchets = Vec::with_capacity(leaf_count.0);
-        for _ in 0..leaf_count.0 {
-            ratchets.push(None);
-        }
+        let mut ratchets = Vec::new();
```
leejw51crypto, line 161: let mut ratchets: Vec<..>=vec![]  
as one line?

chain-tx-enclave-next/mls/src/astree.rs
```diff
@@ -173,16 +172,14 @@ impl<D: Input + BlockInput + FixedOutput + Reset + Default + Clone> GroupKeySour
...
                 .get_base_secret(group_context_hash.clone(), sender_leaf);
-            self.ratchets[sender_leaf.0] = Some(HashRatchet::new(
-                NodeSize::from_leaf_index(sender_leaf),
-                base_secret,
-            ));
+            self.ratchets[sender_leaf.0 as usize] =
+                Some(HashRatchet::new(NodeSize::from(sender_leaf), base_secret));
         };
-        let ratchet = self.ratchets[sender_leaf.0].as_mut().unwrap();
+        let ratchet = self.ratchets[sender_leaf.0 as usize].as_mut().unwrap();
```
leejw51crypto, line 182: expect()?

chain-tx-enclave-next/mls/src/keypackage.rs
```diff
@@ -266,24 +264,26 @@ impl OwnedKeyPackage {
...
+            },
+            keypackage,
+        ))
     }
 
     /// re-sign payload
-    pub fn update_signature(&mut self) {
-        self.keypackage.signature = self
+    pub fn update_signature(&self, keypackage: &mut KeyPackage) {
+        keypackage.signature = self
```
leejw51crypto, line 278: why &mut?

chain-tx-enclave-next/mls/src/message.rs
```diff
@@ -554,3 +569,93 @@ impl Codec for Sender {
...
+            }
+        };
+
+        // "Verify that the path value is populated if either of the updates or removes vectors has length greater than zero
+        // all of the updates, removes, and adds vectors are empty."
+        if commit.path.is_none()
+            && ((!commit.updates.is_empty() || !commit.removes.is_empty())
+                || (commit.adds.is_empty()
+                    && commit.updates.is_empty()
+                    && commit.removes.is_empty()))
```
leejw51crypto, line 607: can be simplified using variable?

chain-tx-enclave-next/mls/src/astree.rs
yihuang: Oh, it can't do that because the secret type don't implement `Clone`.

chain-tx-enclave-next/mls/src/message.rs
```diff
@@ -554,3 +569,93 @@ impl Codec for Sender {
...
+                    .and_then(|add| add.get_add().cloned())
+                    .ok_or(())
+            })
+            .collect::<Result<Vec<_>, _>>()?;
+        let updates = commit
+            .updates
+            .iter()
+            .map(|proposal_id| {
+                proposals_ids
+                    .get(proposal_id)
```
leejw51crypto, line 632: depth is too deep, can it be simpilfied?

chain-tx-enclave-next/mls/src/astree.rs
yihuang: it can't do that because the secret type don't implement `Clone`.

chain-tx-enclave-next/mls/src/keypackage.rs
yihuang: because it updates the signature field

chain-tx-enclave-next/mls/src/message.rs
leejw51crypto: can it be like 'haskell' currying style?

yihuang: yeah, Haskell syntax would be better for this style 😄, but I'm not sure how to do it better in rust.

chain-tx-enclave-next/mls/src/astree.rs
yihuang: it's set to `Some` immediately previous line of code.

chain-tx-enclave-next/mls/src/message.rs
yihuang: done

yihuang:
> conflict with the latest master

rebased.

chain-tx-enclave-next/mls/src/astree.rs
tomtau: or `Vec::with_capacity`?

chain-tx-enclave-next/mls/src/astree.rs
tomtau: should those two be `self.ratchets[sender_leaf.node_index()]` instead of `self.ratchets[sender_leaf.0 as usize]`?

chain-tx-enclave-next/mls/src/tree_math.rs
```diff
@@ -28,125 +28,140 @@
...
+use std::convert::{From, TryFrom};
+use std::iter;
+
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub struct LeafSize(pub u32);
 #[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord)]
-pub struct LeafSize(pub usize);
+pub struct ParentSize(pub u32);
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub struct NodeSize(pub u32);
```
tomtau, line 39: I wonder if it's necessary to have the internal member as `pub`? there could instead be `Into` implementation -- or I guess this is needed in MLS types? Perhaps it could just be if these types have a decoder implementation?

yihuang: > should those two be self.ratchets[sender_leaf.node_index()] instead of self.ratchets[sender_leaf.0 as usize]?

`node_index()` returns the index in the binary tree(`leaf_i * 2`), but ratchets is indexed by `leaf_i` itself.

yihuang: > or Vec::with_capacity?

I think `resize_with` will do single heap allocation, so there should be no performance difference here.

yihuang: Ideally we should encapsulate tree math within a tree abstraction, together with the `Vec<Node>`. I'm not sure how far can we go that way without abstraction leak. I want to try this idea, also separating secrets from public tree in another PR.

PR #1850, opened 2020-06-30
Problem: (Fix #1849) Makefile run chain-abci with tx_query set to docker hostname

### LINKED ISSUE #1849: Problem: Makefile run chain-abci with tx_query set to docker hostname
When `client-*` running on local or remote machine tries to connect, the `chain-abci` returns the docker hostname which is not accessible.

Solution: Provide an option to customize the hostname

calvinlauco:
Solution: Provide an option to customize the hostname

-----

Fix #1849

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,7 @@ data_path ?= /tmp/data
 prefix ?=
 sgx_mode ?= HW
 build_mode ?= debug
+TX_QUERY_HOSTNAME ?=
 MAKE_CMD = make
 
 # SGX_DEVICE can be /dev/sgx or /dev/isgx
@@ -276,6 +277,7 @@ run-abci:
 	-e NETWORK_ID=$(NETWORK_ID) \
 	-e CHAIN_ID=$(CHAIN_ID) \
 	-e PREFIX=$(prefix) \
+	-e TX_QUERY_HOSTNAME=$(TX_QUERY_HOSTNAME) \
 	-e APP_HASH=$(APP_HASH) \
 	-e TX_VALIDATION_CONN=ipc:///root/sockets/enclave.socket \
 	--name $(prefix)chain-abci \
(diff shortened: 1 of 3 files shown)


calvinlauco:
> looks ok, do any images on dockerhub need to be rebuilt?

In the `Makefile`, the `docker/Dockerfile` is built locally, do our Dockerhub images use this `Dockerfile`?

allthatjazzleo:
it is built locally

> > looks ok, do any images on dockerhub need to be rebuilt?
> 
> In the `Makefile`, the `docker/Dockerfile` is built locally, do our Dockerhub images use this `Dockerfile`?

it is built locally from docker/Dockerfile. No need to push to dockerhub. Dockerhub images use ci-scripts/Dockerfile instead.

Makefile
```diff
@@ -276,6 +277,7 @@ run-abci:
 	-e NETWORK_ID=$(NETWORK_ID) \
 	-e CHAIN_ID=$(CHAIN_ID) \
 	-e PREFIX=$(prefix) \
+	-e TX_QUERY_HOSTNAME=$(TX_QUERY_HOSTNAME) \
```
allthatjazzleo, line 280: need to change container name in run-sgx-query-next to match
https://github.com/crypto-com/chain/blob/eb69f7e278331a40672b2de5fc992852cd4649a2/Makefile#L255

```
--name $(prefix)$(TX_QUERY_HOSTNAME) \
```

calvinlauco: Sorry, can you elaborate?

allthatjazzleo: if user provide ENV `TX_QUERY_HOSTNAME=somethingelse` to run make command,
chain-abci in `ci-scripts/run_chain_abci.sh` cannot connect to tx-query container as the container name is still sgx-query-next

allthatjazzleo: `TX_QUERY_HOSTNAME` serves different purpose than I expected
so it makes sense now

PR #1859, opened 2020-07-01
Problem (Fix #1835): non-determinism in reward distribution

### LINKED ISSUE #1835: Problem: non-determinism in reward distribution
(being investigated)

yihuang:
Solution:
- Add block height assertion in begin block to prevent consensus
  connection re-connect

### DIFF
--- a/chain-abci/src/app/mod.rs
+++ b/chain-abci/src/app/mod.rs
@@ -148,6 +148,10 @@ impl<T: EnclaveProxy> abci::Application for ChainNodeApp<T> {
             .last_state
             .as_mut()
             .expect("executing begin block, but no app state stored (i.e. no initchain or recovery was executed)");
+        assert_eq!(
+            block_height,
+            last_state.block_height.checked_add(1).unwrap()
+        );
         last_state.block_time = block_time;
         last_state.block_height = block_height;
 
--- a/chain-abci/tests/abci_app.rs
+++ b/chain-abci/tests/abci_app.rs
@@ -516,9 +516,10 @@ fn get_block_proposer(app: &ChainNodeApp<MockClient>) -> TendermintValidatorAddr
         .validator_address()
 }
 
-fn begin_block(app: &mut ChainNodeApp<MockClient>) {
+fn begin_block(app: &mut ChainNodeApp<MockClient>, height: i64) {
     let mut bbreq = RequestBeginBlock::default();
     let mut header = Header::default();
+    header.set_height(height);
     header.set_time(::protobuf::well_known_types::Timestamp::new());
     header.set_proposer_address(Into::<[u8; 20]>::into(&get_block_proposer(app)).to_vec());
     bbreq.set_header(header);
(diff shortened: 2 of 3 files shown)


devashishdxt:
What was the cause of non-determinism?

yihuang:
> What was the cause of non-determinism?

- begin block n
- consensus connection closed
- tendermint restarted
- begin block n again

begin block n event might be processed more than once before commit.

devashishdxt:
> > What was the cause of non-determinism?
> 
> * begin block n
> * consensus connection closed
> * tendermint restarted
> * begin block n again
> 
> begin block n event might be processed more than once before commit.

Why tendermint’s consensus connection got closed? Was there any error on tendermint side that caused tendermint to terminate connection?

yihuang:
> > > What was the cause of non-determinism?
> > 
> > 
> > 
> > * begin block n
> > * consensus connection closed
> > * tendermint restarted
> > * begin block n again
> > 
> > begin block n event might be processed more than once before commit.
> 
> Why tendermint’s consensus connection got closed? Was there any error on tendermint side that caused tendermint to terminate connection?

There are logs in [slack discussions](https://crypt0-hq-sz.slack.com/archives/GMM494MAT/p1593480718071100?thread_ts=1593271479.043000&cid=GMM494MAT)

devashishdxt:
Can we figure out the reason for tendermint’s consensus connection disconnection from logs?

yihuang:
> Can we figure out the reason for tendermint’s consensus connection disconnection from logs?

It logged receiving an empty begin block response for some reason:
```
Jun 26 20:19:17 sgx-testnet-uksouth-fullnode-202005040919-1 chain-abci[3801]: [2020-06-26T20:19:17Z DEBUG abci::server] Return Response! begin_block {}
```

tomtau:
> Can we figure out the reason for tendermint’s consensus connection disconnection from logs?

"funky" deployment where processes are restarted individually

yihuang:
> I think this sort of assertion should be in rust-abci instead of in each application.
> upgrading to abci = "0.7.1" doesn't work?

I think current abci don't assert the internals of messages, add assertion on block height could solve the issue more thoroughly. maybe abci can add more assertions on state machine transition or internals of messages.

tomtau:
> > I think this sort of assertion should be in rust-abci instead of in each application.
> > upgrading to abci = "0.7.1" doesn't work?
> 
> I think current abci don't assert the internals of messages, add assertion on block height could solve the issue more thoroughly. maybe abci in the future can add more assertions on state machine transition or internals of messages.

yes https://github.com/tendermint/rust-abci/issues/49

there are also a few ignored tests: https://github.com/crypto-com/chain/blob/master/chain-abci/tests/abci_app.rs#L633

current abci doesn't assert them, but a new version doesn't throw away the errors like before, so it'd be good to check whether this fixes the connection drop / it'll fail with the connection drop (so one restarts from the committed state)

devashishdxt:
@tomtau What should ideally be the behaviour in such scenarios? Should ABCI app panic or should it discard current uncommitted state and start processing new `BeginBlock`?

tomtau:
@devashishdxt it's not specified anywhere at the moment.
maybe the easiest is if it gracefully shutdowns on connection drops, so that it restarts altogether with the TM process and reloads the committed state -- I tried with "0.7.1", and it doesn't shutdown/panic

PR #1860, opened 2020-07-01
Problem: (fix #1790) client-* cannot obtain HD path for HW wallet operations

### LINKED ISSUE #1790: Problem: client-* cannot obtain HD path for HW wallet operations
the high level methods for working with the HW wallet device will be provided in a library and they will look something like:
```
    /// Retrieves the public key and address
    pub async fn get_address(
        &self,
        path: &BIP44Path,
        require_confirmation: bool,
    ) -> Result<(Address, PublicKey), LedgerError> {
....

    /// Sign a transaction
    pub async fn sign(&self, path: &BIP44Path, message: &[u8]) -> Result<Signature, LedgerError> {
...
```

For signing with HW wallet, client-* should know which BIP44 path to pass to the device, but this information can't be reconstructed or obtained from anywhere in the current implementation

linfeng-crypto:
solution:
* add hd_path in client-* side
* modify mock_hardware_wallet
* add mock hardware wallet in integration test
* add hd_path and hdkey information when export and import wallet

### DIFF
--- a/client-core/src/service/wallet_service.rs
+++ b/client-core/src/service/wallet_service.rs
@@ -1,14 +1,16 @@
 use indexmap::IndexSet;
 use parity_scale_codec::{Decode, Encode, Input, Output};
 
-use crate::service::{load_wallet_state, WalletState};
+use crate::hd_wallet::ChainPath;
+use crate::service::{load_wallet_state, HdKey, WalletState};
 use crate::types::WalletKind;
 use chain_core::common::H256;
 use chain_core::init::address::RedeemAddress;
 use chain_core::state::account::StakedStateAddress;
 use chain_core::tx::data::address::ExtendedAddr;
 use client_common::{
-    Error, ErrorKind, PrivateKey, PublicKey, Result, ResultExt, SecKey, SecureStorage, Storage,
+    Error, ErrorKind, MultiSigAddress, PrivateKey, PublicKey, Result, ResultExt, SecKey,
+    SecureStorage, Storage,
 };
 use secstr::SecUtf8;
 use serde::de::{self, Visitor};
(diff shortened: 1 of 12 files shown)


client-core/src/service/wallet_service.rs
```diff
@@ -109,7 +114,11 @@ pub struct WalletInfo {
     /// passphrase used when import wallet
     pub passphrase: Option<SecUtf8>,
     ///public_key -> encoded private_key pairs, private key is None for hardware wallet
-    pub key_pairs: BTreeMap<PublicKey, Option<Vec<u8>>>,
+    pub key_pairs: BTreeMap<PublicKey, Vec<u8>>,
+    /// public_key -> hd path pairs for hardware wallet
+    pub key_chainpath: BTreeMap<PublicKey, String>,
+    /// hdkey for hd wallet and hw wallet
+    pub hdkey: Option<Vec<u8>>,
```
tomtau, line 121: I had to lookup what this is supposed to be -- so I assume it's that "struct" HdKey that has latest "change" address index? Why is it as `Vec<u8>` and not directly `HdKey`?
Also one thing to add a comment / explain regarding HW wallet -- there's a "seed" which won't be relevant to HW wallet (as the seed is on the device) -- so I assume here it's a seed that's different from HW wallet, only used for view keys / the mixed mode?

linfeng-crypto: I use`Vec<u8>` here because if I use sturct `HdKey` when export wallet,  I first get it from storage which is `Vec<u8>`, then decode to `HdKey`,  then encode to `Vec<u8>` to print out or store to a file, when import wallet, I should do decode/encode  to. If I use `Vec<u8>` directly, it's not need to do encode/decode.

integration-tests/tx-query2-enclave-app.sgxs
```diff
@@ -0,0 +1 @@
+../target/x86_64-fortanix-unknown-sgx/debug/tx-query2-enclave-app.sgxs
```
tomtau, line 1: what are these paths in files for?

linfeng-crypto: oh, the file created after I run the integration test, I'll remove it.

tomtau:
bors try

Issue #1845, opened 2020-06-30
Problem: outdated c-binding example 


leejw51crypto:
1. modify example to run in latest master
2. check sync example working

leejw51crypto:
working on it

tomtau:
are Java/Swift samples also affected?

leejw51crypto:
yes, need to modify
1. sync protocol- adding sync option
2. hdwallet creation- mnemonic count

tomtau:
seems this was resolved

PR #1861, opened 2020-07-01
Problem: outdated c-binding example (wip #1845)

leejw51crypto:
Solution: update sync json-rpc protocol  and added automatic `chain.h` modification for gcc

### DIFF
--- a/cro-clib/examples/Makefile
+++ b/cro-clib/examples/Makefile
@@ -1,5 +1,7 @@
 all: build run
 build:
+# for rocksdb link errors
+	export AR=/usr/bin/ar
 	cargo build
 	gcc -g test.c tx.c rpc.c -o cro -lcro_clib -lssl -lm -lcrypto -ldl -lpthread -L../../target/debug/ -lrocksdb -lstdc++
 run:
--- a/cro-clib/chain.h
+++ b/cro-clib/chain.h
@@ -19,8 +19,6 @@ typedef struct CroJsonRpc CroJsonRpc;
 
 typedef struct CroTx CroTx;
 
-typedef struct Option_ProgressCallback Option_ProgressCallback;
-
 typedef struct CroResult {
   int result;
 } CroResult;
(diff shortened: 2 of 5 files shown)


cro-clib/Cargo.toml
```diff
@@ -6,7 +6,9 @@ edition = "2018"
 
 [lib]
 name = "cro_clib"
-crate-type = ["staticlib"]
+#in ios, android, use staticlib
+#crate-type = ["staticlib"]
```
yihuang, line 10: remove this line, might confuse the sed script in `docker/build.sh`

leejw51crypto: ok, i'll check build.sh

leejw51crypto: also, will inspect travis

leejw51crypto:
rebased

cro-clib/Cargo.toml
```diff
@@ -6,7 +6,8 @@ edition = "2018"
 
 [lib]
 name = "cro_clib"
-crate-type = ["staticlib"]
+#in ios, android, use staticlib instead of cdylib
+crate-type = ["cdylib"]
```
tomtau, line 10: isn't staticlib better default?

cro-clib/examples/Makefile
```diff
@@ -1,17 +1,22 @@
 all: build run
 build:
+# for rocksdb link errors
+	export AR=/usr/bin/ar
 	cargo build
+	sed   's/Option_ProgressCallback/ProgressCallback/' -i  ../chain.h
```
tomtau, line 6: could this be patched in https://github.com/eqrion/cbindgen/blob/master/docs.md#cbindgentoml
there's "rename" or include/exclude for export?

cro-clib/Cargo.toml
leejw51crypto: for staticlib, link is OK, but when i run example on ubuntu, its crashes (seg-fault)

cro-clib/examples/Makefile
leejw51crypto: i'll check

cro-clib/Cargo.toml
tomtau: where does it segfault? can you do strace of what's the last thing it calls?

tomtau: or run it in a debugger?

leejw51crypto: i tried again, it's working fine with ["staticlib"], i'll change back

leejw51crypto:
removed `sed` post-processing,
changed to struct function wrapper
now, testing

leejw51crypto:
ok, i'll check clippy issues

PR #1864, opened 2020-07-01
Problem (Fix #1832): unnecessary rocksdb dependency in clients

### LINKED ISSUE #1832: Problem: `chain-storage` dependency in clients
chain-storge is only necessary for abci core.

rocksdb need c++ library,
so client-side only needs c-language sdk for compatiblity.
also rockdb is very big to include it in wasm, too.

it would be much handier to support `cro-clib` as clibrary (small and compact),
because c-lib has a wide range of compatibilities.

so i suggest as below.

1. make chain-storage only used in abci
2. remove dependency to chain-storage in
client-core,
client-network,
not in any client-side
3. if possible, decouple jelly-fish and kv-rocksdb , and make it feature-guarded, can use sled ,too.

yihuang:
Solution:
- make kvdb-rocksdb dependency in chain-storage optional

### DIFF
--- a/chain-storage/Cargo.toml
+++ b/chain-storage/Cargo.toml
@@ -9,7 +9,7 @@ edition = "2018"
 [dependencies]
 blake3 = "0.3.4"
 kvdb = "0.7"
-kvdb-rocksdb = "0.9"
+kvdb-rocksdb = { version = "0.9", optional = true }
 kvdb-memorydb = "0.7"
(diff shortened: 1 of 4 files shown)


yihuang:
> probably won't work on the workspace level, but shouldn't the `client-*` declare dependency on chain-storage with something like `default-features=false`, so that the `kvdb-rocksdb` isn't included?

yeah, that's better, done now.

PR #1900, opened 2020-07-07
Problem: ring is not updated to latest version

devashishdxt:
Solution: Upgraded ring to latest version


devashishdxt:
> maybe webpki can also be updated?

You mean in `Cargo.lock`?

devashishdxt:
> maybe webpki can also be updated?

Upgraded `webpki` also.

Issue #1868, opened 2020-07-02
Problem: no integration tests for cross-wallet operations


tomtau:
e.g. depositing to a different wallet

tomtau:
- deposit from one wallet to a different one's address
- withdraw one wallet to a different one's address(es)
- maybe transfer (but there may already be test with two different wallets)

PR #1908, opened 2020-07-08
Problem: (fix #1978) Outgoing deposit history is not displayed

linfeng-crypto:
solution: handle the block if transactions in the block is/are the self outgoing staking transaction(s)

### DIFF
--- a/client-common/src/tendermint/types/block_results.rs
+++ b/client-common/src/tendermint/types/block_results.rs
@@ -22,6 +22,10 @@ pub trait BlockResults {
     /// Returns true when the address presents
     fn contains_account(&self, target_account: &StakedStateAddress) -> Result<bool>;
 
+    /// Checks if the block contains a staking stransaction
+    /// Returns true when contains a staking transaction
+    fn contains_staking(&self) -> bool;
+
     /// Returns block filter in block results
     fn block_filter(&self) -> Result<BlockFilter>;
 }
@@ -47,6 +51,22 @@ impl BlockResults for BlockResultsResponse {
                 Ok(fees)
             }
         }
+    }
+
+    fn contains_staking(&self) -> bool {
+        if let Some(deliver_tx) = &self.txs_results {
+            for deliver_tx in deliver_tx.iter() {
+                if deliver_tx
+                    .events
+                    .iter()
+                    .map(|e| &e.type_str)
+                    .any(|x| x == &TendermintEventType::StakingChange.to_string())
+                {
+                    return true;
+                }
+            }
+        }
+        false
     }
 
     fn contains_account(&self, target_account: &StakedStateAddress) -> Result<bool> {
--- a/client-core/src/wallet/syncer_logic.rs
+++ b/client-core/src/wallet/syncer_logic.rs
@@ -124,6 +124,11 @@ pub(crate) fn handle_transaction(
         block_height,
         block_time,
     )?;
+    if TransactionType::from(transaction) == TransactionType::Deposit
+        && transaction_change.inputs.is_empty()
+    {
+        return Ok(());
+    }
     for input in transaction_change.inputs.iter() {
         memento.remove_unspent_transaction(input.pointer.clone());
     }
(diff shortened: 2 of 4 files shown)


linfeng-crypto:
> Can you write some unit tests to check this?

sure, I'll also add some integration test in issue #1868

integration-tests/cleanup.sh
```diff
@@ -14,13 +14,13 @@ fi
 if [ -d data_offline ]; then
     rm -rf data_offline
 fi
-if [ -f tx-query2-enclave-app.sgxs ]; then
-    rm tx-query2-enclave-app.sgxs
+if [ -L tx-query2-enclave-app.sgxs ]; then
```
tomtau, line 17: `-L` checks for symbolic link? why? isn't `-f` correct for the cleanup?

tomtau:
bors try

integration-tests/cleanup.sh
linfeng-crypto: yes, in my test, the file is a link and  always exist after the integration test, and may add into git if I use `git add .` command:
https://github.com/crypto-com/chain/blob/master/integration-tests/run.sh#L18

yihuang: It seems `-f` is indeed incorrect here 👍

client-common/src/tendermint/types/block_results.rs
```diff
@@ -49,6 +53,25 @@ impl BlockResults for BlockResultsResponse {
         }
     }
 
+    fn contains_staking(&self) -> bool {
+        match &self.txs_results {
```
leejw51crypto, line 57: how about exit early, if self.txs_results is Noe, and remove `match`

client-core/src/wallet/syncer_logic.rs
```diff
@@ -274,6 +281,23 @@ mod tests {
             .collect()
     }
 
+    fn outgoing_staking_transaction() -> Transaction {
+        Transaction::DepositStakeTransaction(DepositBondTx {
+            inputs: vec![TxoPointer {
+                id: [3; 32],
```
leejw51crypto, line 287: why fill with value 3, some explanation  would be good

client-core/src/wallet/syncer_logic.rs
```diff
@@ -337,17 +361,30 @@ mod tests {
             .map(|wallet| wallet.view_key.clone())
             .collect::<Vec<_>>();
         let mut state = WalletState::default();
+        let outgoing_staking_tx = outgoing_staking_transaction();
+        let outgoing_staking_tx_id = outgoing_staking_tx.id();
+        state.pending_transactions.insert(
+            outgoing_staking_tx_id.clone(),
+            TransactionPending {
+                used_inputs: vec![TxoPointer::new([3; 32], 0)],
```
leejw51crypto, line 369: use pre-defined const?

client-common/src/tendermint/types/block_results.rs
```diff
@@ -49,6 +53,25 @@ impl BlockResults for BlockResultsResponse {
...
+            None => false,
+            Some(deliver_tx) => {
+                for deliver_tx in deliver_tx.iter() {
```
leejw51crypto, line 60: it's OK, how about using `fold`  or other chaining pattern for simplicity?

client-core/src/wallet/syncer_logic.rs
linfeng-crypto: it is unit test, I think it Ok here using `32`. Because `[0;32]` and `[1;32]` already used in the other type of transactions, and it should be unique here to avoid confliction, so I use `[3; 32]`

client-core/src/wallet/syncer_logic.rs
linfeng-crypto: it's ok because it's unit test.

client-common/src/tendermint/types/block_results.rs
linfeng-crypto: yeah, I use `if let Some(x)...`, it can reduce code.

PR #1925, opened 2020-07-10
Problem: unnecessary `mock` function argument in cro-clib (fix #1792)

### LINKED ISSUE #1792: Problem: unnecessary `mock` function argument in cro-clib
Solution: use `mock-enclave` feature instead of `mock` function agument  in cro-lib

`mock` is used for development but should be kept away from the release binary.

https://github.com/crypto-com/chain/blob/master/cro-clib/src/jsonrpc.rs#L267

leejw51crypto:
Solution: remove mock functions


leejw51crypto:
check more

PR #1943, opened 2020-07-14
Problem: Attestation report is not transmitted and validated properly (fixes #1941)

### LINKED ISSUE #1941: Problem: Attestation report is not transmitted and validated properly
**Describe the bug**
1. When use `ra-enclave` to start a TLS server with RA verification, the signing certificate embedded in the x509 certificate extension is URL encoded.
2. On the client side, `ra-client` receives the certificate and do the verification. It failed to parse the attestation signing certificateD but no exception is returned. The verification is then treated as successful.

**To Reproduce**
1. Start a TLS server with `EnclaveRaContext` in `ra-enclave`.
2. Establish a connection with `ra-client`.
3. Intercept the X509 certificate sent by the server and examine the extensions.
4. The signing certificate of the attestation report is URL encoded.
5. Debug the code on client side. The verification logic in `chain/chain-tx-enclave-next/enclave-ra/ra-client/src/verifier.rs` after parsing the signing certificate chain is not executed.

**Expected behavior**
1. The attestation report and signing certificate should be verified.
2. If the attestation report or signing certificate can't be parsed or verified, errors should be returned
3. the signing certificate provided by the server shouldn't be URL encoded.

jimmyyip-crypto:
Solution:
1. URL decode the RA report signing certificate
2. Return error when missing signing certificate in TLS server certificate
3. Fix signing certificate verification logic
4. Fix attestation report report data verification logic
5. Add test cases

I hereby certify that my contribution is in accordance with the Developer Certificate of Origin (https://developercertificate.org/).


chain-tx-enclave-next/enclave-ra/ra-client/src/verifier.rs
```diff
@@ -294,6 +322,14 @@ impl ClientCertVerifier for EnclaveCertVerifier {
 
 #[derive(Debug, Error)]
 pub enum EnclaveCertVerifierError {
+    #[error("Unable to parse attestation report: {0}")]
+    AttestationReportParsingError(#[source] serde_json::Error),
```
devashishdxt, line 326: Its better to use `#[from]` instead of `#[source]`. From `thiserror` [documentation](https://docs.rs/thiserror/1.0.20/thiserror/#details):

> The `#[from]` attribute always implies that the same field is `#[source]`, so you don't ever need to specify both attributes.

chain-tx-enclave-next/enclave-ra/ra-client/src/verifier.rs
```diff
@@ -201,8 +229,8 @@ impl EnclaveCertVerifier {
         let quote = attestation_report_body.get_quote()?;
 
         if public_key.len() != 65
-            && public_key[0] != 4
-            && public_key[1..] != quote.report_body.report_data[..]
+            || public_key[0] != 4
+            || public_key[1..] != quote.report_body.report_data[..]
```
tomtau, line 233: a good catch!
It may be good to give names to the conditions, as there's currently no documentation e.g. referring to SECG.

```
let has_correct_len = public_key.len() == 65;
let is_uncompressed =  public_key[0] == 4;
let pubkey_matches = public_key[1..] == quote.report_body.report_data[..];
if ! has_correct_len || ! is_uncompressed || !pubkey_matches {
...
}

```

chain-tx-enclave-next/enclave-ra/ra-client/test/Intel_SGX_Attestation_RootCA.pem
```diff
@@ -0,0 +1,31 @@
+-----BEGIN CERTIFICATE-----
```
tomtau, line 1: I guess this one is the same as https://github.com/crypto-com/chain/blob/master/client-common/src/cipher/AttestationReportSigningCACert.pem -- no need to duplicate it?

chain-tx-enclave-next/enclave-ra/ra-client/src/verifier.rs
```diff
@@ -339,3 +377,233 @@ pub struct CertVerifyResult {
...
+        };
+        let verifier = EnclaveCertVerifier::new(verifier_config).unwrap();
+        let result = verifier.verify_attestation_report(attestation_report, public_key, Utc::now());
+
+        assert!(matches!(
+            result.unwrap_err(),
+            EnclaveCertVerifierError::PublicKeyMismatch
+        ));
+    }
+}
```
tomtau: all the tests use `Utc::now()` which would make them non-deterministic (e.g. valid report may become invalid after this expires) -- it may be better to use some fixed reference time in tests

chain-tx-enclave-next/enclave-ra/ra-client/src/verifier.rs
jimmyyip-crypto: Generic Error WebpkiError and JsonError have already done the auto casting of `webpki::Error` and `serde::Error` Can we leave the attestation report related errors to use `#[source]?

chain-tx-enclave-next/enclave-ra/ra-client/src/verifier.rs
tomtau: but not sure if this will still be an issue -- e.g. `webpki` methods may still use the current system time?

jimmyyip-crypto: found the issue as well
fixed in previous commit

tomtau:
@jimmyyip-crypto when you rebase and squash commits with fixes, you can also modify the commit message to the original issue `Problem: Attestation report is not transmitted and validated properly (fixes #1941)`, as that's clearer

jimmyyip-crypto:
> one small clippy issue: https://travis-ci.org/github/crypto-com/chain/jobs/709049393#L1757
> + squash commits and it should be good to go!

I got some more issues when I run `cargo clippy` locally. Do I have to fix them as well?

tomtau:
> > one small clippy issue: https://travis-ci.org/github/crypto-com/chain/jobs/709049393#L1757
> > 
> > * squash commits and it should be good to go!
> 
> I got some more issues when I run `cargo clippy` locally. Do I have to fix them as well?

yep, `cargo fmt` and `cargo clippy` are a part of the CI pipeline

PR #1948, opened 2020-07-14
Problem: Ambiguous message while importing an existing transaction (f…

leejw51crypto:
If the tx is already imported, showed message "This transaction is already imported/in your wallet"

### DIFF
--- a/client-core/src/wallet/default_wallet_client.rs
+++ b/client-core/src/wallet/default_wallet_client.rs
@@ -93,6 +93,21 @@ where
             block_height_ensure,
         }
     }
+
+    fn is_tx_exist(&self, name: &str, enckey: &SecKey, txid: TxId) -> Result<bool> {
+        let tx_change = self
+            .wallet_state_service
+            .get_transaction_history(name, enckey, false)?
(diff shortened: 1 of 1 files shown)


client-core/src/wallet/default_wallet_client.rs
```diff
@@ -1066,6 +1066,15 @@ where
     /// import a plain base64 encoded plain transaction
     fn import_plain_tx(&self, name: &str, enckey: &SecKey, tx_str: &str) -> Result<Coin> {
         let tx_info = TransactionInfo::decode(tx_str)?;
+
+        let found_tx = self.export_plain_tx(name, enckey, &hex::encode(tx_info.tx.id()));
```
calvinaco, line 1070: `export_plain_tx` is slightly overkill for checking transaction existence. Maybe a method to get only/check existence of the transaction from storage?

leejw51crypto: ok i will modifiy it

leejw51crypto:
added another function to check existence tx in the storage

PR #1958, opened 2020-07-16
Problem: (fix #1828)abci info endpoint returns uncommitted state

### LINKED ISSUE #1828: Problem: abci info endpoint returns uncommitted state
https://docs.tendermint.com/master/spec/abci/apps.html#info-connection

```
QueryState should be set to the latest DeliverTxState at the end of every Commit, ie. after the full block has been processed and the state committed to disk. Otherwise it should never be modified.
```

It currently returns uncommitted state, which could potentially cause data inconsistency.

linfeng-crypto:
solution: get info from storage.

### DIFF
--- a/chain-abci/src/app/mod.rs
+++ b/chain-abci/src/app/mod.rs
@@ -31,6 +31,7 @@ use chain_core::init::config::NetworkParameters;
 use chain_core::state::account::PunishmentKind;
 use chain_core::state::tendermint::{BlockHeight, TendermintValidatorAddress, TendermintVotePower};
 use chain_core::tx::TxAux;
+use parity_scale_codec::Decode;
 
 fn get_version() -> String {
     format!(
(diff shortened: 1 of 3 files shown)


chain-abci/src/app/mod.rs
```diff
@@ -68,7 +69,9 @@ impl<T: EnclaveProxy> abci::Application for ChainNodeApp<T> {
         let mut resp = ResponseInfo::new();
         resp.app_version = chain_core::APP_VERSION;
         resp.version = get_version();
-        if let Some(app_state) = &self.last_state {
+        if let Some(raw) = chain_storage::get_last_app_state(&kv_store!(self)) {
```
yihuang, line 72:
```suggestion
        if let Some(raw) = chain_storage::get_last_app_state(&self.storage) {
```
`kv_store!` is the storage combined with in-memory buffer.

tomtau:
bors retry

PR #1960, opened 2020-07-17
Problem (Fix #1959): cro-clib don't enable log

### LINKED ISSUE #1959: Problem: cro-clib don't enable log
logging can help with debugging.

yihuang:
Solution:
- initialize env_logger

Maybe better to make it into an option, but currently the only use case is the integration tests, so I guess it's good so far.

### DIFF
--- a/cro-clib/src/jsonrpc.rs
+++ b/cro-clib/src/jsonrpc.rs
@@ -194,6 +194,9 @@ unsafe fn create_rpc(
     progress_callback_user: CroProgressPtr,
     user_data: *const std::ffi::c_void,
 ) -> Result<CroJsonRpc> {
+    // `try_init` only returns `Err` if already initialized, `create_rpc` can be called multiple times,
+    // so we ignore the error
+    let _ = env_logger::try_init();
     let storage_dir = get_string(storage_dir);
     let websocket_url = get_string(websocket_url);
     let cbindingcallback: Option<CBindingCore> = if !progress_callback_user.is_null() {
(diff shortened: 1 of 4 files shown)


cro-clib/src/jsonrpc.rs
```diff
@@ -194,6 +194,7 @@ unsafe fn create_rpc(
     progress_callback_user: CroProgressPtr,
     user_data: *const std::ffi::c_void,
 ) -> Result<CroJsonRpc> {
+    let _ = env_logger::try_init();
```
devashishdxt, line 197: `try_init` returns a `Result`. Can we handle the error or just use `init`?

yihuang: Because `create_rpc` can be called multiple times in one process, so we have to use `try_init` and ignore the result (it returns error when called multiple times).

tomtau: I think it'll be "Err" if it was already initialised, so it should be ignored (maybe a comment can be added to explain that).
For `init`, it'll panic if it's called multiple times if I'm not mistaken

yihuang: > I think it'll be "Err" if it was already initialised, so it should be ignored (maybe a comment can be added to explain that).
> For `init`, it'll panic if it's called multiple times if I'm not mistaken

yes, added comments now

leejw51crypto: how about separating as another api?

tomtau:
bors retry

cro-clib/src/jsonrpc.rs
yihuang: there's no `env_logger::deinit()` though.
after `env_logger::init()`, use still can disable log by environment variables.

tomtau:
bors retry

PR #1969, opened 2020-07-21
Problem: (fix #1949) non-deterministic test-vectors with upstream schnorrsig

### LINKED ISSUE #1949: Problem: non-deterministic test-vectors with upstream schnorrsig
https://github.com/crypto-com/chain/blob/master/dev-utils/src/commands/test_vector_command.rs#L264
the auxiliary randomness can be fixed and reported as in BIP-340 https://github.com/bitcoin/bips/blob/master/bip-0340/test-vectors.csv

linfeng-crypto:
solution: use `schnorr_sign_aux` instead of `schnorr_sign` and allow users input a `aux_payload`

### DIFF
--- a/client-common/src/key/private_key.rs
+++ b/client-common/src/key/private_key.rs
@@ -2,8 +2,9 @@ use crate::Transaction;
 use chain_core::tx::TransactionId;
 use parity_scale_codec::{Decode, Encode, Error, Input, Output};
 use rand::rngs::OsRng;
-use secp256k1::schnorrsig::{schnorr_sign, SchnorrSignature};
+use secp256k1::schnorrsig::{schnorr_sign, schnorr_sign_aux, AuxRandNonce, SchnorrSignature};
 use secp256k1::{recovery::RecoverableSignature, Message, PublicKey as SecpPublicKey, SecretKey};
+use std::convert::TryInto;
 use zeroize::Zeroize;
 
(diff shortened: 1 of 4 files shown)


client-common/src/key/private_key.rs
```diff
@@ -49,6 +53,23 @@ impl PrivateKeyAction for PrivateKey {
         Ok(signature)
     }
 
+    fn schnorr_sign_aux(&self, tx: &Transaction, aux_payload: &[u8]) -> Result<SchnorrSignature> {
```
tomtau, line 56:
```suggestion
    fn schnorr_sign_unsafe(&self, tx: &Transaction, aux_payload: &[u8]) -> Result<SchnorrSignature> {
```

As this is not meant to be called anywhere else in the client and is only used in dev-utils, I suggest the name should reflect it's "dev-only"

dev-utils/src/dev_utils.rs
```diff
@@ -87,11 +96,27 @@ impl DevUtils {
...
+#[cfg(test)]
+mod tests {
+
+    #[test]
+    fn test_abc() {
+        let a: [u8; 32] = [0; 32];
+        let b = hex::encode(a);
+        println!("{}", b);
+    }
+}
```
tomtau, line 122: ?

linfeng-crypto: I'll remove it.

PR #1974, opened 2020-07-21
Problem: wallet operations slow down with a large number of addresses (fix #1939)

### LINKED ISSUE #1939: Problem: wallet operations slow down with a large number of addresses
Wallet is OK for a few thousands of addresses, for massive addresses,
it can be slow down by many copy-operations.

for example) in Wallet 
/// Returns all the multi-sig transfer addresses in current wallet
fn transfer_addresses(&self, name: &str, enckey: &SecKey) -> Result<IndexSet<ExtendedAddr>>;

Solution is , not to return all addresses, let the user-logic query to DB directly.
this will enhance overall speed with massive addresses both for staking, trasnsfer addresses (billions)

this issue  will also refactor current `memento` implementation, to use DB directly.
some change in sync, wallet state, too.

## bottle-neck
if there are a lof txs, and addresses, sync will take very long time.
because, in sync, per block, per handle_transaction,
all transaction addresses are copied for each handle_transaction.

## current
DB --> Wallet --> Wallet State -->  User Logic

## new
simplify it just to query to DB directly
DB ->  User Logic

## how to do it
because interface with outer will be the same, only internal working will be changed
also it will not need big memory, because it will not store all addresses to the memory.

## backwards compatibility
because it's only internal modification, database will be the same.

leejw51crypto:
## why?
1. removed temporary transfer, staking addresses storing in Wallet
2. query address to sled-storage directly, which is fast in massive address
3. replace 1:1 compare with db-query

## result which this pr solved
1. in massive addresses(over 4,000,000),
sync and other wallet functions will work as fast as when there are small number of addresses
2. also it will not store addresses in memory, which will save some memory space, also memory copying cost.
3. in a block, which contains a lot of txs, it will work very efficiently.
 
## added proxy apis
to query directly, added WalletStorage to Wallet.
WalletStorage is proxy for the storage using `dyn trait` with Arc+Mutex.

1. staking_addresses_contains -> for query
2. transfer_addresses_contains  -> for query
3. get_public_keys -> for import, export
4. get_roothashes -> for import, export

### DIFF
--- a/client-cli/src/command/wallet_command.rs
+++ b/client-cli/src/command/wallet_command.rs
@@ -270,13 +270,13 @@ impl WalletCommand {
 
         let wallet_info_list: Vec<WalletInfo> = serde_json::from_str(&wallet_info_str)
             .chain(|| (ErrorKind::InvalidInput, "Invalid wallet info list"))?;
-        for wallet_info in wallet_info_list {
+        for mut wallet_info in wallet_info_list {
             let name = wallet_info.name.clone();
             let passphrase = match &wallet_info.passphrase {
                 Some(p) => p.clone(),
                 None => ask_passphrase(Some(&format!("Input passphrase for wallet {}:", name)))?,
             };
-            let enckey = wallet_client.import_wallet(&name, &passphrase, wallet_info);
+            let enckey = wallet_client.import_wallet(&name, &passphrase, &mut wallet_info);
             match enckey {
                 Ok(enckey) => success(&format!(
                     "Authentication token of wallet {}: {}",
--- a/client-common/src/tendermint/types/block_results.rs
+++ b/client-common/src/tendermint/types/block_results.rs
@@ -20,7 +20,10 @@ pub trait BlockResults {
 
     /// Checks if a StakedStateAddress is included in devlier_tx account event.
     /// Returns true when the address presents
-    fn contains_account(&self, target_account: &StakedStateAddress) -> Result<bool>;
+    fn contains_account(
+        &self,
+        contain_account_callback: Box<dyn Fn(StakedStateAddress) -> bool>,
+    ) -> Result<bool>;
 
     /// Checks if the block contains a staking stransaction
     /// Returns true when contains a staking transaction
@@ -69,7 +72,10 @@ impl BlockResults for BlockResultsResponse {
         false
     }
 
-    fn contains_account(&self, target_account: &StakedStateAddress) -> Result<bool> {
+    fn contains_account(
+        &self,
+        contain_account_callback: Box<dyn Fn(StakedStateAddress) -> bool>,
+    ) -> Result<bool> {
         match &self.txs_results {
             None => Ok(false),
             Some(deliver_tx) => {
@@ -81,7 +87,7 @@ impl BlockResults for BlockResultsResponse {
                         match find_staking_address_from_event_attributes(&event.attributes)? {
                             None => continue,
                             Some(address) => {
-                                if address == *target_account {
+                                if contain_account_callback(address) {
                                     return Ok(true);
                                 }
                             }
@@ -248,7 +254,13 @@ mod tests {
             let target_account = StakedStateAddress::from(
                 RedeemAddress::from_str("0x0e7c045110b8dbf29765047380898919c5cb56f4").unwrap(),
             );
-            let result = block_results.contains_account(&target_account);
+
+            let check_staking_address =
+                Box::new(move |staked_state_address: StakedStateAddress| {
+                    staked_state_address == target_account
+                });
+
+            let result = block_results.contains_account(check_staking_address);
             assert!(result.is_err());
             assert_eq!(ErrorKind::DeserializationError, result.unwrap_err().kind());
         }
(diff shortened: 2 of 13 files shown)


tomtau:
bors try

tomtau:
bors try

tomtau:
bors try-

tomtau:
bors try

client-core/src/service/wallet_service.rs
```diff
@@ -329,35 +520,17 @@ pub fn load_wallet<S: SecureStorage>(
...
-        }
-
         // load walletkind
         let walletkind: u64 = read_number(storage, &info_keyspace, "walletkind", Some(0))?;
         new_wallet.wallet_kind = walletkind.into();
+        // set name, enckey
+        new_wallet.name = name.into();
+        new_wallet.enckey = Some(enckey.clone());
+        let newstorage = storage.clone();
+        new_wallet.wallet_storage = Some(Arc::new(Mutex::new(WalletStorageImpl::new(newstorage))));
```
yihuang, line 530: the code structure is a bit deviated from original intension, i guess it takes a bigger refactor to make it better ;D

leejw51crypto:
ok,i'll fix it

tomtau:
https://travis-ci.org/github/crypto-com/chain/jobs/710621720#L827

leejw51crypto:
i'll fix travis issue

leejw51crypto:
rebased

tomtau:
bors try

client-common/src/tendermint/types/block_results.rs
```diff
@@ -20,7 +20,10 @@ pub trait BlockResults {
 
     /// Checks if a StakedStateAddress is included in devlier_tx account event.
     /// Returns true when the address presents
-    fn contains_account(&self, target_account: &StakedStateAddress) -> Result<bool>;
+    fn contains_account(
+        &self,
+        contain_account_callback: Box<dyn Fn(StakedStateAddress) -> bool>,
```
devashishdxt, line 25: Can you add some documentation about what this callback function should do and also why this callback pattern is necessary?

client-cli/src/command/wallet_command.rs
```diff
@@ -270,13 +270,13 @@ impl WalletCommand {
 
         let wallet_info_list: Vec<WalletInfo> = serde_json::from_str(&wallet_info_str)
             .chain(|| (ErrorKind::InvalidInput, "Invalid wallet info list"))?;
-        for wallet_info in wallet_info_list {
+        for mut wallet_info in wallet_info_list {
```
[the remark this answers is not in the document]
leejw51crypto: that's to add `WalletStorage` to the Wallet, which enables direct access to the db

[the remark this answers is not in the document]
leejw51crypto: i'll try to do benchmark for `handle transaction`
1. with current master
2. with this pr

client-common/src/tendermint/types/block_results.rs
leejw51crypto: that's because, BlockResults cannot have ownership for DB, also it's complicated to give trait ownership to it.
so decoupling between BlockResults and filter logic in DB

client-cli/src/command/wallet_command.rs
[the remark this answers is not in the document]
devashishdxt: Ok

client-common/src/tendermint/types/block_results.rs
devashishdxt: Ok

tomtau:
bors retry

leejw51crypto:
what's bors message?

tomtau:
> what's bors message?

it failed because I manually merged the new logo PR -- it should be ok

tomtau:
bors retry

leejw51crypto:
## benchmark result

### for 1,000,000 addresses
new:handle_blocks time 12 micro-seconds
old: handle_blocks time 115563 micro-seconds
that would be 9630 times faster
### for 4,000,000 addresses
new: handle_blocks time 14 micro-seconds

tomtau:
@leejw51crypto averages? medians? deviation? perhaps benchmark with criterion?

leejw51crypto:
it's for testing just for sync speed,
because it reduces memory copying a lot.
speed will be the same

leejw51crypto:
i added  cherry pick of the benchmark,
after review i'll drop this commit

client-core/src/wallet/syncer.rs
```diff
@@ -422,8 +422,13 @@ impl<
             self.handle_recover_addresses(&blocks)?;
         }
 
+        let handle_blocks_time = std::time::Instant::now();
         let memento = handle_blocks(&self.wallet, &mut self.wallet_state, &blocks, &enclave_txs)
             .map_err(|err| Error::new(ErrorKind::InvalidInput, err.to_string()))?;
+        println!(
+            "handle_blocks time {} micro-seconds",
+            handle_blocks_time.elapsed().as_micros()
+        );
```
tomtau, line 431: perhaps `log::debug!`?

leejw51crypto: ok, i'll modify now

leejw51crypto:
added log for handle_block elapsed time , progress in generating large addresses

Issue #1784, opened 2020-06-18
Problem: possible panic of operations in mls modules


tomtau:
As TDBE will use the mls library as its part when processing data obtained from the light client, it shouldn't crash e.g. from malformed encryption etc.

tomtau:
@devashishdxt fyi, one other point raised on that call today -- besides eliminating unwraps, it'd be good to separate out library API logically, so it is clearer what's internal and its invariant is that the inputs already were validated for certain properties, versus external

tomtau:
e.g. something similar what's done with https://docs.rs/untrusted/0.7.1/untrusted/ and parsing in rustls

tomtau:
altogether with this or #1785 #1786 #1787 #1788, it'd be good to increase code coverage of the mls module -- e.g. add negative test cases for the possible error types

yihuang:
- Sign proposals
  - `get_signed_add`
  - `get_signed_remove`
  - `get_signed_self_update`
- commit proposals
  - `init_commit`
  - `commit_proposals`
- `process_commit`
- `init_group_from_welcome`

These are the public facing APIs directly get external input.

PR #2004, opened 2020-07-23
Problem (Fix #1979): ratchet tree mixed public and private informations

### LINKED ISSUE #1979: Problem: ratchet tree mixed public and private informations
Separate private keys from the public tree, reference to https://github.com/cisco/go-mls/pull/71

yihuang:
Solution:
- separated `TreeSecret`

### DIFF
--- a/chain-tx-enclave-next/mls/Cargo.toml
+++ b/chain-tx-enclave-next/mls/Cargo.toml
@@ -13,13 +13,14 @@ rustls = "0.18"
 x509-parser = "0.8.0-beta4"
 secrecy = "0.7.0"
 sha2 = "0.9"
-hkdf = "0.9"
+hkdf = { version = "0.9", features = ["std"] }
 hpke = { version = "0.1.8", default-features = false, features = ["p256"] }
 aead = "0.3"
 rand = "0.7"
 chrono="0.4.13"
 ra-client = { path = "../enclave-ra/ra-client" }
 subtle = "2.2.3"
+chain-util = { path = "../../chain-util", default-features = false }
 
 [dev-dependencies]
 chrono = "0.4"
--- a/chain-tx-enclave-next/mls/src/tree.rs
+++ b/chain-tx-enclave-next/mls/src/tree.rs
@@ -13,6 +13,7 @@ use crate::message::DirectPathNode;
 use crate::message::*;
 use crate::tree_math::{LeafSize, NodeSize, ParentSize};
 use crate::utils::{decode_option, encode_option, encode_vec_u32, read_vec_u32};
+use chain_util::non_empty::NonEmpty;
 use ra_client::AttestedCertVerifier;
 use rustls::internal::msgs::codec::{self, Codec, Reader};
 use secrecy::{ExposeSecret, SecretVec};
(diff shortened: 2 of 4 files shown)


chain-tx-enclave-next/mls/src/tree.rs
```diff
@@ -595,91 +532,165 @@ impl Tree {
...
-                            .expect("decrypt failed");
-                        return Some((ancestor, SecretVec::new(secret)));
-                    }
-                }
-                panic!("don't find private key to decrypt secret");
+                let pos = NodeSize::from(tree.my_pos)
+                    .direct_path(leaf_len)
+                    .iter()
+                    .position(|&p| p == ancestor)
+                    .expect("ancestor in direct path");
```
tomtau, line 677: I guess this is to be solved in https://github.com/crypto-com/chain/issues/1784 (as one may get a malformed commit)

chain-tx-enclave-next/mls/src/tree.rs
```diff
@@ -595,91 +532,165 @@ impl Tree {
...
+                let secret = match pos.checked_sub(1) {
```
tomtau, line 678: maybe some comment why `pos.checked_sub(1)` and what those two cases represent (no ancestors for None?)?

chain-tx-enclave-next/mls/src/tree.rs
```diff
@@ -689,89 +700,61 @@ impl Tree {
     /// find the overlap parent node, decrypt the secret and implant it
     ///
     /// returns (leaf_parent_hash, commit_secret)
-    fn apply_path_secrets(
+    pub fn apply_path_secrets(
```
tomtau, line 703: or `pub (crate)` ? with some many public APIs, it's hard to know which ones are to be called from outside + what checks are expected to be performed beforehand on arguments

chain-tx-enclave-next/mls/src/tree.rs
yihuang: The ancestor is computed from `my_pos` in previous step, so by definition it should appear in `direct_path` of `my_pos`, unless there's internal bugs.

chain-tx-enclave-next/mls/Cargo.toml
```diff
@@ -13,13 +13,14 @@ rustls = "0.18"
...
 sha2 = "0.9"
-hkdf = "0.9"
+hkdf = { version = "0.9", features = ["std"] }
 hpke = { version = "0.1.8", default-features = false, features = ["p256"] }
 aead = "0.3"
 rand = "0.7"
 chrono="0.4.13"
 ra-client = { path = "../enclave-ra/ra-client" }
 subtle = "2.2.3"
+non-empty-vec = "0.1.2"
```
tomtau, line 23: it can use chain-util now?

yihuang: yes, done

chain-tx-enclave-next/mls/src/tree.rs
yihuang: done

chain-tx-enclave-next/mls/src/tree.rs
yihuang: yes, added some comments now.

PR #2013, opened 2020-07-24
Problem:no option for disabling light client (fix #2012)

### LINKED ISSUE #2012: Problem:no option for disabling light client
add --disable-light-client option
because in zerofee, we cannot use --enable-fast-forward option
so
1. --enable-fast-forward : disable light client + enable fastforward
2. --disable-light-client: disable light client

leejw51crypto:
Solution: add --disable-light-client option to skip `light client`

because need to skip `light client` when fee is zero.
also with zero fee, we cannot use fast forward option.
so this option is necessary.

default is
1. not use fastward
2. use light client check
which is the most safest option

### DIFF
--- a/client-cli/src/command.rs
+++ b/client-cli/src/command.rs
@@ -209,6 +209,12 @@ pub enum Command {
         )]
         enable_fast_forward: bool,
         #[structopt(
+            name = "disable-light-client",
+            long,
+            help = "Disable light client, which is not secure when connecting to outside nodes"
+        )]
+        disable_light_client: bool,
+        #[structopt(
             name = "disable-address-recovery",
             long,
             help = "Disable address recovery, which is not necessary, if addresses already exist"
@@ -382,6 +388,7 @@ impl Command {
                 batch_size,
                 force,
                 enable_fast_forward,
+                disable_light_client,
                 disable_address_recovery,
                 block_height_ensure,
             } => {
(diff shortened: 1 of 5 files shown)


client-cli/src/command.rs
```diff
@@ -208,6 +208,12 @@ pub enum Command {
             help = "Enable fast forward, which is not secure when connecting to outside nodes"
         )]
         enable_fast_forward: bool,
+        #[structopt(
+            name = "disable-light-client",
+            long,
+            help = "Disable light client, which is not secure when connecting to outside nodes"
```
yihuang, line 214:
```suggestion
            help = "Disable light client, only use for trusted nodes"
```

client-core/src/wallet/syncer.rs
```diff
@@ -449,8 +450,9 @@ impl<
         }
 
         let (target_height, target_app_hash, target_block_hash) =
-            if self.env.options.enable_fast_forward {
+            if self.env.options.enable_fast_forward || self.env.options.disable_light_client {
                 (
+                    /* use api api directly */
```
linfeng-crypto, line 455: `use api directly` ?

leejw51crypto: yes, not using light client protocol

tomtau: the comment says "use **api api** directly" -- maybe "use Tendermint RPC directly" is clearer ?

leejw51crypto: that's better,i'll change

leejw51crypto:
fixing travis issue

PR #2018, opened 2020-07-24
Problem: didn't verify secrets against public key in DirectPath

yihuang:
Solution:
- verify decrypted secret against public key
- also verify leaf keypackage against leaf secret

### DIFF
--- a/chain-tx-enclave-next/mls/src/group.rs
+++ b/chain-tx-enclave-next/mls/src/group.rs
@@ -1,11 +1,13 @@
+use std::collections::BTreeMap;
 use std::convert::TryFrom;
 
 use crate::ciphersuite::*;
 use crate::extensions as ext;
-use crate::key::IdentityPublicKey;
-use crate::keypackage::Timespec;
-use crate::keypackage::PROTOCOL_VERSION_MLS10;
-use crate::keypackage::{self as kp, KeyPackage, KeyPackageSecret};
+use crate::key::{HPKEPrivateKey, IdentityPrivateKey, IdentityPublicKey};
+use crate::keypackage::{
+    self as kp, verify_keypackage_and_secrets, KeyPackage, KeyPackageSecret, Timespec,
+    PROTOCOL_VERSION_MLS10,
+};
 use crate::message::*;
 use crate::secrets::*;
 use crate::tree::*;
@@ -14,7 +16,6 @@ use crate::utils::{
     encode_vec_option_u32, encode_vec_u8_u16, encode_vec_u8_u8, read_vec_option_u32,
     read_vec_u8_u16, read_vec_u8_u8,
 };
-use hkdf::InvalidLength;
 use ra_client::AttestedCertVerifier;
 use rustls::internal::msgs::codec::{self, Codec, Reader};
 use secrecy::{ExposeSecret, SecretVec};
@@ -29,8 +30,13 @@ pub struct GroupAux {
     pub tree_secret: TreeSecret,
     pub secrets: EpochSecrets<Sha256>,
     pub kp_secret: KeyPackageSecret,
-    // record the new secret when waiting for self update commit
-    pub kp_secret_pending: Option<KeyPackageSecret>,
+
+    /// record the pending credential secret for self-update proposals.
+    /// inserted when commit self update proposal, removed when processing the commit.
+    pub pending_updates: BTreeMap<ProposalId, IdentityPrivateKey>,
+    /// record the new init private key generate in commit.
+    /// inserted when commit proposals, removed when processing self commit.
+    pub pending_commit: BTreeMap<ProposalId, HPKEPrivateKey>,
 }
 
 impl GroupAux {
@@ -47,7 +53,8 @@ impl GroupAux {
             tree_secret: TreeSecret::empty(cs),
             secrets,
             kp_secret,
-            kp_secret_pending: None,
+            pending_updates: BTreeMap::new(),
+            pending_commit: BTreeMap::new(),
         }
     }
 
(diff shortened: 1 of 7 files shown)


chain-tx-enclave-next/mls/src/group.rs
```diff
@@ -443,6 +446,37 @@ impl GroupAux {
         let commit_content = CommitContent::new(self.tree.cs, &commit, proposals)
             .map_err(|_| ProcessCommitError::CommitError)?;
 
+        // update kp_secret for self updating proposal
+        let self_update_proposal = commit_content
+            .updates
+            .iter()
+            .filter_map(|(sender, update)| {
```
leejw51crypto, line 453: can use find? because it's finding the first one?

chain-tx-enclave-next/mls/src/group.rs
```diff
@@ -835,6 +848,8 @@ pub enum ProcessCommitError {
     CommitError,
     #[error("decrypted path secret don't match the public key")]
     PathSecretPublicKeyDontMatch,
+    #[error("cached secret key don't match commit leaf keypackage")]
+    LeafSecretPublicKeyDontMatch,
```
tomtau, line 852: seems this error is not used?

yihuang: yes, let me remove it, this error case is now keypackage verify error.

chain-tx-enclave-next/mls/src/tree.rs
```diff
@@ -570,6 +570,28 @@ impl TreePublicKey {
         }
         self.nodes[path[0].node_index()].compute_parent_hash(self.cs)
     }
+
+    fn verify_node_private_key(
```
tomtau, line 574: is it ok to make this pub(crate)? I may need to use this -- just document the intended usage/args

yihuang:
Added `pending_updates` and `pending_commit`, it's more clear to handle pending private keys this way.

chain-tx-enclave-next/mls/src/tree.rs
yihuang: done

chain-tx-enclave-next/mls/src/group.rs
yihuang: done

chain-tx-enclave-next/mls/src/group.rs
yihuang: done

chain-tx-enclave-next/mls/src/tree.rs
```diff
@@ -570,6 +570,25 @@ impl TreePublicKey {
...
+            if !bool::from(
+                private_key
+                    .public_key()
+                    .marshal()
+                    .ct_eq(&node.public_key.marshal()),
+            ) {
+                false
+            } else {
+                true
+            }
```
tomtau, line 587: could this be rewritten as:
```
bool::from(private_key
                      .public_key()
                      .marshal()
                      .ct_eq(&node.public_key.marshal())
```
?

chain-tx-enclave-next/mls/src/tree.rs
```diff
@@ -714,27 +751,46 @@ impl TreeSecret {
...
+            if !tree.verify_node_private_key(&overlap_secret, overlap) {
+                return Err(ProcessCommitError::PathSecretPublicKeyDontMatch);
+            }
+            for (secret, &parent) in secrets.iter().zip(overlap_path) {
+                if !tree.verify_node_private_key(secret, parent) {
+                    return Err(ProcessCommitError::PathSecretPublicKeyDontMatch);
+                }
+            }
+
+            self.path_secrets.truncate(overlap_pos);
```
tomtau, line 790: was a bit puzzled about `truncate` here -- I guess the meaning is that it'll keep secrets up to the overlap start (as they are unchanged), and then replace with was decrypted / computed?
a comment could be good here

chain-tx-enclave-next/mls/src/tree.rs
yihuang: yes, nice catch

chain-tx-enclave-next/mls/src/tree.rs
yihuang: yes, added a comment

yihuang:
> looks all right, but would be good to add a test where the encrypted secret doesn't match, so `ProcessCommitError` is correctly returned (and group state etc. wasn't changed)

Another issue is, for multiple pending commits, there'll be only one is valid, the pending secrets for other commits will not cleaned up. don't know when or how to cleanup the `pending_updates`/`pending_commits` of unconfirmed commits currently.

tomtau:
> > looks all right, but would be good to add a test where the encrypted secret doesn't match, so `ProcessCommitError` is correctly returned (and group state etc. wasn't changed)
> 
> Another issue is, for multiple pending commits, there'll be only one is valid, the pending secrets for other commits will not cleaned up. don't know when or how to cleanup the `pending_updates`/`pending_commits` of unconfirmed commits currently.

cleanup after applying the "committed" commit (block time when it was included + NACK timeout)?

yihuang:
> > > looks all right, but would be good to add a test where the encrypted secret doesn't match, so `ProcessCommitError` is correctly returned (and group state etc. wasn't changed)
> > 
> > 
> > Another issue is, for multiple pending commits, there'll be only one is valid, the pending secrets for other commits will not cleaned up. don't know when or how to cleanup the `pending_updates`/`pending_commits` of unconfirmed commits currently.
> 
> cleanup after applying the "committed" commit (block time when it was included + NACK timeout)?

I guess just cleanup all the pending secrets after processing any commit, because all the pending commits will be invalid anyway(the epoch is old)?

tomtau:
needs to be documented https://github.com/crypto-com/chain-docs/issues/141#issuecomment-662813075

in here https://github.com/crypto-com/chain-docs/blob/master/docs/modules/tdbe.md#abci-processing-sketch , it'll be in transitions 8. and 10.

tomtau:
> I guess just remove all the pending secrets after processing any commit, because all the pending commits will be invalid anyway(the epoch is old)?
Yes, I think so -- if e.g. you happen to generate self update proposal while there was another on-going update/removal, it'll be rejected... so you'll need to re-generate and re-submit.

yihuang:
Fixed partial mutation when process invalid commit.

PR #2030, opened 2020-07-27
Problem: (fix #1951) no certificate expiration handling tests

### LINKED ISSUE #1951: Problem: no certificate expiration handling tests
could use https://edp.fortanix.com/docs/tasks/testing/ (may be necessary to mock aesm / some ra-enclave stuff)

test:
- generate a cert that expires soon
- verify that after expiration, the tx-query / ra-enclave ... code correctly refreshes / generates a new certificate

--

another option:
- make the validity configurable via a compile-time parameter
- add an integration test with a short validity time -- check cert gets refreshed

linfeng-crypto:
solution:
* make certificate expiration as a compile-time parameter
* set default validity_duration to 90 days, which is as same as certificate expiration
* add a small time duration when check certificate is expired or not
* use the min(default validity_duration, cert_expiration) as the validity_duration to check the certificate is expired or not

### DIFF
--- a/chain-tx-enclave-next/enclave-ra/ra-enclave/src/certificate.rs
+++ b/chain-tx-enclave-next/enclave-ra/ra-enclave/src/certificate.rs
@@ -16,7 +16,13 @@ impl Certificate {
     /// Checks if current certificate is valid or not
     pub fn is_valid(&self, validity_duration: Duration) -> bool {
         let current_time = Utc::now();
-        self.created + validity_duration >= current_time
+        // add extra 15 seconds offset to avoid the impact of network transfer delay
+        // and difference of the os system time
+        if validity_duration > Duration::seconds(15) {
+            current_time - self.created <= validity_duration - Duration::seconds(15)
+        } else {
+            current_time - self.created < validity_duration
+        }
     }
 
     /// Sets current certificate in given `rustls` server config
--- a/chain-tx-enclave-next/enclave-ra/ra-enclave/src/context.rs
+++ b/chain-tx-enclave-next/enclave-ra/ra-enclave/src/context.rs
@@ -19,11 +19,15 @@ use crate::{
     config::EnclaveRaConfig,
 };
 
+/// 90 days
+pub const DEFAULT_EXPIRATION_SECS: i64 = 7776000;
+
 /// Wraps all the in-enclave operations required for remote attestation
 pub struct EnclaveRaContext {
     certificate: Arc<Mutex<Option<Certificate>>>,
     sp_ra_client: SpRaClient,
     validity_duration: Duration,
+    expiration_duration: Duration,
 }
 
 impl EnclaveRaContext {
@@ -31,11 +35,15 @@ impl EnclaveRaContext {
     pub fn new(config: &EnclaveRaConfig) -> Result<Self, EnclaveRaContextError> {
         let sp_ra_client = SpRaClient::connect(&config.sp_addr)?;
         let validity_duration = Duration::seconds(config.certificate_validity_secs.into());
+        let expiration_duration = config
+            .certificate_expiration_time
+            .unwrap_or_else(|| Duration::seconds(DEFAULT_EXPIRATION_SECS));
 
         Ok(Self {
             certificate: Default::default(),
             sp_ra_client,
             validity_duration,
+            expiration_duration,
         })
     }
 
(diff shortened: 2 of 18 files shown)


chain-tx-enclave-next/enclave-ra/ra-enclave/src/context.rs
```diff
@@ -40,7 +42,10 @@ impl EnclaveRaContext {
     }
 
     /// Returns current certificate. If current certificate is no longer valid, then it creates a new one
-    pub fn get_certificate(&self) -> Result<Certificate, EnclaveRaContextError> {
+    pub fn get_certificate(
+        &self,
+        expiration_time: Option<Duration>,
```
devashishdxt, line 47: Can you add `certificate_expiration_time` in `EnclaveRaConfig` and set it in a struct member instead of taking this as function argument (similar to `validity_duration`)?

integration-tests/pytests/test_client_cli.py
```diff
@@ -223,7 +223,8 @@ def test_deposit_to_other_wallet():
 
 @pytest.mark.zerofee
 def test_withdraw_to_other_wallet():
-    time.sleep(3)
+    # the tx-query cert_validation is 200 (setted in .drone.yaml), check the cert refresh or not
+    time.sleep(210)
```
tomtau, line 227: how is this checked / asserted? If it's "implicitly" tested that one can connect, it can still be the case that the cert has a long validity, so nothing was really tested -- it's preferable to have a dedicated test where one would dump and check the certificate (`openssl s_client -connect <tx-query-host:port> -state -debug`  equivalent in python)

chain-tx-enclave-next/enclave-ra/ra-enclave/src/certificate.rs
```diff
@@ -16,7 +16,8 @@ impl Certificate {
     /// Checks if current certificate is valid or not
     pub fn is_valid(&self, validity_duration: Duration) -> bool {
         let current_time = Utc::now();
-        self.created + validity_duration >= current_time
+        // add extra 15 seconds here to avoid the impact of network transfer delay
+        current_time - self.created <= validity_duration - Duration::seconds(15)
```
tomtau, line 20: can this underflow?

leejw51crypto: possible, how about adding underflow check above

chain-tx-enclave-next/enclave-ra/ra-enclave/src/context.rs
```diff
@@ -49,7 +54,7 @@ impl EnclaveRaContext {
         };
 
         if needs_creating {
-            let new_certificate = self.create_certificate();
+            let new_certificate = self.create_certificate(expiration_time);
```
leejw51crypto, line 57: how about `expire_time_in_seconds` ?

chain-tx-enclave-next/enclave-ra/ra-enclave/src/context.rs
```diff
@@ -148,15 +156,20 @@ impl EnclaveRaContext {
     }
 
     /// Creates new certificate params
-    fn create_certificate_params(&self) -> Result<CertificateParams, EnclaveRaContextError> {
+    fn create_certificate_params(
+        &self,
+        expiration_time: Option<Duration>,
+    ) -> Result<CertificateParams, EnclaveRaContextError> {
+        let expiration_time =
+            expiration_time.unwrap_or_else(|| Duration::days(DEFAULT_EXPIRATION_DAYS));
```
leejw51crypto, line 164: DEFAULT_EXPIRATION_DAYS  
adding unit to this name or comment about the `unit`?

chain-tx-enclave-next/tx-query-next/enclave-app/src/sgx_module.rs
```diff
@@ -15,24 +15,39 @@ use thread_pool::ThreadPool;
...
     log::info!("Connecting to ZeroMQ");
     let zmq_stream = Arc::new(Mutex::new(TcpStream::connect("zmq")?));
 
     let num_threads = 4;
+
+    // use the smaller Duration as certificate validity, so that we can check the certification is expired or not correctly
+    let default_expiration_time = Duration::days(DEFAULT_EXPIRATION_DAYS);
+    let certificate_validity = match cert_expiration {
+        None => default_expiration_time,
+        Some(s) => {
```
leejw51crypto, line 40: can be refactored
s= std::min(default_expiration_time,s);

tomtau:
bors try

PR #2033, opened 2020-07-28
Problem: no paging in client-rpc, address_list (fix #2023)

### LINKED ISSUE #2023: Problem: no paging in client-rpc, address_list
add paging (offset, limit) to 
wallet_listTransferAddresses,
wallet_listStakingAddresses,

because addresses are millions, rpc call hangs

leejw51crypto:
add paging to client-rpc, and cli
this is breaking change to current protocol

also check_staking_address_for_deposit is refactored ( get rid of staked addresses copying)

with this pr, after 4,000,000 addresses, query address is very fast.
previously it fetches all, it hangs `postman`

### DIFF
--- a/client-cli/src/command/transaction_command.rs
+++ b/client-cli/src/command/transaction_command.rs
@@ -645,14 +645,21 @@ fn check_staking_address_for_deposit<T: WalletClient, N: NetworkOpsClient>(
     address: &StakedStateAddress,
 ) -> Result<()> {
     // if the to_address belongs to current wallet, we do not check the state
-    let staking_addresses = wallet_client.staking_addresses(name, enckey)?;
-    if !staking_addresses.contains(address) {
-        let staking = network_ops.get_staked_state(name, address, true).err_kind(ErrorKind::ValidationError,|| "Address not found in the current wallet and is not yet initialized on the blockchain")?;
-        if staking.is_jailed() {
-            return Err(Error::new(
-                ErrorKind::ValidationError,
-                "staking address is jailed",
-            ));
+    match address {
+        StakedStateAddress::BasicRedeem(ref redeem_address) => {
+            // if to_address doesn't belong to current wallet, we check the state
+            if wallet_client
+                .find_staking_key(name, enckey, redeem_address)?
+                .is_none()
+            {
+                let staking = network_ops.get_staked_state(name, address, true).err_kind(ErrorKind::ValidationError,|| "Address not found in the current wallet and is not yet initialized on the blockchain")?;
+                if staking.is_jailed() {
+                    return Err(Error::new(
+                        ErrorKind::ValidationError,
+                        "staking address is jailed",
+                    ));
+                }
+            }
         }
     }
     Ok(())
--- a/client-core/src/service/wallet_service.rs
+++ b/client-core/src/service/wallet_service.rs
@@ -531,6 +531,29 @@ pub fn load_wallet<S: SecureStorage + 'static>(
     }
     Ok(None)
 }
+
+fn generate_page(all_items_count: u64, offset: u64, limit: u64, reversed: bool) -> Vec<u64> {
+    let ret: Vec<u64>;
+    if reversed {
+        let start = all_items_count.saturating_sub(offset);
+        let mut end = 0;
+        if limit > 0 {
+            end = start.saturating_sub(limit);
+        }
+        let mut tmp: Vec<u64> = (end..start).collect();
+        tmp.reverse();
+        ret = tmp;
+    } else {
+        let start = offset;
+        let mut end = all_items_count;
+        if limit > 0 {
+            end = std::cmp::min(offset + limit, all_items_count);
+        }
+        ret = (start..end).collect();
+    }
+    ret
+}
+
 /// Maintains mapping `wallet-name -> wallet-details`
 #[derive(Debug, Default, Clone)]
 pub struct WalletService<T: Storage> {
@@ -618,7 +641,6 @@ where
         redeem_address: &RedeemAddress,
     ) -> Result<Option<PublicKey>> {
         let stakingkey_keyspace = get_stakingkeyset_keyspace(name);
-
         if let Ok(value) = read_pubkey(
             &self.storage,
             &stakingkey_keyspace,
@@ -626,10 +648,7 @@ where
         ) {
             Ok(Some(value))
         } else {
-            Err(Error::new(
-                ErrorKind::InvalidInput,
-                "Address is not found in current wallet",
-            ))
+            Ok(None)
         }
     }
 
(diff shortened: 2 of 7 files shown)


leejw51crypto:
ok, i'll fix it

leejw51crypto:
inspecting travis issue

client-cli/src/command/transaction_command.rs
```diff
@@ -645,16 +645,24 @@ fn check_staking_address_for_deposit<T: WalletClient, N: NetworkOpsClient>(
...
-            return Err(Error::new(
-                ErrorKind::ValidationError,
-                "staking address is jailed",
-            ));
+    match address {
+        StakedStateAddress::BasicRedeem(ref redeem_address) => {
+            // if to_address doesn't belong to current wallet, we check the state
+            if wallet_client
+                .find_staking_key(name, enckey, redeem_address)
+                .is_err()
```
tomtau, line 653: Are all the possible errors meaning that redeem address doesn't belong to the wallet? For some errors (e.g. can't access the storage), the logic may not be correct?

client-core/src/service/wallet_service.rs
```diff
@@ -774,9 +781,36 @@ where
...
+            let mut i = start;
+            while i >= end {
+                let pubkey = read_pubkey(&self.storage, &stakingkey_keyspace, &format!("{}", i))?;
+                ret.insert(pubkey);
+                if 0 == i {
+                    break;
+                }
+                // iterate
+                i -= 1;
+            }
```
tomtau, line 803: I'm thinking if this could be simplified with something like:
```
let start = std::cmp::max(0, staking_count.saturating_sub(1).saturating_sub(offset));
let end = std::cmp::max(0, start.saturating_sub(limit.saturating_sub(1));
(start..end).map(|i| read_pubkey(&self.storage, &stakingkey_keyspace, &format!("{}", i))?).collect()
```

client-core/src/service/wallet_service.rs
```diff
@@ -810,12 +855,44 @@ where
...
+            for i in start..end {
+                let value = self.storage.get(&roothash_keyspace, format!("{}", i))?;
+                if let Some(raw_value) = value {
+                    let mut roothash_found: H256 = H256::default();
+                    roothash_found.copy_from_slice(&raw_value);
+                    ret.insert(roothash_found);
+                }
             }
         }
```
tomtau, line 897: seems this code is exactly the same as with staking_keys (the only difference is the key space) -- perhaps can be extracted out to a helper function used by both?

integration-tests/bot/chainrpc.py
```diff
@@ -117,7 +117,7 @@ def list(self, name=DEFAULT_WALLET, type='staking', enckey=None):
         '''list addresses
         :param name: Name of the wallet. [default: Default]
         :params type: [staking|transfer]'''
-        return self.client.call('wallet_listStakingAddresses' if type == 'staking' else 'wallet_listTransferAddresses', [name, enckey or get_enckey()])
+        return self.client.call('wallet_listStakingAddresses' if type == 'staking' else 'wallet_listTransferAddresses', [name, enckey or get_enckey()],0,0,False)
```
tomtau, line 120: do these parameters now need to be provided? could in the API be put as `Option`, so if they are not provided, it'll just take a default value?

client-cli/src/command/transaction_command.rs
leejw51crypto: agree, i'll fix it

client-core/src/service/wallet_service.rs
leejw51crypto: looking good to me, i'll replace

client-core/src/service/wallet_service.rs
leejw51crypto: ok

integration-tests/bot/chainrpc.py
leejw51crypto: agree, i'll try to use Option for default value

client-cli/src/command/transaction_command.rs
leejw51crypto: 1. for default value, i'll try with  using Option
2. for limit, how about maximum of 1000 ?

leejw51crypto:
maximum entries is 10,000, which will be enough for general users

now fixing unit, integration,
it will be removed (limit, offset, reversed)

leejw51crypto:
working on 
0,0,false to fetch all

leejw51crypto:
what's bors messsage?

tomtau:
@leejw51crypto client-cli integration test fails
```
=================================== FAILURES ===================================
1314	_________________________ test_deposit_to_other_wallet _________________________
1315	
1316	    @pytest.mark.zerofee
1317	    def test_deposit_to_other_wallet():
1318	        # test deposit to other wallet staking address
1319	        wallet_2.sync()
1320	        state1 = wallet_2.state(staking_address_wallet_1)
1321	>       deposit_transaction(wallet_2, staking_address_wallet_1, 20000)
1322	
1323	pytests/test_client_cli.py:219: 
1324	_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
1325	pytests/test_client_cli.py:126: in deposit_transaction
1326	    tx.deposit(staking_address, amount_cro)
1327	_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
1328	
1329	self = <client_cli.Transaction object at 0x7f17e0d71460>
1330	to_staking_address = '0x5e7e1e79d80b861a94598c721598951098dd3825'
1331	amount = 20000
1332	
1333	    def deposit(self, to_staking_address, amount):
1334	        if self.hardware:
1335	            cmd = ["client-cli", "transaction",  "--hardware", self.hardware, "new", "-t", "deposit", "-n", self.wallet.name]
1336	        else:
1337	            cmd = ["client-cli", "transaction", "new", "-t", "deposit", "-n", self.wallet.name]
1338	        args = [self.wallet.auth_token, to_staking_address, str(amount), "Y"]
1339	        text = run(cmd, args)
1340	        if "Error" in text:
1341	>           raise Exception(text)
1342	E           Exception: Enter authentication token: Enter staking address: Invalid input: Address is not found in current wallet
1343	E           Are you sure to proceed? [yN]Error: Invalid input:
1344	
1345	bot/client_cli.py:355: Exception
1346	============ 1 failed, 12 passed, 1 deselected in 126.00s (0:02:05) ============
```

leejw51crypto:
thanks, i'm checking the issue

leejw51crypto:
found the cause, fixing now

tomtau:
bors retry

PR #2036, opened 2020-07-28
Problem: un-necessary WalletState writing to DB (fix #2032)

### LINKED ISSUE #2032: Problem: un-necessary WalletState writing to DB
```
[2020-07-28T04:20:55Z DEBUG client_core::service::wallet_state_service] modify_wallet_state name a
[2020-07-28T04:20:55Z DEBUG client_core::service::wallet_state_service] fetch_and_update_secure name a
```

when there is no memento, don't call modify_wallet_state

it will save a lot of time, when there are a lot of utxos in the wallet

leejw51crypto:
Solution: pre-check memento before fetch & update
it will do writing when there is change(it's maximum 2 per block), so speed will be OK

in case , a lot of utxo updates in the block, just one group of mementos will be applied at once.

### DIFF
--- a/client-core/src/service/wallet_state_service.rs
+++ b/client-core/src/service/wallet_state_service.rs
@@ -411,6 +411,18 @@ enum MementoOperation {
 }
 
 impl WalletStateMemento {
+    /// get length of memento
+    #[inline]
+    pub fn len(&self) -> usize {
+        self.0.len()
+    }
+
+    /// is memento empty?
+    #[inline]
+    pub fn is_empty(&self) -> bool {
+        self.0.is_empty()
+    }
+
     /// Adds transaction change to memento
     #[inline]
     pub fn add_transaction_change(&mut self, transaction_change: TransactionChange) {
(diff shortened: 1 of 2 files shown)


client-core/src/service/wallet_state_service.rs
```diff
@@ -411,6 +411,12 @@ enum MementoOperation {
 }
 
 impl WalletStateMemento {
+    /// get length of memento
+    #[inline]
+    pub fn get_length(&self) -> usize {
```
yihuang, line 416:
```suggestion
    pub fn len(&self) -> usize {
```

keep the naming convention with std lib?

devashishdxt: Clippy will also complain about `is_empty`.

leejw51crypto: ok, i'll change to len
and fix clippy issue

leejw51crypto:
working on travis issue now

tomtau:
bors retry

PR #2058, opened 2020-07-31
Problem: (fix #2057)makefile cannot handle /dev/sgx and /dev/isgx coe…

### LINKED ISSUE #2057: Problem: makefile cannot handle /dev/sgx and /dev/isgx coexist situation
makefile should add logic to alert if user's dcap driver is higher v1.31 as it is no longer supported by fortanix lib.
and allow user to keep using dcap driver if it is not higher v1.31.

allthatjazzleo:
Solution:
        1. Makefile alert if user's dcap driver is higher v1.31 as it is no longer supported by fortanix lib.
and allow user to keep using dcap driver if it is not higher v1.31.
        2. Makefile choose /dev/isgx first if exist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,13 +10,16 @@ build_mode ?= debug
 TX_QUERY_HOSTNAME ?=
 MAKE_CMD = make
 
-# SGX_DEVICE should be /dev/isgx as latest dcap driver /dev/sgx is not supported by fortanix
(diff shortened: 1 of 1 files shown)


tomtau:
bors retry

tomtau:
bors retry

PR #2068, opened 2020-08-04
Problem: light client witness nodes can't be configured in client-* (…

leejw51crypto:
Soution: add witness, peer configuration
option example) 
--light-client-peers "429......02665@10.11.12.13:26657,96663a3....21c693@10.11.12.14:26657"
first will be primary, others witnesses

### DIFF
--- a/client-core/src/wallet/syncer.rs
+++ b/client-core/src/wallet/syncer.rs
@@ -10,6 +10,7 @@ use std::str::FromStr;
 use std::sync::Arc;
 use std::thread;
 use std::time::Duration;
+use tendermint_light_client::peer_list::PeerListBuilder;
 pub use tendermint_light_client::supervisor::Handle;
 use tendermint_light_client::{
     components::{
@@ -22,7 +23,6 @@ use tendermint_light_client::{
     fork_detector::ProdForkDetector,
     light_client::{self, LightClient},
     operations::hasher::{Hasher, ProdHasher},
-    peer_list::PeerList,
     state::State,
     store::{sled::SledStore, LightStore},
     supervisor::{Instance, Supervisor},
@@ -112,6 +112,7 @@ pub struct SyncerOptions {
     pub enable_address_recovery: bool,
     pub batch_size: usize,
     pub block_height_ensure: u64,
+    pub light_client_peers: String,
 }
 
 /// Common configs for wallet syncer with `TransactionObfuscation`
@@ -1045,6 +1046,7 @@ mod tests {
                     enable_address_recovery: false,
                     batch_size: 20,
                     block_height_ensure: 50,
+                    light_client_peers: "".into(),
                 },
             },
             |_txids: &[TxId]| -> Result<Vec<Transaction>> { Ok(vec![]) },
@@ -1168,6 +1170,7 @@ mod tests {
                     enable_address_recovery: false,
                     batch_size: 20,
                     block_height_ensure: 50,
+                    light_client_peers: "".into(),
                 },
             },
             |_txids: &[TxId]| -> Result<Vec<Transaction>> { Ok(vec![]) },
@@ -1221,6 +1224,7 @@ mod tests {
                     enable_address_recovery: true,
                     batch_size: 20,
                     block_height_ensure: 50,
+                    light_client_peers: "".into(),
                 },
             },
             |_txids: &[TxId]| -> Result<Vec<Transaction>> { Ok(vec![]) },
@@ -1281,6 +1285,7 @@ mod tests {
                     enable_address_recovery: true,
                     batch_size: 20,
                     block_height_ensure: 50,
+                    light_client_peers: "".into(),
                 },
             },
             |_txids: &[TxId]| -> Result<Vec<Transaction>> { Ok(vec![]) },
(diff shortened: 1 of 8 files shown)


tomtau:
bors try

leejw51crypto:
i'll check

leejw51crypto:
writing unit test

yihuang:
I think it's supposed to be one primary node and multiple witness nodes.
So maybe only one parameter is enough: `--light-client-peers node0,node1,node2`, node0 is primary, the other is witnesses.
Also use node0 to replace the old `CRYPTO_CLIENT_TENDERMINT` parameter.

leejw51crypto:
ok, that looks good
i'll modify

leejw51crypto:
modified , and testing now

tomtau:
bors try

leejw51crypto:
modifying integration test

leejw51crypto:
rebased

tomtau:
bors try

leejw51crypto:
ok, i'll modify

client-core/src/wallet/syncer.rs
```diff
@@ -1337,50 +1342,135 @@ mod tests {
...
-    let addr = tendermint::net::Address::from_str(&addr).unwrap();
-
-    // FIXME use actual tendermint node id as peer id
-    let primary: PeerId = "BADFADAD0BEFEEDC0C0ADEADBEEFC0FFEEFACADE".parse().unwrap();
-    let witness: PeerId = "CEFEEDBADFADAD0C0CEEFACADE0ADEADBEEFC0FF".parse().unwrap();
 
-    let primary_path = db_path.join(primary.to_string());
-    let witness_path = db_path.join(witness.to_string());
+    if "" == light_client_peers {
+        log::info!("use default primary peer {}", default_primary);
```
tomtau, line 1417: maybe `warn!` -- or should this just reject? I don't know if using a wrong Peer ID could cause any issues later on (e.g. if one connects to the same address, but uses a different/correct Peer ID)?

Could the Peer ID be fetched from TM RPC before this? e.g. in "status", you get "node_info" where it has "id"?

client-core/src/wallet/syncer.rs
```diff
@@ -1337,50 +1342,135 @@ mod tests {
...
+    let all_peers_list = peerlist_builder.build();
     let mut supervisor = Supervisor::new(
-        peer_list,
+        all_peers_list,
         ProdForkDetector::default(),
-        ProdEvidenceReporter::new(peer_addr),
+        ProdEvidenceReporter::new(all_peers_map),
     );
-    let handle = supervisor.handle();
+    // handle is supervisor lient client implementation
```
tomtau, line 1469:
```suggestion
    // handle is supervisor light client implementation
```

client-core/src/wallet/syncer.rs
tomtau: https://docs.tendermint.com/master/rpc/#/Info/status

leejw51crypto: how about reject?
i added default for passing integration test.
because peer id  is to tell between peers, the same address will be ignoreded (from tenderment manual)

leejw51crypto:
> I'm not sure if we can directly use `light_client_peers: Vec<String>` in `Command`. `structopt` supports multiple values using a `Vec` ([documentation](https://docs.rs/structopt/0.3.16/structopt/#type-magic)).

that would be possible, but this pattern is the same with that of tendermint, one arg is easier , i think

leejw51crypto:
fixed clippy issue,

working on `removing default` and fix integration test

client-core/src/wallet/syncer.rs
tomtau: Inside the light client, peerid seems to be only used for storage indexing -- @yihuang  ?
with one fixed default ID, I guess one can corrupt the storage by mashing up blocks from different peers? or for multiple witnesses, if one is faulty, it'll consider all of them faulty?

maybe reject is a correct behaviour?

leejw51crypto: i'll add rejection if
0000000@127.0.0.1,
0000000@127.1.2.3,
IDs overlap

leejw51crypto:
working python script for integration test,
after will add "redundant peerid" check,

if redundant peerid found, it will give error

leejw51crypto:
done

tomtau:
bors try

client-core/src/wallet/syncer.rs
```diff
@@ -1337,50 +1342,115 @@ mod tests {
...
+    }
+    Ok(all_nodes)
+}
+
 /// [new light client design](https://github.com/informalsystems/tendermint-rs/blob/master/docs/architecture/adr-006-light-client-refactor.md)
 pub fn spawn_light_client_supervisor(
     db_path: &Path,
-    addr: &str,
+    // ws://localhost:26657/websocket
+    _websocket_addr: &str,
```
tomtau, line 1380: this argument is not necessary?

leejw51crypto: now it's not necessary, because , we don't use `default` values,
i'll remove it

leejw51crypto:
rebased

Issue #2090, opened 2020-08-10
Problem: un-necessary light-client-peers option with --disable-light-client


leejw51crypto:
if we don't use light-client,  it's not necessary to specify light-client-peers option

leejw51crypto:
started

leejw51crypto:
done, checking with integration test

leejw51crypto:
it's merged into issue 2005, pr is submitted

leejw51crypto:
it's merged

PR #2095, opened 2020-08-10
Problem: (fix #2094)client-rpc panic without specify option --light-c…

### LINKED ISSUE #2094: Problem: client-rpc panic without specify option --light-client-peers
```bash
Network type Devnet id AB
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Invalid input: there is no light-client peers.
specify at least two peers, use light-client-peers option', client-rpc/server/src/program.rs:103:5
```
Should add --light-client-peers to have temporary fix in Makefile until it is solved in #2090

Also, `set-genesis-fingerprint` target in Makefile rely on libzmq3-dev installed to compute genesis-fingerprint.

allthatjazzleo:
Solution: 
1. add --light-client-peers to have temporary fix in Makefile
2. install necessary libzmq3-dev to run set-genesis-fingerprint target

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -182,7 +182,7 @@ endif
 
 # build the chain binary in docker
 build-chain:
-	@if [ -e "./target/${BUILD_MODE}/client-cli" ] && [ -e "./target/${BUILD_MODE}/chain-abci" ] && [ -e "./target/${BUILD_MODE}/client-rpc" ] && [ -e "./target/${BUILD_MODE}/dev-utils" ]; then \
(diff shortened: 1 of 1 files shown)


tomtau:
The real problem is this one: https://github.com/crypto-com/chain/issues/2090
so I'll close this PR / issue as it won't be relevant

tomtau:
@allthatjazzleo if https://github.com/crypto-com/chain/issues/2090 blocks you, we can try to prioritise it

allthatjazzleo:
it just block the daily deployment but this pr also fix another issue that `CRYPTO_GENESIS_FINGERPRINT` cannot be calculated without ibzmq3-dev installed

PR #2101, opened 2020-08-11
Problem: client-rpc doen't flush sled database, possibly causes lost data (fix #2100)

### LINKED ISSUE #2100: Problem: client-rpc doen't flush sled database, possibly causes lost data
In recent sled database, it doesn't flush automatically,
in normal shutdown, `drop` is called, so OK

but like client-rpc, not graceful close,
some data will be lost
so 
1. call flush after dbcalls
2. add polling function which does flushing periodically

leejw51crypto:
Solution: add flushing

in recent sled, flushing is not working as previous version.
flushes on drop only.
so added manual flushing to `db writing` functions
(sled Config.flush_every_ms didn't work)

### DIFF
--- a/client-core/src/wallet/syncer.rs
+++ b/client-core/src/wallet/syncer.rs
@@ -364,6 +364,7 @@ impl<
     fn save(&mut self, memento: &WalletStateMemento) -> Result<()> {
         service::save_sync_state(&self.env.storage, &self.env.name, &self.sync_state)?;
         self.update_state(memento)?;
+        self.env.storage.flush()?;
         Ok(())
     }
 
--- a/client-rpc/src/handler.rs
+++ b/client-rpc/src/handler.rs
@@ -1,10 +1,13 @@
 use jsonrpc_core::IoHandler;
 
+#[cfg(feature = "experimental")]
+use crate::rpc::multisig_rpc::{MultiSigRpc, MultiSigRpcImpl};
 use chain_core::tx::fee::FeeAlgorithm;
 use client_common::cipher::TransactionObfuscation;
 use client_common::storage::SledStorage;
 use client_common::tendermint::{types::GenesisExt, Client, WebsocketRpcClient};
 use client_common::Result;
+use client_common::Storage;
 use client_core::service::HwKeyService;
 use client_core::signer::WalletSignerManager;
 use client_core::transaction_builder::DefaultWalletTransactionBuilder;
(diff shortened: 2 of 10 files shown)


leejw51crypto:
ok, i'll check clippy issue

leejw51crypto:
testing sled 0.34.2

client-rpc/src/handler.rs
```diff
@@ -49,6 +49,16 @@ impl RpcHandler {
...
         let storage = SledStorage::new(&storage_dir)?;
+
+        let polling_storage = storage.clone();
+        std::thread::spawn(move || {
+            loop {
+                polling_storage.flush().expect("sled storage flush");
+                // every 1 second
+                std::thread::sleep(std::time::Duration::from_secs(1));
+            }
+        });
```
tomtau, line 60: sled has this config: https://github.com/spacejam/sled/blob/master/src/config.rs#L394

that doesn't work?

tomtau: Oh I see the commit comment
> (sled Config.flush_every_ms didn't work)

leejw51crypto: strange, that should work, i'll try again

leejw51crypto:
added `flushing periodically every 1 second`,
so if , flushing is not done,  data will be written safely to the disk

tomtau:
bors try

tomtau:
@leejw51crypto can you change the commit message to reflect the updated issue name?

client-core/src/wallet/syncer.rs
```diff
@@ -597,6 +597,7 @@ impl<
                 self.sync_state.last_block_hash = block.block_hash.clone();
 
                 self.update_progress(block.block_height);
+                self.env.storage.flush()?;
```
yihuang, line 600: Should it be put in the `save` method?

leejw51crypto: looking good, i'll check

leejw51crypto:
i'll check integration test

tomtau:
```
Traceback (most recent call last):
630	  File "./multinode/reward_test.py", line 45, in <module>
631	    assert int(state['bonded']) == init_bonded + minted // 2
632	AssertionError
```

it seems it may be some timing issue in the rewards test ? @yihuang

tomtau:
bors retry

tomtau:
bors retry

PR #2115, opened 2020-08-12
Problem: client-rpc server cannot compile with "experimental" feature

calvinlauco:
Solution: Add feature to client-rpc server

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,8 @@ endif
 
 
 chain_abci_features ?= ${CHAIN_ABCI_FEATURES}
-client_features ?= ${CLIENT_FEATURES}
+client_cli_features ?= ${CLIENT_CLI_FEATURES}
+client_rpc_features ?= ${CLIENT_RPC_FEATURES}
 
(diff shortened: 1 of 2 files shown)


calvinlauco:
> @calvinaco @calvinlauco I think it'll still fail to compile, because the experimental feature-guard wasn't maintained in client-core
Oh, I opened this issue because I realize I cannot compile a ClientRPC with multi-signature support. Is the `experimental` feature-guard deprecated or will be updated in the future?

tomtau:
> Oh, I opened this issue because I realize I cannot compile a ClientRPC with multi-signature support. Is the `experimental` feature-guard deprecated or will be updated in the future?

it's kind of abandoned for now, but it will be revisited. At this moment, multi-signature support is very unstable: 1) API itself is still in flux and subject to change, 2) there's a very recent new development of a **_safe_** 2-round variant: https://slides.com/real-or-random/taproot-and-schnorr-multisig#/23/0/6  -- so it may better to revisit that later

PR #2123, opened 2020-08-15
Problem (Fix #1785): unrestricted byte vector sizes in mls modules

### LINKED ISSUE #1785: Problem: unrestricted byte vector sizes in mls modules
instead of it being dynamically checked, the size can be checked statically (for hash outputs, secret lengths, nonce lengths...) -- e.g. with Rust array or GenericArray

yihuang:
Solution:
- refactor ciphersuite to trait
- make many other data types generic over `CipherSuite` trait
- restructure code a bit to make the internal dependency less cyclic

---
This change is [Reviewable](https://reviewable.io/reviews/crypto-com/chain/2123)


yihuang:
make the dependency between internal modules less interwined, currently internal modules split into follow categories, the bottom one depend on top one, but not vise-versa.

- utils, tree_math
- ciphersuite, credential, key, secrets
- keypackage, extensions, error, message
- crypto, tree, group, extra

tomtau:
I had a quick look -- the diff is a bit big, so I had to visualise in https://reviewable.io/reviews/crypto-com/chain/2123 and left a few comments there

Issue #2051, opened 2020-07-30
Problem: indirect communication between tx-validation and tx-query


tomtau:
can open unix-domain socket -- expose via usercallextension to both tx-validation (TVE) and tx-query (TQE) enclaves.
TVE / TQE code for TLS can reuse the implemented enclave-to-enclave comms from TDBE, but needs a change that MRENCLAVE measurement of TQE is a compile-time parameter for TVE https://github.com/crypto-com/chain-docs/blob/master/docs/modules/tdbe.md#svn-verification--compilation-order

TQE->chain-abci: for fetching / decrypting, it can have "read only storage" direct requests to chain-abci / usercall extension with some request-reply buffer

leejw51crypto:
started

tomtau:
@leejw51crypto @devashishdxt is now working on https://github.com/crypto-com/chain/issues/1994 which has some overlap/similarity with this issue (attested TLS, compile-time MRENCLAVE and checking measurements, unix socket pair passed via `UsercallExtension` methods...), so you can sync up for any common intermediate PRs (e.g. changes to the build process)

leejw51crypto:
ok, i'll refer to #1994 pr,
using common code-base already done

leejw51crypto:
start coding , basic unix socket first

leejw51crypto:
made communication structure, and protocol
now working on replace current code


tx_query -> abci : call tx-validation here..    
========>
tx_query->abci : read storage prepare tx
tx_query->tx_validation: encrypt

tomtau:
discussed with @leejw51crypto on slack -- it may not be worth putting a direct communication channel between TQE and TVE right now, as it still needs a storage access (for looking up staked state and sealed transactions from chain-abci).

The code written so far can still be reused/repurposed for:
https://github.com/crypto-com/chain/issues/1994

Issue #1995, opened 2020-07-23
Problem: no persistence of TDBE-fetched transaction payloads


tomtau:
this needs a doc: https://github.com/crypto-com/chain-docs/issues/195
it can start off with the "simple" / full-history (no utreexo) fetching and no abci state sync.
tdbe still should do a basic validation fo the fetched data (txid matches, ...)

in particular, chain-abci would contain the "prefetched" sealed transactions, so it needs to store:
1) block height to which it prefetched
2) which sealed tx / txid was prefetched

when tendermint replays blocks to chain-abci, it can just do a simplified validation + create/remove bitvectors to keep track of which outputs are spent UP to the block height it prefetched.

And before chain-abci runs ABCI protocol, it needs to start up TDBE and have some way (user call extension) for it to send back that info above and signal when it's finished -- https://github.com/crypto-com/chain/issues/1993

Issue #1994, opened 2020-07-23
Problem: no tx-validation enclave attested TLS listener for receiving secrets from TDBE


tomtau:
the mechanism would be similar to tx-query -> tx-validation mutually attested TLS https://github.com/crypto-com/chain/issues/1993
https://github.com/crypto-com/chain-docs/blob/master/docs/modules/tdbe.md#new-obfuscation-key

before the MLS handshake processing and message generation etc. is in TDBE, this could begin with pushing a static mock key (instead of static mock key being embedded in TVE) after fetching the transaction payloads

Issue #1992, opened 2020-07-23
Problem: light client not used in TDBE to retrieve txids of transactions to fetch


tomtau:

tomtau:
one other thing is that there's no "trusted time" in enclaves -- one can only use the previous committed block's block time or genesis time (can be one of compile-time parameters)

PR #2136, opened 2020-08-18
Problem: Missing runner for TDBE

devashishdxt:
Solution: Created a WIP `TdbeApp` and `spawn_tdbe()` which can be called from `main()` once all the components are ready. Related to #1994, #1995, #2051

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -569,6 +569,7 @@ dependencies = [
  "sgxs-loaders",
  "sha3",
  "structopt",
+ "tdbe-common",
  "test-common",
  "thiserror",
  "tokio 0.2.20",
--- a/chain-abci/Cargo.toml
+++ b/chain-abci/Cargo.toml
@@ -46,6 +46,7 @@ enclave-runner = {version = "0.4", optional = true}
 sgxs-loaders = {version = "0.2", optional = true}
 tokio = { version = "0.2", features = ["uds"], optional = true }
 rand = "0.7"
+tdbe-common = { path = "../chain-tx-enclave-next/tdbe/tdbe-common" }
 
 [build-dependencies]
 cc = "1.0"
--- a/chain-abci/src/enclave_bridge/edp/mod.rs
+++ b/chain-abci/src/enclave_bridge/edp/mod.rs
@@ -1,4 +1,5 @@
 mod server;
+pub mod tdbe;
 
 use crate::enclave_bridge::EnclaveProxy;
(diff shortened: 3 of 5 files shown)


tomtau:
https://github.com/crypto-com/chain/blob/master/chain-abci/src/main.rs#L42

devashishdxt:
> 1. transaction IDs will be fetched by TDBE, not provided from outside -- so it should be provided the remote node's TM RPC from which it can fetch both TX IDs and its TDBE address (it'll just need a chain-abci query endpoint for that)

Won't this be done in #1992?

Just FYI, this is an intermediate PR which adds the base code which was present earlier. It can modified for each issue separately.

devashishdxt:
> > 1. transaction IDs will be fetched by TDBE, not provided from outside -- so it should be provided the remote node's TM RPC from which it can fetch both TX IDs and its TDBE address (it'll just need a chain-abci query endpoint for that)
> 
> Won't this be done in #1992?
> 
> Just FYI, this is an intermediate PR which adds the base code which was present earlier. It can modified for each issue separately.

Or do you think all these related issues should be fixed in a single PR?

tomtau:
> Won't this be done in #1992?
Yes, so this PR should provide something that will help with that -- e.g. runner will be getting/providing RPC config + address that will be exposed via chain-abci query.

> Or do you think all these related issues should be fixed in a single PR?
> Just FYI, this is an intermediate PR which adds the base code which was present earlier. It can modified for each issue separately.

No, but the PR should solve some problem -- if the runner here isn't useful / needs to be replaced in any follow up PRs, this PR can be solved by using the default "ftxsgx-runner" (it won't work, but neither will the runner in this PR)

devashishdxt:
> > Won't this be done in #1992?
> 
> Yes, so this PR should provide something that will help with that (runner needs to be compatible with that) -- e.g. runner will be getting/providing RPC config + address that will be exposed via chain-abci query.
> 
> > Or do you think all these related issues should be fixed in a single PR?
> > Just FYI, this is an intermediate PR which adds the base code which was present earlier. It can modified for each issue separately.
> 
> No, but the PR should solve some problem -- if the runner here isn't useful / needs to be replaced in any follow up PRs, this PR's problem can be solved by using the default "ftxsgx-runner" (it won't work, but neither will the runner in the current state -- and it's simpler, as no new code needs to be added)

Ok. I'll modify. Closing this PR for now.

Issue #2138, opened 2020-08-18
Problem: client-cli sync produces panic after sync


leejw51crypto:
it produces this error after sync
```
thread 'main' panicked at 'get light-client' ..
```

PR #2139, opened 2020-08-18
Problem: client-cli sync produces panic after sync #2138

leejw51crypto:
Solution: add `Option` check

### DIFF
--- a/client-cli/src/command.rs
+++ b/client-cli/src/command.rs
@@ -575,11 +575,11 @@ impl Command {
                     handle.clone(),
                 );
                 Self::resync(config, name.clone(), enckey, *force, storage)?;
-                handle
-                    .as_ref()
-                    .expect("get light-client")
-                    .terminate()
-                    .expect("terminate light client supervisor in client-cli");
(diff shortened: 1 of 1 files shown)


client-cli/src/command.rs
```diff
@@ -575,11 +575,13 @@ impl Command {
...
-                    .expect("get light-client")
-                    .terminate()
-                    .expect("terminate light client supervisor in client-cli");
+                if handle.is_some() {
+                    handle
+                        .as_ref()
+                        .expect("get light-client")
+                        .terminate()
+                        .expect("terminate light client supervisor in client-cli");
+                }
```
linfeng-crypto, line 584: how about:
```
                if let Some(handle) = handle {
                    handle
                    .as_ref()
                    .terminate()
                    .expect("terminate light client supervisor in client-cli");
                }
```

leejw51crypto: looking good, i'll modify

devashishdxt (approved):

Issue #2093, opened 2020-08-10
Problem: client-cli wallet sync in integration test fail occasionally


yihuang:
```
pytests/test_client_cli.py:224:
--
1768 | _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
1769 |  
1770 | self = <client_cli.Wallet object at 0x7fc42dee65e0>, force = False
1771 | disable_address_recovery = False, enable_fast_forward = True, batch_size = 20
1772 | block_height_ensure = 50
1773 | light_client_peers = '0000000000000000000000000000000000000000@127.0.0.1:26657,1000000000000000000000000000000000000000@127.0.0.1:26657'
1774 |  
1775 | def sync(
1776 | self,
1777 | force=False,
1778 | disable_address_recovery=False,
1779 | enable_fast_forward=True,
1780 | batch_size=20,
1781 | block_height_ensure=50,
1782 | light_client_peers="0000000000000000000000000000000000000000@127.0.0.1:26657,1000000000000000000000000000000000000000@127.0.0.1:26657"
1783 | ):
1784 | cmd = ["client-cli", "sync", "-n", self.name, "--batch-size", str(batch_size), "--block-height-ensure", str(block_height_ensure),
1785 | "--light-client-peers", light_client_peers]
1786 | if force:
1787 | cmd.append("--force")
1788 | if disable_address_recovery:
1789 | cmd.append("--disable-address-recovery")
1790 | if enable_fast_forward:
1791 | cmd.append("--enable-fast-forward")
1792 | args = self.auth_token
1793 | text = run(cmd, args)
1794 | if "Error" in text:
1795 | >           raise Exception(text)
1796 | E           Exception: Enter authentication token: Error: Tendermint RPC error: Error while calling tendermint RPC call
1797 | E            => Error response from tendermint RPC for request method (block_results) and params ([String("444")]): RPC error response: JsonRpcError { code: -32603, message: "Internal error", data: Some(String("could not find results for height #444")) }
1798 | E
1799 | E           Stack backtrace:
1800 | E              0: anyhow::private::new_adhoc
1801 | E                        at /drone/src/drone/cargo/registry/src/github.com-1ecc6299db9ec823/anyhow-1.0.32/src/backtrace.rs:10
[...]

PR #2147, opened 2020-08-20
Problem (Fix #2145): incorrect confirmed_transcript_hash calculation in mls

### LINKED ISSUE #2145: Problem: incorrect confirmed_transcript_hash calculation in mls
https://github.com/crypto-com/chain/blob/master/chain-tx-enclave-next/mls/src/group.rs#L395

`get_init_confirmed_transcript_hash` is only valid for the initial group state (0), as it hardcodes "" as `interim_transcript_hash`... for the rest, it should be calculated as:
```
interim_transcript_hash_[n+1] =
    Hash(confirmed_transcript_hash_[n] ||
        MLSPlaintextCommitAuthData_[n]);
```

yihuang:
Solution:
- update implementation according to spec

### DIFF
--- a/chain-tx-enclave-next/mls/src/group.rs
+++ b/chain-tx-enclave-next/mls/src/group.rs
@@ -49,6 +49,7 @@ pub struct GroupAux<CS: CipherSuite> {
     // public and shared
     pub context: GroupContext<CS>,
     pub tree: TreePublicKey<CS>,
+    pub interim_transcript_hash: Option<HashValue<CS>>,
 
     // public and specific to current participant
     /// position of the participant in the tree
@@ -80,6 +81,7 @@ impl<CS: CipherSuite + Ord> GroupAux<CS> {
         Ok(GroupAux {
             context,
             tree,
+            interim_transcript_hash: None,
             my_pos,
             tree_secret: TreeSecret::default(),
             secrets,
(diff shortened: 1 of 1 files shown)


chain-tx-enclave-next/mls/src/group.rs
```diff
@@ -179,24 +181,28 @@ impl<CS: CipherSuite + Ord> GroupAux<CS> {
...
             sender,
             commit.clone(),
         )
         .get_encoding();
-        let to_hash = [interim_transcript_hash, content_to_commit].concat();
+        let to_hash = [
+            self.interim_transcript_hash
+                .as_ref()
+                .map(|h| h.as_ref())
+                .unwrap_or_default(),
```
tomtau, line 197: maybe worth expanding on what is going on here / why it matches the spec -- the type is (`&Vec<u8>.as_ref()`) `&[u8]` and its default == `b""` ? is that correct?

yihuang: yes, changed to more explicit `unwrap_or(b"")`, and copied the spec as comment.

yihuang:
> Build failed (retrying...):
> 
> * [drone-hw/push](http://13.82.172.109/crypto-com/chain/1222)

#2093 again

tomtau:
bors retry

PR #2166, opened 2020-08-23
Problem: indirect communication between tx-validation and tx-query (wip #2051)

leejw51crypto:
intermediate pr for #2051
- expose enclave 2 enclave communication
- tx-validation as server
- tx-query as client
- using unix-stream
- calls are using file socket, it's synchronous call
- concurrent calls in tx-query protected by   `Arc<Mutex<TcpStream>>`

### DIFF
--- a/chain-abci/src/main.rs
+++ b/chain-abci/src/main.rs
@@ -16,7 +16,6 @@ use std::env::var;
 use std::fs::{create_dir_all, write, File};
 use std::io::BufReader;
 use std::net::SocketAddr;
-#[cfg(all(not(feature = "mock-enclave"), feature = "edp", target_os = "linux"))]
 use std::os::unix::net::UnixStream;
 use std::path::{Path, PathBuf};
 use structopt::StructOpt;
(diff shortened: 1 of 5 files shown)


enclave-protocol/src/lib.rs
```diff
@@ -61,6 +61,7 @@ pub enum IntraEnclaveRequest {
     },
     EndBlock,
     Encrypt(Box<IntraEncryptRequest>),
+    General(String),
```
tomtau, line 64: what is "General(String)" supposed to do?

chain-abci/src/main.rs
```diff
@@ -191,13 +190,13 @@ pub struct AbciOpt {
...
-fn get_enclave_proxy() -> TxValidationApp {
-    launch_tx_validation()
+fn get_enclave_proxy(stream_to_txquery: UnixStream) -> TxValidationApp {
+    launch_tx_validation(stream_to_txquery)
 }
 
 /// for development
 #[cfg(any(feature = "mock-enclave", not(target_os = "linux")))]
-fn get_enclave_proxy() -> MockClient {
+fn get_enclave_proxy(stream_to_txquery: UnixStream) -> MockClient {
```
tomtau, line 199:
```suggestion
fn get_enclave_proxy(_stream_to_txquery: UnixStream) -> MockClient {
```

chain-tx-enclave-next/tx-query-next/enclave-app/src/sgx_module.rs
```diff
@@ -24,12 +26,37 @@ use self::handler::{
 };
 use chrono::Duration;
 
+fn callToTxquery(
```
tomtau, line 29: why camel case?

tomtau (changes requested):
if this PR is meant to mean to just add the communication channel, but not add any communication -- then there is no need to extend the enclave protocol (~> "General(String)" variant can be removed)

enclave-protocol/src/lib.rs
leejw51crypto: it's just for testing,i'll remove it.

leejw51crypto:
ok, i'll remove General
this pr is for basic communication between tx-query and tx-enclave.
after that, TLS will be added here

chain-tx-enclave-next/tx-query-next/enclave-app/src/sgx_module.rs
leejw51crypto: i'll change  to _ style

tomtau:
discussed with @leejw51crypto on Slack -- this is to be closed, but can be used for reference for https://github.com/crypto-com/chain/issues/1994

PR #2169, opened 2020-08-24
Problem: legacy redeem address textual conversion code (fix #2150)

### LINKED ISSUE #2150: Problem: legacy redeem address textual conversion code
https://github.com/crypto-com/chain/blob/master/chain-core/src/init/address.rs#L391
`to_cro` and `from_cro` on `RedeemAddress` generally don't make sense with the existing conventions / aren't used anywhere besides the test above

leejw51crypto:
Solution: remove un-necessary unit test

### DIFF
--- a/chain-core/src/init/address.rs
+++ b/chain-core/src/init/address.rs
@@ -11,7 +11,6 @@
 //!   These 20 bytes are the address.
 //!
 //! [Recommended Read](https://kobl.one/blog/create-full-ethereum-keypair-and-address/)
-use bech32::{self, u5, FromBase32, ToBase32};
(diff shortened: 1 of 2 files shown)


tomtau (changes requested):
this can also be removed: https://github.com/crypto-com/chain/blob/master/chain-core/src/init/address.rs#L199 ?

leejw51crypto:
yes, that need to be removed, too.
i'll remove it

tomtau (changes requested):
a few unused imports:
https://github.com/crypto-com/chain/runs/1020049723#step:6:770

leejw51crypto:
ok, i'll check that

tomtau:
squash commits?

leejw51crypto:
ok

PR #2178, opened 2020-08-25
Problem: ra-sp-server crashes sometimes

### LINKED ISSUE #2175: Probem: ra-sp-server crashes sometimes
it crashes with message
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: RaContextError(AesmError(AesmCommunication(Os { code: 11, kind: WouldBlock, message: "Resource temporarily unavailable" })))', chain-tx-enclave-next/enclave-ra/ra-sp-server/src/main.rs:15:18
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
currently, any error occurs, ra-sp-server exit from the loop,
instead of it,
sleep a little
continue processing (try again)

even though aesm service is down and up again,
make ra-sp-server can continue the service

devashishdxt:
Solution: Added 6 retries for `AesmClient` at an interval of 2 secs each. Fixes #2175.


leejw51crypto (approved):
lgtm

Issue #2174, opened 2020-08-24
Problem:frequent tx-query BadCerficate error on devnet


leejw51crypto:
change current from TQE_MRENCLAVE to 
let: Option<[u8; 32]> = Some(include_bytes!("../../../tqe.mrenclave").clone());
with incorrect TQE_MRENCLAVE,
Bad Certificate error appears in chain-abci

because with TQE_MRENCLAVE, it's necessary to recompile all for dependency.
but include_bytes , it compiles dependent crate automatically.

PR #2180, opened 2020-08-25
Problem (WIP #1996): no support for genesis build procedure involving TDBE

### LINKED ISSUE #1996: Problem: no support for genesis build procedure involving TDBE

yihuang:
Solution:
- preliminary try implemented in build.sh

### DIFF
--- a/chain-tx-enclave-next/tx-validation-next/src/sgx_module.rs
+++ b/chain-tx-enclave-next/tx-validation-next/src/sgx_module.rs
@@ -10,7 +10,7 @@ use aes_gcm_siv::Aes128GcmSiv;
 use chain_core::tx::TX_AUX_SIZE;
 use chain_tx_filter::BlockFilter;
 use chain_tx_validation::Error;
-use enclave_macro::{get_network_id, get_tdbe_mrenclave};
+use enclave_macro::get_network_id;
 use enclave_protocol::{IntraEnclaveRequest, IntraEnclaveResponse, IntraEnclaveResponseOk};
 use enclave_utils::tls::{create_ra_context, create_tls_client_stream};
 use parity_scale_codec::{Decode, Encode};
@@ -24,6 +24,7 @@ use std::sync::Arc;
 
 /// FIXME: genesis app hash etc.?
 pub const NETWORK_HEX_ID: u8 = get_network_id!();
+pub const TDBE_MRENCLAVE: &[u8; 32] = include_bytes!("tdbe.mrenclave");
 
 pub(crate) fn write_response<I: Write>(response: IntraEnclaveResponse, output: &mut I) {
     if let Err(e) = output.write_all(&response.encode()) {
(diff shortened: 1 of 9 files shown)


tomtau (commented):
bors try

chain-tx-enclave-next/tdbe/tdbe-common/src/lib.rs
```diff
@@ -10,3 +11,10 @@ pub struct TdbeConfig {
     /// Transaction IDs to fetch from another TDBE server
     pub transaction_ids: Vec<TxId>,
 }
+
+const LIGHT_GENESIS: &str = env!("LIGHT_GENESIS");
```
tomtau, line 15: if "light_genesis" is in json, it'll be fairly error-prone (escaping-characters, length limitations on environment variables, etc.) -- `include_str!` may be better

chain-tx-enclave-next/tx-validation-next/src/sgx_module.rs
```diff
@@ -19,6 +19,7 @@ use std::sync::Arc;
 
 /// FIXME: genesis app hash etc.?
 pub const NETWORK_HEX_ID: u8 = get_network_id!();
+pub const TDBE_MRENCLAVE: &str = env! {"TDBE_MRENCLAVE"};
```
tomtau, line 22: maybe similar to https://github.com/crypto-com/chain/issues/2174 ?
```suggestion
pub const TDBE_MRENCLAVE: [u8; 32] = include_bytes! {"tdbe.mrenclave"};
```

chain-tx-enclave-next/tdbe/tdbe-common/src/lib.rs
yihuang: done

tomtau:
bors try

tomtau (approved):

tomtau (commented):
@yihuang for the genesis build procedure, there may still be a few things missing:
* generating the keypackage and sealing the secrets -- currently in the integration tests, this is done via "mls" enclave, but that should be removed -- instead `TdbeStartupConfig` can have option for this where it'll generate the keypackage + it'll also send back the sealed state with secrets
* for the group creator, there should also be`TdbeStartupConfig` option to generate genesis-time Welcome/Commit/Add payloads (and provide the other nodes' keypackages)
* other nodes TDBE should be able to process those handshake payloads
* when starting up, recovered from that sealed genesis state / check it

ok to merge this PR, but it doesn't fix #1996 yet

tomtau (commented):
@yihuang for the genesis build procedure, there may still be a few things missing:

* generating the keypackage and sealing the secrets -- currently in the integration tests, this is done via "mls" enclave, but that should be removed -- instead TdbeStartupConfig can have option for this where it'll generate the keypackage + it'll also send back the sealed state with secrets
* for the group creator, there should also beTdbeStartupConfig option to generate genesis-time Welcome/Commit/Add payloads (and provide the other nodes' keypackages)
* other nodes TDBE should be able to process those handshake payloads
* when starting up, recovered from that sealed genesis state / check it

ok to merge this PR, but it doesn't fix #1996 yet

leejw51crypto (approved):
lgtm

chain-tx-enclave-next/tx-validation-next/src/sgx_module.rs
```diff
@@ -19,6 +19,7 @@ use std::sync::Arc;
 
 /// FIXME: genesis app hash etc.?
 pub const NETWORK_HEX_ID: u8 = get_network_id!();
+pub const TDBE_MRENCLAVE: &[u8; 32] = include_bytes!("tdbe.mrenclave");
```
tomtau, line 22: `get_tdbe_mrenclave` proc macro can be removed (in the enclave-macro)
and this call can be replaced with `TDBE_MRENCLAVE`
https://github.com/crypto-com/chain/blob/master/chain-tx-enclave-next/tx-validation-next/src/sgx_module.rs#L38

yihuang: done

PR #2184, opened 2020-08-25
 Problem: Ambiguous "synchronization complete" message (fix #2168)

### LINKED ISSUE #2168: Problem: Ambiguous "synchronization complete" message
**Describe the bug**
When trying it out with @brianatcrypto, syncing works fine until it reaches the block that contains transaction(s) related to the wallet.  
When we try to sync it again with different the `batch-size`,the client-cli was "fooled" and show `Synchronization complete!`
 
**Example:**
 ```client-cli sync  --name Default_HD  -b 1000
Enter authentication token:
Synchronizing: 813 / 813 [==============================================] 100.00 % 133.16/s
Synchronization complete!

Error: Connection error: Unable to connect to TQE address: 20.49.200.60:3322
```

**To Reproduce**
1. Sync the wallet;
2. The sync is running until it reaches the "problematic"  block. "`Unable to connect to TQE address ...`" error message is shown;
3. Sync it again with different `batch-size`;
4. It shows `Synchronization complete!` but infact, it is not synced.

leejw51crypto:
1. remove "sync complete" message in fetch blocks
display only in "validation" step

2. fix redundant sync complete message , display only once

### DIFF
--- a/client-cli/src/command.rs
+++ b/client-cli/src/command.rs
@@ -853,7 +853,7 @@ impl Command {
                 } => {
                     if let Some(ref mut pb) = progress_bar {
                         if current_block_height == final_block_height {
-                            pb.finish_println("Synchronization complete!\n");
+                            pb.finish_print("Synchronization complete!");
                         } else {
                             pb.set(current_block_height - init_block_height);
                         }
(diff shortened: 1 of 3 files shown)


devashishdxt (changes requested):
There are some build failures.

leejw51crypto:
reproduced sync error,
i'll re-submit

PR #2185, opened 2020-08-25
 Problem: Ambiguous "synchronization complete" message (fix #2168)

### LINKED ISSUE #2168: Problem: Ambiguous "synchronization complete" message

leejw51crypto:
Solution: display progress only in validation

### DIFF
--- a/client-cli/src/command.rs
+++ b/client-cli/src/command.rs
@@ -853,7 +853,7 @@ impl Command {
                 } => {
                     if let Some(ref mut pb) = progress_bar {
                         if current_block_height == final_block_height {
-                            pb.finish_println("Synchronization complete!\n");
+                            pb.finish_print("Synchronization complete!");
                         } else {
                             pb.set(current_block_height - init_block_height);
                         }
@@ -868,8 +868,7 @@ impl Command {
         if force {
             syncer.reset_state()?;
         }
-        syncer.sync(progress_callback)?;
-        Ok(())
+        syncer.sync(progress_callback)
     }
 }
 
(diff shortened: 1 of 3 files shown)


leejw51crypto:
for changing batch size
let's say, txblock is 28, and batch-size is 20
it will stop at block 21

if you change to batch-size to 1
it'll continue till block 27

tomtau (commented):
seems ok, but @lezzokafka / @brianatcrypto may be better to comment if this resolves the problem

client-cli/src/command.rs
```diff
@@ -853,7 +853,7 @@ impl Command {
                 } => {
                     if let Some(ref mut pb) = progress_bar {
                         if current_block_height == final_block_height {
-                            pb.finish_println("Synchronization complete!\n");
+                            pb.finish_print("Synchronization complete!");
```
tomtau, line 856: why no new lines?

leejw51crypto: println + \n makes double display of the same message , so changed to like this

PR #2204, opened 2020-09-04
Problem (Fix #2203): validator crash after re-joined with different validator key

### LINKED ISSUE #2203: Problem: validator crash after re-joined with different validator key

yihuang:
Solution:
- deduplicate staking address when doing cleanup
- index used_validator_addresses when initialize after load from storage
- changed integration test to covert the use case


tomtau (approved):
bors r+

PR #2459, opened 2021-01-18
Problem: no nonce for remote attestation report

trevor-crypto:
This was an issue reported by kudelski. Intel claims that the nonce for remote attestation is optional, but I suppose it's better to be safer and include it.

@devashishdxt do you happen to know why it was left out? Maybe I am missing something, or it is truly unnecessary. Thanks

### DIFF
--- a/chain-tx-enclave-next/enclave-ra/ra-sp-client/src/client.rs
+++ b/chain-tx-enclave-next/enclave-ra/ra-sp-client/src/client.rs
@@ -65,8 +65,9 @@ impl SpRaClient {
     pub fn get_attestation_report(
         &self,
         quote: Vec<u8>,
+        ias_nonce: String,
     ) -> Result<AttestationReport, SpRaClientError> {
-        let request = Request::GetAttestationReport { quote };
+        let request = Request::GetAttestationReport { quote, ias_nonce };
         serde_json::to_writer(&self.stream, &request)?;
 
         let response: Response = serde_json::Deserializer::from_reader(&self.stream)
@@ -92,4 +93,6 @@ pub enum SpRaClientError {
     NoResponse,
     #[error("Unexpected response: {0:?}")]
     UnexpectedResponse(Response),
+    #[error("{0}")]
+    Nonce(String),
 }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3784,6 +3784,7 @@ dependencies = [
  "chrono",
  "cmac",
  "crypto-mac 0.8.0",
+ "hex",
  "log",
  "ra-common",
  "ra-sp-client",
--- a/chain-tx-enclave-next/enclave-ra/ra-common/src/sp/attestation_evidence.rs
+++ b/chain-tx-enclave-next/enclave-ra/ra-common/src/sp/attestation_evidence.rs
@@ -14,11 +14,11 @@ pub struct AttestationEvidence {
 
 impl AttestationEvidence {
     /// Creates attestation evidence from quote
-    pub fn from_quote(quote: &[u8]) -> Self {
+    pub fn from_quote(quote: &[u8], nonce: Option<String>) -> Self {
         Self {
             isv_enclave_quote: base64::encode(quote),
             pse_manifest: None,
-            nonce: None,
+            nonce,
         }
     }
 }
--- a/chain-tx-enclave-next/enclave-ra/ra-common/src/sp/protocol.rs
+++ b/chain-tx-enclave-next/enclave-ra/ra-common/src/sp/protocol.rs
@@ -10,7 +10,7 @@ pub enum Request {
     /// Generates a new quote from QE using AESM
     GetQuote { report: Vec<u8>, nonce: [u8; 16] },
     /// Generate attestation report using IAS
-    GetAttestationReport { quote: Vec<u8> },
+    GetAttestationReport { quote: Vec<u8>, ias_nonce: String },
 }
 
 /// Responses for request sent by enclave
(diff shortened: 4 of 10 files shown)


chain-tx-enclave-next/enclave-ra/ra-sp-client/src/client.rs
```diff
@@ -65,8 +65,9 @@ impl SpRaClient {
     pub fn get_attestation_report(
         &self,
         quote: Vec<u8>,
+        nonce: [u8; 16],
     ) -> Result<AttestationReport, SpRaClientError> {
-        let request = Request::GetAttestationReport { quote };
+        let request = Request::GetAttestationReport { quote, nonce };
         serde_json::to_writer(&self.stream, &request)?;
 
         let response: Response = serde_json::Deserializer::from_reader(&self.stream)
```
tomtau, line 73: should also verify the response contains the same nonce ?

tomtau (changes requested):
looks ok, but after retrieving a report from IAS, I assume one should check that the nonce in a response is the same as the one in a corresponding request?

chain-tx-enclave-next/enclave-ra/ra-sp-client/src/client.rs
trevor-crypto: Yes, that seems correct according to the docs. Thanks

chain-tx-enclave-next/enclave-ra/ra-sp-client/src/client.rs
```diff
@@ -76,7 +77,17 @@ impl SpRaClient {
             .ok_or(SpRaClientError::NoResponse)?;
 
         match response {
-            Response::GetAttestationReport { attestation_report } => Ok(attestation_report),
+            Response::GetAttestationReport { attestation_report } => {
+                // validate nonce
```
trevor-crypto, line 81: Unfortunate that we have to deserialize here, and then downstream, but it's because the nonce is not available to the caller in the other crates

chain-tx-enclave-next/enclave-ra/ra-sp-client/src/client.rs
```diff
@@ -76,7 +77,20 @@ impl SpRaClient {
...
+                let body: AttestationReportBody = serde_json::from_slice(&attestation_report.body)?;
+                let ret_nonce = body.nonce.ok_or_else(|| {
+                    SpRaClientError::Nonce("No nonce found in attestation report.".to_string())
+                })?;
+                if ret_nonce.as_bytes() == nonce {
```
tomtau, line 86: probably ok (not sure if this timing leak can be hijacked in enclaves), but perhaps constant time comparison for sanity.

tomtau (changes requested):
there's a small compilation error: https://github.com/crypto-com/thaler/runs/1721564502#step:6:880
+ could you squash the commits https://rfc.zeromq.org/spec/42/#23-patch-requirements ?

chain-tx-enclave-next/enclave-ra/ra-sp-client/src/client.rs
trevor-crypto: Putting all my faith in sgx :smile:. Besides, won't LLVM muck up things anyway, if we do a constant time comparison?

tomtau: `subtle` tries to prevent the compiler from optimizing away constant time comparisons: https://doc.dalek.rs/src/subtle/lib.rs.html#211 so with that lib or similar, it shouldn't

tomtau:
bors retry

tomtau:
bors retry

tomtau (commented):
btw, given there's "nonce" in quote, maybe better to call this new one "ias_nonce" or something like that, as there are places where both of them are used in the same context (so that it doesn't get confused)

trevor-crypto:
bors try

devashishdxt:
bors try

devashishdxt:
```
[DEBUG rustls::client::tls13] Attempting client auth
[WARN  rustls::session] Sending fatal alert BadCertificate
[ERROR rustls::session] TLS alert received: Message {
        typ: Alert,
        version: TLSv1_3,
        payload: Alert(
            AlertMessagePayload {
                level: Fatal,
                description: BadCertificate,
            },
        ),
    }
[ERROR tx_validation_next::sgx_module] failed to read the mock key
```

tomtau:
@trevor-crypto 
```
/drone/src/integration-tests
156	Starting ra-sp-server...
157	Wait for tcp port 8989
158	[Wed Jan 20 01:36:07 UTC 2021] Tcp port 8989 not available yet, sleep 2 seconds and retry
159	[Wed Jan 20 01:36:09 UTC 2021] Tcp port 8989 not available yet, sleep 2 seconds and retry
160	Tcp port 8989 is available now
161	Preparing... WITH_FEE
162	[2021-01-20T01:36:18Z ERROR ra_sp_server::server] SP RA error: RA context error: IAS client error: Invalid nonce in attestation verification report
163	Error: Run enclave error: enclave runner return error code: Some(255), stderr: Error while executing SGX enclave.
164	Enclave panicked: thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: RaContextError(CertificateCreationError)', chain-tx-enclave-next/mls/src/main.rs:21:27
165	note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
166	
167	
```

tomtau:
bors try

tomtau:
bors retry

